PR #112, opened 2016-09-02
Add kappa to list of valid classes and family-specific docs.

aforren1:
This allows us to use `set_prior` for the kappa parameter in the `von_mises` distribution, e.g.

```
set_prior('normal(50,1)', class = 'kappa')
```

Also added a short blurb for kappa in the docs, and did some nitpicky doc fixes ('it' -> 'is' in priors.R, and lowercase 'c' for 'credible intervals' in the readme).


paul-buerkner:
@aforren1 Many thanks!

PR #330, opened 2018-01-22
Prepare for use with precompiled model dmenne/brmspack

dmenne:
- Added getCall.brmsfit (required, see update.default
- Added recompile = "never". Not the perfect code, but avoid portability problems. Maybe we find out why precompile code seems to be different, then this change could be dropped.


paul-buerkner:
Thanks! One small question: Why do we need the `getCall` method. Is there a reason, we cannot use the `formula` method or extract the formula directly?

dmenne:
Without getCall, I get the error "need element with call signature"  when I use a save and load in another environment. Alternatively, add a $call = $formula to the output, but that's not so clean because we double an element.

paul-buerkner:
I see, thanks. A few hours ago, I already made a change to `recompile` so that `recompile = FALSE` really means not recompiling the Stan code regardless of what happens. I also just added the `getCall` method, so that you should be able to proceed further. Thank you so much for going into this!

dmenne:
>> I already made a change to recompile so that recompile = FALSE

Never dared to do that. Fine for me, but I suspect it could lead to some nasty errors in "normal" cases.

paul-buerkner:
I also changed the default to `NULL` so that users not specifying this argument explicitely won't notice any difference. This structure (if `NULL` determine internally whether `TRUE` or `FALSE`) is actually much nicer I think and I already use in some other places across the package.

dmenne:
Just noted it... looks fine.

PR #347, opened 2018-02-12
Fix minor mistakes in the cor_ar() and cor_ma() documentation

lemna:
Hi Paul,

the default value for `p` in `cor_ar()` and for `q` in `cor_ma()` is 1.  The documentation inherited the defaults from `cor_arma()` although they are different.

With kind regards,
Peter.


paul-buerkner:
Your are absolutely right! Many thanks for fixing it!

PR #411, opened 2018-04-24
Use brms-package.R

jgabry:
This PR adds `R/brms-package.R` which, after you delete `man/brms-package.Rd`, will recreate `man/brms-package.Rd` and can be used for updating that file going forward. 

This only matters because it works better for the way the documentation is generated for the website if this file exists in the R directory. While I was at it, I also added the Stan logo to this file in the same way the other R packages have it (see e.g. `help("loo-package")` or http://mc-stan.org/loo/reference/loo-package.html or for the other R packages). 

I still need to figure out some of the details for getting the website to work, but this is a low hanging fruit we can get out of the way now. 

Also, I checked the `Allow edits from maintainers` box so you can push directly to my master branch before merging the PR, although I guess you can just edit how you like after merging too.


paul-buerkner:
Thanks so much Jonah! If I need to adjust anything, I can just do it on the Master branch.

PR #415, opened 2018-04-26
Replace inst/NEWS.rd with NEWS.md

jgabry:
This is going to be necessary for the website if you want it to have a News tab. We can have it automatically generate the page from NEWS.md, e.g., http://mc-stan.org/bayesplot/news/index.html. 

(Now that I’m typing this I’m realizing I need to do this for rstanarm too since it still uses inst/NEWS.)


paul-buerkner:
Nice! Thank you!

jgabry:
No problem! There are more things related to the website that I’ll make PRs
for so you won’t have to spend time figuring out what Imad and I already
had to spend time figuring out when we were doing the sites for the other
packages.

Issue #155, opened 2016-12-12
More vignettes


paul-buerkner:
With the implementation of several new features in the upcoming **brms** 1.3.0 release, some more vignettes are required to explain certain aspects of the package in detail.

- [x] Multilevel models (actually write a full paper)
- [ ] Prior distributions
- [ ] Noise-in-variables models (within the missing values vignette?)
- [x] Response time models
- [ ] Mixture models
- [x] Gaussian processes
- [ ] Survival models
- [x] Bayes factors
- [x] Multivariate models
- [x] Missing value imputation

t-barrett:
It would be great to have a vignette for response time models! Thank you for all of your progress.

ianhussey:
Seconded. Mixed models for RT data are increasingly commonplace (Balota & Yap, 2011; Lo & Andrews, 2015), but I'm still encountering many researcher degrees of freedom in constructing them in brms. Mixed within-between designs (e.g,. block*condition) cover a large range of paradigms and user cases, such as stroop, IAT, flanker, task switching, etc.

paul-buerkner:
I agree that a vignette discussing RT models is needed. I wonder though, how such a vignette would lead to a reduction in researchers degrees of freedom? After all, I think it should be about reasonable options and serve as a starting point, which may even increase the researchers degrees of freedom.

paul-buerkner:
Henrik Singmann has written a blog post about the Wiener Diffusion model (http://singmann.org/wiener-model-analysis-with-brms-part-i/), which might be helpful to many using brms for the analysis of response times.

paoloeusebi:
It would be great to have something related to multilevel mixture models in the "Advanced Bayesian Multilevel Modeling with the R Package brms" vignette. I don't know if it's possible to estimate such models with brms (see Vermunt 2003. "Multilevel Latent Class Models". Sociological Methodology, 33(1), 213–239.)

paul-buerkner:
Not in that vignette as it corresponds to an article published in the R journal but of course having a vignette about mixture models would be desireble. If these exact models are possible in brms is another story I have to dig into once I have the time.

paul-buerkner:
brms is already quite large in term of MB pre-compliation which makes CRAN a bit unhappy. More vignettes will just make it even larger. Given the amount of blog posts and tutorials about brms on the internet, I think it is okay to not add more vignettes to brms itself for the time being. Closing this issue now.

PR #420, opened 2018-04-29
 Fix missing parameter problem in shinystan

jgabry:
I think [shinystan issue #155](https://github.com/stan-dev/shinystan/issues/155) can actually be fixed easier than I thought from inside brms. This PR should take care of it I believe. The trick is to not use the method for stanfit objects and instead to create the shinystan object from the relevant various components.

### DIFF
--- a/R/brmsfit-methods.R
+++ b/R/brmsfit-methods.R
@@ -968,7 +968,31 @@ launch_shinystan.brmsfit <- function(
   object, rstudio = getOption("shinystan.rstudio"), ...
 ) {
   contains_samples(object)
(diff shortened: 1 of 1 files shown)


paul-buerkner:
Many thanks! Do I understand it correctly, that this PR fixes it by effectively removing the knowledge about which scalar parameter belongs to which parameter vector?

jgabry:
Yeah exactly. Eventually it would be better to still retain that information but I’ll need to change some things in shinystan that I haven’t sorted out yet. for now this should work and be better than not having the parameters!

PR #434, opened 2018-05-20
gh-pages branch updates for website

jgabry:
This PR contains the website structure and preliminary versions of the site content (to be added to the `gh-pages` branch). It won't be connected automatically to the mc-stan.org domain since this repo is not part of the `stan-dev` GitHub organization yet (still need to figure that part out), but this should set up pretty much everything else. 

From the `gh-pages` branch, to preview the site locally use:

```r
# devtools::install_github("r-lib/pkgdown")
pkgdown::preview_site(path = "../") 
```

and to rebuild the site use 

```r
pkgdown::build_site()
```

but then move _all_ the contents it adds to the `docs` directory into the main package directory (we don't actually use the `docs` directory).


paul-buerkner:
Thank you so much Jonah!!

jgabry:
No problem. Sorry it took a while. I still need to figure out how to configure the domain name using an external repo. Also, are you able to edit something and rebuild locally using pkgdown without any errors?

paul-buerkner:
It works perfectly! Thanks! I have two small questions.

On the citation.html, it doesn't show the umlaut ü in my name correctly, even though the bibtex entries have UTF-8 encoding. Do you know away to work around that?

Out of interest, why don't we use the docs folder and instead copy everything into the main directory?

jgabry:
Not sure about the umlaut. If we can’t figure it out maybe we open a pkgdown issue? With the docs folder, I think it’s because using the gh-pages branch requires everything in the main directory. Could use docs if site content is on master branch (I think) but then the master branch and git history for master branch include all these hundreds of files that are just for the website and contain a lot of duplicate content already on the master branch.

jgabry:
It would also be easy to write a wrapper around build_site that Automatically moves everything out of the docs folder when building. I’ve been meaning on doing that for a while

jgabry:
Actually when I build the site the ü is displayed correctly (on both Chrome and Safari). Which browser are you using?

paul-buerkner:
The problem with the umlauts occurs only at one specific place within the html for me and nowhere else. I think it may not even be worth investigating in more detail since I can just fix those 2 characters whenever we rebuild the page (which happens only when releasing a new version). I also think that copying everything to the main directly once every release is ok and we don't actually need a wrapper for it.

PR #461, opened 2018-06-21
Minor fix in the Examples section

torkar:
Changed to prior_samples from posterior_samples since I assume that's what PC meant.


paul-buerkner:
You are right, thanks! Please note that doc fixes should not be done in the .Rd files directly since they are  autogenerated by roxygen. Rather the fix should be done in the .R files of the /R subdirectory, which contains the doc to be translated to the .Rd files.

torkar:
Aaah, you're completely right of course. Will keep it in mind for the future.

PR #522, opened 2018-09-17
improve citation infomation

XiangyunHuang:


paul-buerkner:
Thank you for updating the citation info!

Issue #540, opened 2018-10-26
Basis function approximation to Gaussian processes


paul-buerkner:
GPs in brms or Stan in general are currently terribly slow for larger data sets.
The GP approximations discussed in https://github.com/gabriuma/basis_functions_approach_to_GP seem to scale much better with data and I think would make a nice feature for brms.

paul-buerkner:
Multidimensional approximate GPs should now work as well. Here is a small example:

```R
dat <- mgcv::gamSim(2, n = 120, scale = 0.3)$data
fit <- brm(y ~ gp(x, z, k = 5), data = dat, chains = 1)
marginal_effects(fit, "x:z")
marginal_effects(fit, "x:z", surface = TRUE, resolution = 30)

```

paul-buerkner:
I have a question for @gabriuma about multidimensional approximate GPs:

In your stan code you write `M_nB` is `M * D`. Does this imply that the basis functions of the multidimensional GP are just the basis functions of the univariate GPs bound together into a matrix with more columns? To what kind of multidimensional splines does that correspond (addtive, multiplicative, or something else)?

gabriuma:
Hi @paul-buerkner, 

@avehtari
@MichaelRiis,

We are working now on the study for the suitable values of L and its implications. Meanwhile, the default of L as |5/4 * (max(x) - min(x))| seems to be appropiate. Inputs x has to enter into the model always individually centered; they need to meet  x \in [-L,L].

About your last comment, the basis functions in the multidimensional case are not just bounding together the univariate basis functions. M^D new basis functions have to be created. Each new one is the product of D univariate basis functions. One example: 

M=2  (two basis functions)
D=3   (three dimensions)

Univariate basis functions:
bf11, bf12, bf13, bf21, bf22, bf23

New (multivariate) basis functions:
bf1 = bf11 .* bf12 .* bf13 
bf2 = bf11 .* bf12 .* bf23
bf3 = bf11 .* bf22 .* bf13 
bf4 = bf11 .* bf22 .* bf23 
bf5 = bf21 .* bf12 .* bf13 
bf6 = bf21 .* bf12 .* bf23
bf7 = bf21 .* bf22 .* bf13
bf8 = bf21 .* bf22 .* bf23 

Also, there are some changes in the spectral density and the lambda expressions. You have all of these defined as functions in the Stan code:

spd_nD: is the expression for the spectral density in the n-dimensional case.
[...]

PR #553, opened 2018-11-19
Approximate Gaussian Processes

paul-buerkner:
This PR implementes approximate Gaussian processes as discussed in #540.

@avehtari, @gabriuma, @MichaelRiis anything you want me to add or change before I merge this PR?


gabriuma:
Some thoughts:
Due to symmetry of the input space can not be guaranteed, even it is mean centered, I think we should define L=c*max(abs(x)). The X-label of the figure could also be changed by rho/max(abs(x)), or something like that.
Although c=5/2 can be appropriate for many datasets, there will be many others where this will not. From the different experiments I've done, I noticed some sensitivity to the choice of c, which can lead to long computational times and poor results.
I think some a priori guess of the lengthscale would be interesting and then, setting the factor c looking at the figure. (Maybe some iterative process would also be interesting to think about).
In the multivariate case, when the lengthscale is significantly different in each dimension, it would also be interesting to set different L and M for each dimension in order to improve the performance. We are working on that.

avehtari:
Before we understand L and c better, this is better to be labeled as experimental and possible that something will change. It's good idea now to require user to make choices.

paul-buerkner:
Ok. A follow up on this. Would you be opposed to using the term `L` as the multiplicative factor in brms (at least user facing), instead of `c` as the latter is a commonly used function in R?

gabriuma:
It is probably better not to use the letter L as notation for the factor c,
in order to be coherent with the related literature. Any other appropiate
notation for the factor c would be fine, I think. Then, we can change the
label in the figure accordingly.

paul-buerkner:
Thanks! Upon further thinking about it, I think it's ok to just use `c`. I will change the accordingly

paul-buerkner:
Would an upper-case `C` work for you as well?

paul-buerkner:
Sorry for the back and forth. I decided that lower-case `c` is ok even if the base combine function has the same name (they won't mask each other so it doesn't matter). That is, we can continue to use the your case study notation when working with brms.

gabriuma:
I meant that 'C' seems good for me.

paul-buerkner:
thanks for the clarification. Indeed, I was unsure between C and c but then decided to use the latter for immediate consistency with the case studies. I will merge this PR now, but we still have roughly a month before the next CRAN release which gives us plenty of time to change argument names, defaults, etc.

Issue #696, opened 2019-07-04
Elicit warning if sample_prior = TRUE but prior was not sampled?


lindeloev:
It seems that there are cases when prior samples are not collected even though `sample_prior = TRUE`. For example when the default priors are used, but also for intercept-only models as [noted on Twitter](https://twitter.com/MichiganWater2/status/1146519717256953856). 

I suggest throwing an informative warning in `brm` when `sample_prior = TRUE` is not obeyed for one or several parameters in the model? For example: "Prior sample were not collected for [list of parameters here]. Specify a proper prior for these."

Either mention `hypothesis` in the warning above, or explicitly elicit a warning when testing a point hypothesis on a parameter with unsampled prior, e.g., `hypothesis(fit, 'Intercept = 0').

Relates to #566

paul-buerkner:
I added this information to the doc of `sample_prior` a few minutes ago. Would you mind taking a loo if this is sensible and understandable. Then we can think of whether an additional warning needs to be thrown.

lindeloev:
I was just about to do a pull request on an alternative formulation, but maybe we should clarify a potential bug first: `hypothesis` use the prior on `b` for the Intercept:

```
data = data.frame(
  y = rnorm(50),
  x = 1:50
)


library(brms)
fit = brm(
  formula = y ~ x,
  data = data,
  prior = set_prior('normal(2, 1)', class='b'),
  sample_prior = TRUE
)
```

```
> head(prior_samples(fit))
         b     sigma
1 3.271250 11.919315
2 1.802418  9.679331
3 3.117318  2.309235
4 1.078793 12.128314
5 2.730517  7.998656
6 2.799320 13.723959
```

```
plot(hypothesis(fit, 'Intercept = 0'))
```
![image](https://user-images.githubusercontent.com/1439417/60655515-282a2880-9e4e-11e9-92d7-0e8ce1dc7ffe.png)

paul-buerkner:
I can confirm this is a bug, I will try to fix it asap.

lindeloev:
I just created a pull request, (hopefully!) improving even more on your updated doc. I tried being more explicit about not sampling from improper default priors. This was just an easier way to propose it than pasting it here, so feel free to reject and make other improvement.

[...]

PR #697, opened 2019-07-04
DOC: sample_prior for improper (default) priors.

lindeloev:
Addresses #696


paul-buerkner:
Thanks! I like the doc improvements!

Issue #680, opened 2019-06-03
Speed issue with cor_lagsar


aslez:
I recently tried to run a cor_lagsar model and found that it was prohibitively slow when working with data with even just a few hundred observations.  I took a look at the code found at least two sources of inefficiency, which I have since realized are not entirely unintentional.  

First, the code calculates the log of the Jacobian---W_tilde---using the log_determinant command.  An alternative approach going back to Ord (1975) is to take the sum of 1 minus the vector of eigenvalues associated with the spatial weights matrix W.  What I have come to realize is that Stan does not allow you to take the eigenvalues of an asymmetric matrix, which makes it difficult to apply the approximation on the fly without taking the eigenvectors as an argument in the normal_lagsar_lpdf function that is produced when brm generates the code for Stan.

Second, the model parameterization follows the discussion here: https://mc-stan.org/loo/articles/loo2-non-factorizable.html.  This makes sense, but it requires inverting W_tilde, which can really slow things down, especially as the number of observations grows.

I've include an R script that shows what happens if you rewrite the code to address these two points.  It seems to go much faster, but I'm wondering whether the change in parameterization will effect the ability to calculate the various LOO-related measures.
[stan_lagsar.zip](https://github.com/paul-buerkner/brms/files/3248636/stan_lagsar.zip)

paul-buerkner:
Thanks a lot! I will take a detailed look in the upcoming days!

aslez:
Paul,

I didn’t say this in my post, but I looked over the LOO link again and realized that the parameterization used in brm follows from there.  It makes sense to me now, but it is definitely going to be much slower than the usual approach.  Eyeballing the run times for the Columbus example, the brm version was taking 15 times longer per chain than the traditional parameterization.  I haven’t tried this with the bigger data yet.  

At the moment, I think the most reasonable way forward for my purposes may be to do everything using the alternative code and figure out how to get LOO to work by hand.

Thanks for taking a look at this.

Adam

Adam Slez 
Assistant Professor 
Department of Sociology 
University of Virginia 
aslez@virginia.edu

paul-buerkner:
Are the two implementations equivalent in terms of the implied likelihood? Because if they are, we could use the Ord (1975) parameterization in the Stan code and (at least for the time being) use the existing implementation for computing the log-likelihood values, even if this is slow.

paul-buerkner:
[...]

PR #704, opened 2019-07-10
Improve efficiency of SAR models

paul-buerkner:
This PR improves the efficiency of SAR models drastically as discussed in #680.

Thanks to @aslez for making this possible!

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,11 @@
 * Predict `sigma` in combination with several 
 autocorrelation structures. (#403)
 
+### Other Changes
+
+* Substantially improve the sampling efficiency of SAR models
+thanks to the GitHub user aslez. (#680)
+
 ### Bug Fixes
 
 * Allow grouping variables to start with a dot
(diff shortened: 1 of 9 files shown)


aslez:
The changes to the error models look right to me.  Unlike the case with the lag models, the existing version of half_pred was fine to begin with (there are technically new lines of code here, but they look the same as what you had before), so all that was required was swapping out the old version of the determinant for the new one.  If I had to guess, I would say that the speed up on the error models should be less pronounced than on the lag models due to the fact that the error models didn't have the matrix inversion issues that the lag models did.

aslez:
One minor point that I forgot to mention.  Not that it matters, but the autocorrelation parameter in the error model is sometimes referred to as lambda, which runs into the naming conventions for the eigenvalues.

paul-buerkner:
It is perhaps slightly less pronounce but still close to an order of magnitude speed up it seems.
I will change lambda to eigenW, which is more explicit and also matches the names in the data block. After that, I think this PR is ready to be merged. Thanks for all your help!

PR #720, opened 2019-07-31
Use the nsamples generic from rstantools (#719)

### LINKED ISSUE #719: use nsamples() generic from rstantools?
`nsamples()` is one of the generics defined in `rstantool`, but it's being defined again in`generics.R`. Shouldn't the `rstantool` version be imported (similarly for what it's done for, say, `posterior_linpred` and `predictive_error`)?

I don't think it really matters either way, but if you want a patch for consistency I'll be happy to contribute it.

mcol:
The requires `rstantools` version 1.5.1 (released August 2018), which I think it's fine, since we are already requiring much more recently released packages (such as `loo` 2.1.0). All tests and R CMD check pass.
This fixes #719.


paul-buerkner:
Perfect, thank you!

PR #794, opened 2019-11-23
Correct minor errors in missing values vignette

markseeto:
Correct typos, and `\code{m}` was not displaying.


paul-buerkner:
Thank you so much!

PR #815, opened 2019-12-25
Fix a bug: as.mcmc(inc_warmup = TRUE)

hoxo-m:
Hi.
This PR is a continuation of the pull request #811.

Your change is not perfect.
For instance, the same error occurs when the following code runs.

```r
library(brms)
fit <- brm(dist ~ speed, cars, thin = 3)
mcmc_list <- as.mcmc(fit, inc_warmup = TRUE)
head(mcmc_list)
```

The cause is that warmup samples and real MCMC samples are thinned separately.
It is difficult to connect them well into one `mcmc` object.
I suggest prohibiting `inc_warmup = TRUE` if `thin >= 2`.

### DIFF
--- a/R/brmsfit-methods.R
+++ b/R/brmsfit-methods.R
@@ -557,11 +557,14 @@ as.mcmc.brmsfit <- function(x, pars = NA, exact_match = FALSE,
     attr(out, "mcpar") <- mcpar
     class(out) <- "mcmc"
   } else {
+    thin <- x$fit@sim$thin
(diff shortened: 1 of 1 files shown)


paul-buerkner:
Thanks for following up on this! I made the changes accordingly. I couldn't easily merge your PR for this purpose because I have moved a whole bunch of files to different locations earlier today so that your PR had several conflicts with master.

hoxo-m:
Thank you.

PR #897, opened 2020-05-13
Added emmeans support

rvlenth:
This adds support for the **emmeans** package to the package code for **brms**. Because of the way it is dispatched, this code will override what already exists in the **emmeans** package.

This added code is essentially what is already provided in the **emmeans** package, but updated to avoid warnings, e.g., it uses `brmsterms` rather than `parse_bf`. It provides basic support for rather standard models, but not fancier ones, e.g. hurdle models. I would hope that by having this in your own package, you can add features to support those models too. I'd be happy to answer questions.

The contributed code consists of methods `recover_data.brmsfit` and `emm_basis.brmsfit`; added code in `zzz.R` to register these methods when **emmeans** is installed on the user's system; and a few tests on basic functionality.

### DIFF
--- a/R/emmeans-support.R
+++ b/R/emmeans-support.R
@@ -1,0 +1,42 @@
+### Rudimentary support for brms objects in emmeans package. 
+### Obviously this is way less than is needed, but it does support simpler models
+
+# These are NOT exported here. They are dynamically exported if emmeans is installed.
+# See code in zzz.R
+
+# Note from RVL: It may be desirable to add optional arguments, especially to
+# the emm_basis method, to allow special options -- e.g., for hurdle
+# models, which part of the model to estimate. If you do that, obviously
+# you will want to document it in some way.
+
+
+# Recover the predictors used in the fixed-effects part of the model
+recover_data.brmsfit <- function(object, data, ...) {
+    bt <- brmsterms(formula(object))
+    if (class(bt) != "brmsterms")
+        stop("This model is currently not supported.")
+    mt <- attr(model.frame(bt$dpars$mu$fe, data = object$data), "terms")
+    ### we don't have a call component so I'll just put in a dummy
+    emmeans::recover_data(call("brms"), mt, "na.omit", data = object$data, ...)
+}
+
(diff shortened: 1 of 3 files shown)


paul-buerkner:
Thank you! I agree that those functions should better live in brms than in emmeans. There are so changes and additions necessary that I need to make before this can be merged into master (some of which you already mentioned). So I will merge this into another branch first and then to master from there once everything works as desired.

rvlenth:
OK, well I have tried a few of your example model, with a bit more success. Perhaps the most interesting was `fit5`, for which the only fixed-effect predictor is `Age`. I got the following result when I used it for a whole set of ages:
```r
emmip(fit5, ~Age, dpar="mu2", at = list(Age = seq(-3,3,by=.25)), type="response")
emmip(fit5, ~Age, dpar="mu1", at = list(Age = seq(-3,3,by=.25)), type="response")
```
image
image

Is this right???

I had no luck with `fit2`. 
```r
> ref_grid(fit2, dpar = "mu")
Error : The select parameter is not predicted by a linear formula. Use the 'dpar' and 'nlpar' arguments to select the parameter for which marginal means should be computed.
> ref_grid(fit2, dpar = "shape")
Error : Parameter 'shape' is not part of the model.
```
I finally did get results with `fit6`, once I discovered the `resp` argument. For multivariate models on the same scale (e.g. repeated measures), there is a provision to cast the multivariate dimensions as they were the levels of a factor; that might be of interest for the future.

I'm fumbling in the dark a bit as the documentation for my brms install doesn't seem to work right. I'm looking at the pdf version on CRAN. Maybe I can install differently and get that.

paul-buerkner:
The example models have just a few samples and didn't really converge. It is not unlikely that things look that way in the mixture model of `fit2`. Also, without any uncertainty interval it is unclear hos substantial the changes are. So I have no reason to not trust the results based on these plots. I have also no reason to trust them on that basis but the code looks to be run correctly on the brms side.

Thank you for testing these models! This was really helpful! I am going to merge this now.

Issue #418, opened 2018-04-29
emmeans interface


rvlenth:
I was alerted to some discussion on interfacing to the **emmeans** package in the Google group. I was going to contribute to the discussion there, but (based on browser warnings) was unable to find a way to securely log in. So I will comment on a few things here, in hopes that it will clarify some things

> **emmeans** primarily does frequentist analysis (even for Bayesian models) and I don’t want to actively support that kind of analysis. 

I admit that **emmeans** places a lot of emphasis on frequentist methods, but I want to adequately support Bayesian approaches as well.

> Also, most of emmeans summaries are frequentist ... there is some mcmc functionality implemented, but more as a side track than as a main feature I feel.

Please enlighten me. My impression has been that Bayesians already have the tools they need in packages like **bayesplot**, **coda**, etc., to do what they want, once they get a posterior sample; and that I do provide. So can you give me an idea of what else I could provide? I’d be happy to consider additional or alternative methods that are needed to make it more appealing to Bayesians. Perhaps the `summary()` method should display something different. I can believe that; describe what you'd like to see.

> I have been reading the documentation of extending **emmeans**. The problem I see is that we can only support a very minor class of **brms** models with **emmeans**, which implies a lot of special case coding and checking.

Is this really true? Most of what **emmeans** does for Bayesian models is very, very simple in concept: It just calculates a posterior sample of predictions at each node in the reference grid, which the user may then access via `as.mcmc()`. I have hardly looked at this package yet, but I find it pretty difficult to believe that it would be hard to support a good share of the models, assuming they are based on linear or generalized linear models -- especially since you don't really want all those frequentist results like variances and covariances. All it really needs to do is create a model matrix for the reference grid. One of the real advantages of Bayesian methods is that you don't get into thorny issues like standard errors and degrees of freedom.  I will take a look and see what I can do easily; and report back.

rvlenth:
One very simple idea: In the `summary()` method for `emmGrid` objects, it could check and if there is a posterior sample, it displays estimates and HPD intervals for each case. Then users who fitted a Bayesian model would see that summary instead of the frequentist one. Would `coda::HPDinterval()` be a good choice?

paul-buerkner:
Hi Russel, thanks for the feedback! Don't misunderstand what I wrote as a critque to **emmeans**. It does a lot of cool stuff both frequentist and Bayesian. I know (and think have also written it) that you are supporting posterior samples, which is really all what Bayesian models need, as you point out. 

[...]

PR #898, opened 2020-05-14
Full support for 'emmeans'

paul-buerkner:
This PR enables full support of the emmeans package (#418) thanks to the help of @rvlenth.

@rvlenth would you mind running a few tests with the new code before I merge this PR?

Thanks again for your help!


rvlenth:
I'd be happy to test some things, but am not a power user of **brms**. I am curious to see if ordinal models are supported to the degree needed; same with hurdle and zip models. I'll see if I can muddle through some tests to see. 

Also, a kinda dumb question that shows my git cluelessness -- is there a way I can merge this PR with your changes and extensions into my own fork of brms? I can't really test what you have without doing that, or manually copying the changed files.

Any particular reason for suggesting emmeans >= 1.4.2?

Are there particular things you have in mind that you'd like to see me test?

paul-buerkner:
Thank you! ordinal and hurdle etc. models should be supported. You can use the `dpar` argument to select the chosen parameter if required for hurdle and similar models.

I am not a git expert either but I would expect it would make most sense if you would update your brms fork from my brms repo and then merge the emmeans branch of brms into yours. For the former part not sure how this is exactly realized.

No specific reason for 1.4.2 I guess that was just the one I appeared to have on my machine. Would you recommend a specific version requirement?

If you could run an ordinal and a hurdle model and check whether the results match your expectation, that would be perfect! I am not a regular emmeans user so I don't know what behavior you would expect in those cases.

rvlenth:
On github stuff, hmmm. There are only 7 files, so maybe the simplest thing is for me to just manually copy the code!

Very early report: I did this to get some examples:
```r
fitj <- rename_pars(brms:::brmsfit_examplej)      # for j = 1,2,...,6
```
and so far, doing `emmeans::ref_grid(fitj)` works only for `fit1` and `fit3`. Perhaps it'll be better when I upgrade to your version of the code.

paul-buerkner:
I just fixed some problems so that now all example fit models in brms can be run (perhaps by using the `resp`, `dpar` and `nlpar` arguments). 

You can also just install brms from github via `remotes::install_github("paul-buerkner/brms", ref = "emmeans")` so that you don't need to copy over any code.

Issue #891, opened 2020-05-06
Support "cmdstanr" as additional backend


paul-buerkner:
Due to restrictions of CRAN, rstan moves more slowely than the development of Stan itself. For this reasons, I think it would be good to support the cmdstanr as a second backend in brms

rok-cesnovar:
Any preference on how one would select the backend? Separate command or an argument?

paul-buerkner:
Argument, as brms has only one primary fitting function. 

Btw. what is the best way to transform the output of cmdstanr into a stanfit object? Not that I like stanfit objects in particular but it would make so much more code reusable.

rok-cesnovar:
Right now its using rstan's read csv function. `stanfit <- rstan::read_stan_csv(fit$output_files())` where fit is the return of $sample().

paul-buerkner:
Ok sounds easy enough. I will see if I can work on this laer on.

paul-buerkner:
Closed via #900

PR #900, opened 2020-05-16
Support cmdstanr as additional backend

paul-buerkner:
The PR implements cmdstanr as additional backend (#891). It should stay open until brms is CRAN releasable either because cmdstanr itself is on CRAN or if we found another way to make this possible (see https://github.com/stan-dev/cmdstanr/issues/168).


paul-buerkner:
Tagging @jgabry and @mitzimorris

jgabry:
Looks good!

paul-buerkner:
Although cmdstanr is not on CRAN yet, I will merge this PR into master now. If the next brms CRAN release needs to happen before cmdstanr gets on CRAN, I will either make my own package repo (https://github.com/stan-dev/cmdstanr/issues/168) or comment out the cmdstanr requiring functions for the release (which is easy due to the chosen modularization).

jgabry:
Awesome!

> I will either make my own package repo ([stan-dev/cmdstanr#168](https://github.com/stan-dev/cmdstanr/issues/168)) or comment out the cmdstanr requiring functions for the release (which is easy due to the chosen modularization).

Maybe by then we'll also have that Stan R package repository that @rok-cesnovar suggested on discourse

PR #914, opened 2020-06-01
import posterior_epred generic from rstantools

jgabry:
import `posterior_epred()` generic from rstantools v2.1.0, which was just accepted on CRAN


paul-buerkner:
Thank you!

PR #916, opened 2020-06-05
fix a bunch of minor typos in vignettes

jgabry:
These are all really minor and inconsequential typos, but I was reading through some of the vignettes and figured I might as well fix them.


paul-buerkner:
Thank you!

PR #944, opened 2020-07-02
CmdStanR updates

jgabry:
* Adds usage of new `show_messages` argument in CmdStanR's `$sample()` method
* Gets rid of deprecation warning from `cores` changing to `parallel_chains` in CmdStanR
* Use `cmdstanr::write_stan_tempfile()` for convenience
* Adds version requirement for CmdStanR (alternatively if you prefer we could add logic in the R code to not try to use `show_messages` for earlier versions of CmdStanR but that seemed messier) 

#### Example

```r
# no more informational messages from cmdstanr since silent defaults to TRUE
# and no deprecation warning from cores=4 (uses parallel_chains internally)
fit <- brm(count ~ zAge, data = epilepsy, backend = "cmdstanr", refresh = 0, cores = 4)

Start sampling
Running MCMC with 4 parallel chains...

Chain 1 finished in 0.2 seconds.
Chain 2 finished in 0.2 seconds.
Chain 3 finished in 0.1 seconds.
Chain 4 finished in 0.1 seconds.

All 4 chains finished successfully.
Mean chain execution time: 0.2 seconds.
Total execution time: 0.2 seconds.
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -35,7 +35,7 @@ Imports:
 Suggests: 
     testthat (>= 0.9.1),
     emmeans (>= 1.4.2),
-    cmdstanr,
+    cmdstanr (>= 0.0.0.9008),
     RWiener,
     rtdists,
     mice,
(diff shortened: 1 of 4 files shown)


jgabry:
Also added a few doc clarifications about `inits`, as well as some edits related to there being two backends now.

paul-buerkner:
Thank you! I also added you as a contributor, which was long overdue.

jgabry:
No problem. And thanks!

PR #1005, opened 2020-09-21
Support dpar = "mean" in emmeans

paul-buerkner:
This PR is a follow-up on #993 with some additional fixes. There are still some problems which need to be resolved together with @rvlenth before the PR can be merged.


paul-buerkner:
Closing as I was working from the wrong branch...

PR #1032, opened 2020-11-05
Collapse adjacent loops over same ranges

andrjohns:
A consequence of model generation in ```brms``` is that the models can generate with multiple loops that each iterate over the same indexes/ranges. This could become costly with larger datasets, and so there would be a (hopefully) decent benefit to collapsing these multiple loops into one.

This PR introduces a post-processing of the generated Stan model that collapses loops that iterate over the same indexes/ranges. The collapse only happens with loops that immediately follow one-another, so that there are no intermediate computations that subsequent loops depend on.

For example,  if we generate the Stan code from one of the models in the vignette:
```
make_stancode(formula = time | cens(censored) ~ age * sex + disease 
                    + (1 + age|patient), 
                    data = kidney, family = lognormal(),
                    prior = c(set_prior("normal(0,5)", class = "b"),
                              set_prior("cauchy(0,2)", class = "sd"),
                              set_prior("lkj(2)", class = "cor")))
```

It contains this set of adjacent loops over the same range:
```
    for (n in 1:N) {
      // add more terms to the linear predictor
      mu[n] += r_1_1[J_1[n]] * Z_1_1[n] + r_1_2[J_1[n]] * Z_1_2[n];
    }
    for (n in 1:N) {
    // special treatment of censored data
      if (cens[n] == 0) {
        target += lognormal_lpdf(Y[n] | mu[n], sigma);
      } else if (cens[n] == 1) {
        target += lognormal_lccdf(Y[n] | mu[n], sigma);
      } else if (cens[n] == -1) {
        target += lognormal_lcdf(Y[n] | mu[n], sigma);
      }
    }
  }
```

With the proposed changes, the code would instead generate to:
```
    for (n in 1:N) {
      // add more terms to the linear predictor
      mu[n] += r_1_1[J_1[n]] * Z_1_1[n] + r_1_2[J_1[n]] * Z_1_2[n];
    // special treatment of censored data
      if (cens[n] == 0) {
        target += lognormal_lpdf(Y[n] | mu[n], sigma);
      } else if (cens[n] == 1) {
        target += lognormal_lccdf(Y[n] | mu[n], sigma);
      } else if (cens[n] == -1) {
        target += lognormal_lcdf(Y[n] | mu[n], sigma);
      }
    }
  }
```

### DIFF
--- a/R/make_stancode.R
+++ b/R/make_stancode.R
@@ -271,7 +271,7 @@ make_stancode <- function(formula, data, family = gaussian(),
     scode_generated_quantities
   )
   
-  scode <- expand_include_statements(scode)
+  scode <- collapse_loops(expand_include_statements(scode))
   if (parse) {
     scode <- parse_model(scode, backend, silent = silent)
   }
--- a/R/misc.R
+++ b/R/misc.R
@@ -1008,3 +1008,80 @@ expect_match2 <- function(object, regexp, ..., all = TRUE) {
   }
   invisible(NULL)
 }
+
+# Get indices and ranges from loop statement
+get_loop_pars = function(loop_arg) {
+  spl1 = strsplit(loop_arg," in ")[[1]]
+  spl2 = strsplit(spl1[2],":")[[1]]
+  
+  strt_ind = spl2[1]
+  ind_mnr = gsub("\\s*for \\(","\\1", spl1[1])
+  ind_mjr = gsub("\\) \\{","\\1", spl2[2])
+  
+  return(c(ind_mnr,strt_ind,ind_mjr))
+}
+
+# Find line number of closing brace of loop, given starting line number
+get_encl_brace = function(start, code_lines) {
+  char_split = strsplit(code_lines[(start+1):length(code_lines)],"")
+
+  cnt = 1
+  for(i in 1:length(char_split)) {
+    for(j in 1:length(char_split[[i]])) {
+        chr = char_split[[i]][j]
+        cnt = cnt + ifelse(chr == "{", 1, ifelse(chr == "}", -1, 0))
+        if(cnt == 0) {
+          return(start + i)
+        }
+    }
+  }
+}
+
+# Get line numbers to remove to collapse adjacent loops
+loop_inds = function(index_minor, start_index, index_major, code_split) {
+  # Find line numbers for the opening and closing braces of loops
+  l_start = as.numeric(grep(paste0("for (", index_minor, " in ",
+                                   start_index,":", index_major, ") {"),
+                            code_split, fixed = T))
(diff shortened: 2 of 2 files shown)


paul-buerkner:
Thank you! I wasn't aware such kind of loops are costly. What is the reason? Do you have an example that shows speedup with collapsed loops?

andrjohns:
My intuition was that with larger values of ```N``` it would be less efficient to traverse loops of length ```N``` multiple times than to just have the one loop. It was mostly driven by some work in the Math library where we found that it was faster to extract the values and adjoints of a container in a single pass, rather than in separate calls.

But alas, benchmarking has shown me wrong. I used the example model:
```
brm(formula = time | cens(censored) ~ age * sex + disease 
               + (1 + age|patient), cores = 4,
               data = kidney, family = lognormal(),
               prior = c(set_prior("normal(0,5)", class = "b"),
                         set_prior("cauchy(0,2)", class = "sd"),
                         set_prior("lkj(2)", class = "cor")))
```

Across K replications of the ```kidney``` dataset, where K = 1, 2, 5, 10, 20, 50:
```
mult_dat = do.call("rbind", replicate(K, kidney, simplify = FALSE))
```

Which gives sample sizes of: 76,  152,  380,  760, 1520, 3800

The performance of the github version and this loop collapsing version were highly consistent (if a little worse).

The table below shows the runtime (in seconds) averaged across four chains:

| N | Github | Loops Collapsed |
| ------------- | ------------- | ------------- |
| 76  | 3.199  | 4.320  |
| 152  | 10.478  | 10.220  |
| 380  |  34.410 | 34.919  |
| 760  | 88.030  | 91.668  |
| 1520  | 864.672  | 872.310  |
| 3800  | 2171.908  | 2182.165  |

There goes that idea!

andrjohns:
Did you have any smaller projects/issues for ```brms``` that would be good for getting started in the package?

paul-buerkner:
Thank you for checking this in detail! Later on today, I will go through all the brms issues and mark some as "good first issue". Thank you for being willing to contributing to brms! This makes me really happy :-)

andrjohns:
Great! Looking forward to making some contributions!

paul-buerkner:
I have added a "good first issue" tag on a lot of issues. Not sure if all
of them are actually good first issues (it strongly depends on the
background of people) but at least they may give some inspiration. Before
starting to work on one of these issues, I think it is best to check in on
the issue to have some discussion of how to move the issue forward in the
best way possible.



Am Fr., 6. Nov. 2020 um 14:16 Uhr schrieb Andrew Johnson <
notifications@github.com>:

andrjohns:
Thanks Paul! I'll have a read over the issues

Issue #1039, opened 2020-11-09
Put adjacent-category logit calcs in log scale


andrjohns:
Similar to #1038, the adjacent-category ordinal family with a logit link would be able to handle significantly larger inputs if the calculations were performed on the log scale:

``` r
library(rstan)
#> Loading required package: StanHeaders
#> Loading required package: ggplot2
#> rstan (Version 2.21.2, GitRev: 2e1f913d3ca3)
#> For execution on a local, multicore CPU with excess RAM we recommend calling
#> options(mc.cores = parallel::detectCores()).
#> To avoid recompilation of unchanged Stan programs, we recommend calling
#> rstan_options(auto_write = TRUE)

acatlogits = "
  real acat_logit_lpmf(int y, real mu, real disc, vector thres) {
     int nthres = num_elements(thres);
     vector[nthres + 1] p;
     p[1] = 0.0;
     for (k in 1:(nthres)) {
       p[k + 1] = p[k] + disc * (mu - thres[k]);
     }
     p = exp(p);
     return log(p[y] / sum(p));
  }

  real acat_logit_lpmf_test(int y, real mu, real disc, vector thres) {
     int nthres = num_elements(thres);
     vector[nthres + 1] p = append_row(0, cumulative_sum(disc * (mu - thres)));
     return p[y] - log_sum_exp(p);
  }
"


stancode = "data{ real y_mean; } parameters{ real y; } model{ y ~ normal(y_mean,1); }"
stancode = paste0("functions{", acatlogits, "} ", stancode)

expose_stan_functions(stanc(model_code=stancode))

# Check equivalent results with standard values
## First category
acat_logit_lpmf(y=1,mu=0,disc=2,thres=c(-0.5,0,0.5))
#> [1] -2.006409
acat_logit_lpmf_test(y=1,mu=0,disc=2,thres=c(-0.5,0,0.5))
#> [1] -2.006409

## Mid category
acat_logit_lpmf(y=2,mu=0.2,disc=2,thres=c(-0.5,0,0.5))
#> [1] -1.26896
acat_logit_lpmf_test(y=2,mu=0.2,disc=2,thres=c(-0.5,0,0.5))
#> [1] -1.26896

## Last category
acat_logit_lpmf(y=4,mu=1.5,disc=2,thres=c(-0.5,0,0.5))
#> [1] -0.1329533
acat_logit_lpmf_test(y=4,mu=1.5,disc=2,thres=c(-0.5,0,0.5))
[...]

PR #1040, opened 2020-11-09
Use log-scale calculations for adjacent-category logit likelihoods

andrjohns:
As covered in #1039 in more depth, this PR puts the intermediate calculations for the adjacent-category logit onto the log scale:
```
  real acat_logit_lpmf(int y, real mu, real disc, vector thres) {
     int nthres = num_elements(thres);
     vector[nthres + 1] p = append_row(0, cumulative_sum(disc * (mu - thres)));
     return p[y] - log_sum_exp(p);
  }
```


paul-buerkner:
Looks good! Thank you!

PR #1041, opened 2020-11-10
switch threading vignettte to fix warmup parameters and extend text

wds15:
This PR runs the threading benchmark a lot faster by fixing the mass matrix and step size. This shuts off any warmup which has apparently blurred our benchmark a bit. In addition I am now visualizing to people the relativley large cost you catch when simply enabling threading - people need to be aware of that. Results may look a bit bad with this, but that is just how it is for this example. With fewer random effects things would look better - we could search for nicer parameter sets, but I don't quite see a need for that.


paul-buerkner:
Thank you!

PR #1074, opened 2021-01-23
backends: rstan: Support within-chain threading with Stan >= 2.25

hsbadr:
This adds support for within-chain threading in the new version of `rstan >= 2.25 | Stan >= 2.25`. It has been tested with https://github.com/stan-dev/rstan/pull/887 and https://github.com/stan-dev/math/pull/2261.

A new `rsran` option `threads_per_chain` has been added to control the within-chain number of threads (`threads$threads`):
```r
rstan::rstan_options(threads_per_chain = threads$threads)
```

If the model is compiled with threading support, the number of threads to use in parallelized sections _within_ an MCMC chain (e.g., when using the `Stan` functions `reduce_sum()` or `map_rect()`). The actual number of CPU cores used is `chains * threads_per_chain` where `chains` is the number of parallel chains. For an example of using threading, see [Reduce Sum: A Minimal Example](https://mc-stan.org/users/documentation/case-studies/reduce_sum_tutorial.html)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,7 @@
 to increase sampling efficiency thanks to Andrew Johnson. (#1017, #1053)
 * Enable `posterior_predict` for truncated continuous models
 even if the required CDF or quantile functions are unavailable.
+* Add support for within-chain threading with `rstan (Stan >= 2.25)` backend.
 
 ### Bug Fixes
 
--- a/R/backends.R
+++ b/R/backends.R
@@ -50,16 +50,6 @@ compile_model <- function(model, backend, ...) {
   message("Compiling Stan program...")
   if (use_threading(threads)) {
     if (rstan::stan_version() >= 2.25) {
-      if (!exists("rstan_threading")) {
-        message("\nrstan version ",
-                utils::packageVersion("rstan"),
-                " (Stan version ",
-                rstan::stan_version(), ")\n",
-                "Using threads_per_chain = ",
-                threads$threads,
-                " for within-chain threading.\n")
-        rstan_threading <<- TRUE
-      }
       threads_per_chain_def <- rstan::rstan_options("threads_per_chain")
       on.exit(rstan::rstan_options(threads_per_chain = threads_per_chain_def))
       rstan::rstan_options(threads_per_chain = threads$threads)
@@ -104,16 +94,6 @@ fit_model <- function(model, backend, ...) {
   # some input checks and housekeeping
   if (use_threading(threads)) {
     if (rstan::stan_version() >= 2.25) {
-      if (!exists("rstan_threading")) {
-        message("\nrstan version ",
-                utils::packageVersion("rstan"),
-                " (Stan version ",
-                rstan::stan_version(), ")\n",
-                "Using threads_per_chain = ",
-                threads$threads,
-                " for within-chain threading.\n")
-        rstan_threading <<- TRUE
-      }
       threads_per_chain_def <- rstan::rstan_options("threads_per_chain")
       on.exit(rstan::rstan_options(threads_per_chain = threads_per_chain_def))
       rstan::rstan_options(threads_per_chain = threads$threads)
--- a/R/backends.R
+++ b/R/backends.R
@@ -60,6 +60,8 @@ compile_model <- function(model, backend, ...) {
                 " for within-chain threading.\n")
         rstan_threading <<- TRUE
       }
+      threads_per_chain_def <- rstan::rstan_options("threads_per_chain")
+      on.exit(rstan::rstan_options(threads_per_chain = threads_per_chain_def))
       rstan::rstan_options(threads_per_chain = threads$threads)
     } else {
       stop2("Threading is not supported by backend 'rstan' version ",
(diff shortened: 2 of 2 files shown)


paul-buerkner:
Thanks for opening this PR! A couple of questions:

Why are we using `rstan_options`? Wouldn't it be more explicit to pass the information to `rstan::stan_model` thus mirroring the workflow in cmdstanr?

You called the rstan argument `threads_per_chain` but it is called simply `stan_threads`. Does it make sense to align these names across interfaces, given that cmdstanr is based on the new standard naming conventions?

NEWS.md
```diff
@@ -6,6 +6,7 @@
 to increase sampling efficiency thanks to Andrew Johnson. (#1017, #1053)
 * Enable `posterior_predict` for truncated continuous models
 even if the required CDF or quantile functions are unavailable.
+* Add support for within-chain threading with `rstan (Stan >= 2.25)` backend.
```
rok-cesnovar, line 9: This should be Stan >= 2.26 as the next release of rstan on CRAN will be with Stan 2.26. There will not be a 2.25 release. Also change that in the code.

hsbadr: Thanks! I'll update it once `Stan 2.26` is released or at least when the version number is bumped on the `develop` branch. Right now, the development version of `rstan` uses `Stan 2.25`.

hsbadr:
> Why are we using `rstan_options`? Wouldn't it be more explicit to pass the information to `rstan::stan_model` thus mirroring the workflow in cmdstanr?

The new option is independent from `rstan::stan_model` as it only sets the number of within-chain threads via `STAN_NUM_THREADS` environment variables that's implemented by `stan::math::init_threadpool_tbb()`. If `Stan` functions `reduce_sum()` or `map_rect()` are used, TBB will use all hardware threads by default which could be dangerous and is now forced to use a single thread by default via rstan::rstan_options(threads_per_chain = 1) unless the option is changed by the user.
 
> You called the rstan argument `threads_per_chain` but it is called simply `stan_threads`. Does it make sense to align these names across interfaces, given that cmdstanr is based on the new standard naming conventions?

I think `stan_threads` and `threads_per_chain` have different roles;  `stan_threads` enables the threading while `threads_per_chain` controls the maximum allowed parallelism via TBB global control in `Stan`/`math`.

paul-buerkner:
So the only way to enable and disable threading is a global option? I am unsure if this is a good approach user-interface wise. For brms, this means we need to set threads to whatever the user chose and then switch it back to the default that was set before to not have unintentional side effects. I am happy to edit the PR in this direction but first I want to figure out if the current rstan threading interface is how we want to have it going forward.

You are right with regard to `stan_threads` and `threads_per_chain`. I mixed them up.

rok-cesnovar:
I think rstan should have a threads per chain arg that handles the environment variable stuff for the user, just like cmdstanr. We will be moving away from this env. variable approach in the April release anyways. Will note that in the rstan PR.

hsbadr:
> So the only way to enable and disable threading is a global option? I am unsure if this is a good approach user-interface wise. For brms, this means we need to set threads to whatever the user chose and then switch it back to the default that was set before to not have unintentional side effects. I am happy to edit the PR in this direction but first I want to figure out if the current rstan threading interface is how we want to have it going forward.

I think that the global option is the most explicit way to deal with an environment driven setting, which should actually be addressed by removing `STAN_NUM_THREADS` from `stan/math`.


> I think rstan should have a threads per chain arg that handles the environment variable stuff for the user, just like cmdstanr. We will be moving away from this env. variable approach in the April release anyways. Will note that in the rstan PR.

Exactly what I was going to say. Though, for now, I don't like setting an environment variable behind the scenes without a global option that controls it. Fortunately, R `Sys.setenv()` only affects the child processes.

@paul-buerkner @rok-cesnovar Feel free to edit the PR here and in `rstan` as you see fit.

hsbadr:
To elaborate on my decision to use the `rstan` options, the current implementation in `math` will use all hardware threads if the model code uses `reduce_sum()` even if we didn't set the `threads_per_chain` argument. This should be ok if the model is created and run internally by `brms`, but it doesn't control the threads if it's called, for whatever reason, from an environment that does or doesn't have a different `STAN_NUM_THREADS` set. The current option forces it to a single thread unless a specific number of threads is requested internally by `brms` or explicitly by the user, if they decided to run the model using a different call.

I've also checked that `brms` will always restore the default setting.

The optimal solution would be, for example, to control the number of the threads by calling `stan::math::init_threadpool_tbb(int num_threads)`, which requires a change in `math`. Until then, I do vote for the current setup.

paul-buerkner:
Thanks. I am ok with this setup as a temporary option although I would prefer the interface to be similar/the same as the one of cmdstanr in the long run. For the brms user, it doesn't matter anyway, as their interface doesn't change. Are you ok with me removing the message from the new code. I think it is a little unnecessary and also comes with the global `<<-` assignment which I would like to avoid if not strictly required.

hsbadr:
> Are you ok with me removing the message from the new code. I think it is a little unnecessary and also comes with the global `<<-` assignment which I would like to avoid if not strictly required.

Sure, I've removed it. I was using those messages for debugging b/c initially `threads$threads` wasn't respected.

paul-buerkner:
Thank you! Merging this now.

Issue #1058, opened 2020-12-07
Feature: Don't use stored fit when model or data changes


martinmodrak:
Currently, when using the `file` the argument, the saved fit is always loaded. It could be useful to extend this feature to recognize when model and/or data changes and discard the stored fit if this is the case. This issue describes some design considerations for further discussion. This is based on using a handmade implementation of this for quite a while.

1) Backwards compatibility: The feature should probably use a new argument to avoid changing behavoir of the `file` argument. To me the most sense would be to add something like `file_options` / `file_format` / `file_contents` with allowed values `fit_only` (default) and `everything` (I don't think it makes sense to let users choose to store only data or only model code).

2) How to store model data? I think it is best to not save the full data but instead use something like the `digest` package to compute a hash and store it. I also wouldn't hash the full dataset as passed to the `data` argument, but rather base the refitting on the result of `make_standata`, which let's us easily ignore changes to irrelevant parts of the dataset (and includes all other data passed to the model e.g. via `stanvars`).

3) How to store model specification/code? There are two options: 
   a) We could store the generated Stan code (or its hash but code is small so easy to store/read). This is easy to implement and would always reliably signal a change. Problem is that with updates to `brms` there are often inconsequential changes to the Stan code - so far I've seen whitespace changes, a function that was never called added and a change from `x'` to `transpose(x)`. While the whitespace changes are easy to handle with some simple normalization procedure for the code stirng, I don't think we could reasonably avoid the latter two changes from triggering a refit. Also would probably not work well with the within-chain parallelization support that makes the Stan code change despite the model staying the same.
   b) We could collect all the relevant parameters to `brm` (formula, prior, stanvars, rescor, ....) and store those. This would add some maintanence burden (whenever a new parameter that influences the model is added, it would need to be added into the file cache code). It is also possible that there would be bugs or improvements/changes in brms that would mean that the model sampled actually changed despite the arguments to `brm` being exactly the same. Is it desriable or a problem if we don't trigger a refit in this case?

4) Large models. For large models it might be disadvantegous to store the data/model info in the same file as the fit as when model/data changes the user will unnecessarily load the whole fit only to discard it. It might make sense to store the info in a separate file. But I think here the added complexity does not outweigh the benefits.

What are you thoughts?

paul-buerkner:
Thank you for opening this issue! I may need a few more days before I respond but I haven't overlooked it.

martinmodrak:
[...]

PR #1088, opened 2021-02-06
Refitting model only when necessary

martinmodrak:
Adresses #1058. Additionally, I also check that the `algorithm` parameter is the same.

I am not sure about many of the naming choices and I am totally OK with changing the names of the functions/parameters/...


paul-buerkner:
Thank you a lot! I will take a look at your PR soon. 

There is a huge diff in the brm.R file that makes it hard to see what has changed. Can you tell why the diff appears to large despite the changes being (likely) minimal?

martinmodrak:
Ah, I forgot to set correct line-ending handling in git... Should be fixed now.

paul-buerkner:
I went through the PR, made some minor edits and moved functions to other places. I think the PR looks almost ready. The only thing I am hesitant about is that `brmsfit_needs_refit` is exported. Do we really need it to be exported at this point? I am asking because we may want to make changes to the interface later on as we may improve the functionality to take even more aspect of the model into account.

martinmodrak:
So I definitely don't insist on `brmsfit_needs_refit` being exported. The idea behind this is that it lets you debug _why_ is brms asking you to refit your model - and then act on it. I actually did need this from time to time when living with my homemade version of the feature - e.g. when there was a change after updating `brms` that swapped (if I recall correctly) `transpose(something)` for `something'`, I knew I don't need to refit. Or the code said my data change although I believed they shouldn't have which then led me to discover a bug elsewhere....  

Maybe we can achieve the same result by providing an example how to compare code/data by hand, without exposing a function and thus commit to a specific interface?

martinmodrak:
Also I realized I didn't build support for brm_mutliple - so will look into this as well....

paul-buerkner:
I agree the functionality of `brmsfit_needs_refit` could be helpful to users. Perhaps we can export it but mark it as `@internal` which should prevent this function to be shown in the documentation index. That way, I would feel safe with exporting it because we could still change its interface without deprecation notice as we did not made it "officially" user-facing. Would this appear as a sensible strategy to you as well?

paul-buerkner:
I want to release brms 2.15 soonish. I think it would be great if that feature made it into the release. Do you have time at some point to make the final additions or tell me what parts you think still need to be added?

martinmodrak:
Thanks for reminding me. I'll push this up on the priority queue. Is it OK if I have a (hopefully final) version on Thursday?

paul-buerkner:
Thank you! Thursday is great. I have a couple of other issues to address
before release anyway.

Martin Modrák <notifications@github.com> schrieb am Di., 2. März 2021,
08:11:

martinmodrak:
So it turns out supporting `brm_multiple` was not straightforward. A code review will definitely  be warranted. But I added support for saving fits when `combine = FALSE` (which is actually currently required for `file_refit = "on_change"` - I need to store all the fits separately, not just the combined fit to be able to check for modifications).

I though it might be useful to add a "mock" backend that would just return some premade stored fits, so that one could write integration tests that go through `brm` and `brm_multiple` without actually invoking Stan during testing. This way I could write tests to cover all the codepaths created by this change... What do you think? 

Alternatively, we could just not support this option for `brm_multiple` at the moment and resolve all of this later?

paul-buerkner:
Thank you for looking into this. Personally I think it would be safer right now to not support this feature in brm_multiple until we have an idea how to cleanly implement it without having to consider too many special cases. So I think we should add the new argument `file_refit` to `brm_multiple` but throw an error if set to `on_change` saying this will be implemented in the future. What do you think?

martinmodrak:
Makes sense. So I pushed my (hopefully working but not thoroughly tested) code for brm_multiple to a new branch and added an error signal here. I'll look into the possibility of having a mock backend as that could IMHO improve the testing situation a bit.

martinmodrak:
So from my side, this is ready to merge.

paul-buerkner:
Great! I will be merging this PR now. Thank you for working on this feature!

wds15:
Whow! I just discovered this feature and now I can write my prior as

```R
sd_inter <- 1.0
nb_prior <- prior(normal(0, sd_inter), class=Intercept) +
    prior(normal(0, 2), class=Intercept, dpar=shape)
```

then I run my brms fit with this magic thing where I plug in a `stanvars(sd_inter)` argument

... and finally I can refit the model with a different prior sd ala

```R
fit2 <- update(fit, stanvars=stanvar(2*sd_inter, "sd_inter"))
```

and the fit starts immediately!!! 

Is this super cool!

Just one comment: From the docs I thought that brms would refit, since it says "recompile when priors change", but that's not exactly true... anyway. This is really cool I must say... maybe plug this into a vignette somewhere as an example? 

I have nothing to report other than that this is a cool feature saving lots of my time in the future.

martinmodrak:
@wds15 I don't think any of the code I wrote should have affected how the example you mentioned executes, so either something else changed or there was some unintended consequence of the changes I made, which might be warrant some further investigation :-D

wds15:
oh... I found a bug which is infact a feature? Cool!

PR #1092, opened 2021-02-08
Check if `rstan` supports `threads_per_chain`

### LINKED ISSUE #1091: Check if `rstan` supports `threads_per_chain`
The following checks need to be updated to catch the cases where a development version of `Stan` is used while `rstan` doesn't support `threads_per_chain` (to set the within-chain number of threads) as well as the potential CRAN release of `StanHeaders 2.26` (check [this](https://github.com/stan-dev/rstan/issues/899#issuecomment-774779057)) that supports the current version of `rstan 2.21`

https://github.com/paul-buerkner/brms/blob/acb98f6c5c1791f8e15ae63d1e754b92fc1b5ef6/R/backends.R#L52

https://github.com/paul-buerkner/brms/blob/acb98f6c5c1791f8e15ae63d1e754b92fc1b5ef6/R/backends.R#L96

This should use `utils::packageVersion("rstan")` based on the actual version of `rstan` that supports `threads_per_chain`.

In general, this ends with a warning message, but the number of threads isn't controlled, if threading is enabled.

hsbadr:
Closes #1091. May also need to catch the development versions that don't support `threads_per_chain` (e.g., by checking if the option exists), but that's ok since they're likely used by experienced users, especially with threading and/or OpenCL.

### DIFF
--- a/R/backends.R
+++ b/R/backends.R
@@ -49,7 +49,7 @@ compile_model <- function(model, backend, ...) {
   args$model_code <- model
   message("Compiling Stan program...")
   if (use_threading(threads)) {
-    if (rstan::stan_version() >= 2.25) {
+    if (utils::packageVersion("rstan") >= 2.26) {
       threads_per_chain_def <- rstan::rstan_options("threads_per_chain")
       on.exit(rstan::rstan_options(threads_per_chain = threads_per_chain_def))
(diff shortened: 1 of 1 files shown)


paul-buerkner:
Thanks!

PR #1121, opened 2021-03-10
Add eemeans version check before registering methods .onLoad

msenn:
brms failed to install for users with older emmeans present because install.packages() by default does not update "Suggests" dependencies.


paul-buerkner:
Thanks! In which version was this supported in emmeans exactly? You write "> 1.4.2", so was it 1.4.3 then?

paul-buerkner:
As per emmeans NEWS file, the change was made in version 1.4.0 and I edited the PR accordingly. Thanks again for opening the PR. Merging now.

PR #1136, opened 2021-04-10
projpred: Pass `cvfun` to `init_refmodel`

hsbadr:
The `cvfun` function is hardcoded here while a variant of it is maintained in `projpred::init_refmodel`. For example, a change like https://github.com/stan-dev/projpred/commit/6310b19a96c1e85ffbc89ff56a051bc2dadbcb95 breaks the hardcoded function here. This PR lets `projpred::init_refmodel` defines `cvfun` from `brmsfit` objects if it isn't passed in `...` (i.e., `cvfun = NULL` by default). Moreover, it will pass a custom function if set by the user.

### DIFF
--- a/R/projpred.R
+++ b/R/projpred.R
@@ -73,6 +73,10 @@ cv_varsel.brmsfit <- function(object, ...) {
 #' @export
 get_refmodel.brmsfit <- function(object, newdata = NULL, resp = NULL, 
                                  folds = NULL, cvfun = NULL, ...) {
+  if (is.null(object$elapsed_time)) {
+    object$elapsed_time <- rstan::get_elapsed_time(object$fit)
+  }
+
   resp <- validate_resp(resp, object, multiple = FALSE)
   formula <- formula(object)
   if (!is.null(resp)) {
--- a/R/projpred.R
+++ b/R/projpred.R
@@ -121,21 +121,11 @@ get_refmodel.brmsfit <- function(object, newdata = NULL, resp = NULL,
     .extract_model_data(object, newdata = newdata, resp = resp, ...)
   }
   
-  # extract a list of K-fold sub-models
-  cvfun <- function(folds) {
-    cvres <- kfold(
-      object, K = max(folds),
-      save_fits = TRUE, folds = folds
-    )
-    fits <- cvres$fits[, "fit"]
-    return(fits)
-  }
-  
   # using default prediction functions from projpred is fine
   args <- nlist(
     object, data, formula, family, folds, dis,
     ref_predfun = NULL, proj_predfun = NULL, div_minimizer = NULL, 
-    cvfun = cvfun, extract_model_data = extract_model_data, ...
+    extract_model_data = extract_model_data, ...
   )
   do_call(projpred::init_refmodel, args)
 }
--- a/R/projpred.R
+++ b/R/projpred.R
@@ -72,7 +72,7 @@ cv_varsel.brmsfit <- function(object, ...) {
 #' @export get_refmodel
 #' @export
 get_refmodel.brmsfit <- function(object, newdata = NULL, resp = NULL, 
-                                 folds = NULL, ...) {
+                                 folds = NULL, cvfun = NULL, ...) {
   resp <- validate_resp(resp, object, multiple = FALSE)
   formula <- formula(object)
   if (!is.null(resp)) {
@@ -122,13 +122,15 @@ get_refmodel.brmsfit <- function(object, newdata = NULL, resp = NULL,
   }
   
   # extract a list of K-fold sub-models
-  cvfun <- function(folds) {
-    cvres <- kfold(
-      object, K = max(folds),
-      save_fits = TRUE, folds = folds
-    )
-    fits <- cvres$fits[, "fit"]
-    return(fits)
+  if (is.null(cvfun)) {
+    cvfun <- function(folds) {
+      cvres <- kfold(
+        object, K = max(folds),
+        save_fits = TRUE, folds = folds
+      )
+      fits <- cvres$fits[, "fit"]
+      return(fits)
+    }
   }
   
   # using default prediction functions from projpred is fine
--- a/R/projpred.R
+++ b/R/projpred.R
@@ -73,10 +73,6 @@ cv_varsel.brmsfit <- function(object, ...) {
 #' @export
 get_refmodel.brmsfit <- function(object, newdata = NULL, resp = NULL, 
                                  folds = NULL, cvfun = NULL, ...) {
-  if (is.null(object$elapsed_time)) {
-    object$elapsed_time <- rstan::get_elapsed_time(object$fit)
-  }
-
   resp <- validate_resp(resp, object, multiple = FALSE)
   formula <- formula(object)
   if (!is.null(resp)) {
@@ -142,7 +138,8 @@ get_refmodel.brmsfit <- function(object, newdata = NULL, resp = NULL,
   args <- nlist(
     object, data, formula, family, folds, dis,
     ref_predfun = NULL, proj_predfun = NULL, div_minimizer = NULL, 
-    cvfun = cvfun, extract_model_data = extract_model_data, ...
+    cvfun = cvfun, extract_model_data = extract_model_data,
+    fit_elapsed_time = elapsed_time(object), ...
   )
   do_call(projpred::init_refmodel, args)
 }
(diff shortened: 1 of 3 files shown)


paul-buerkner:
Thank you for opening this issue. I think `projpred` should not have to know about `brmsfit` objects at all. So I would prefer this function to be defined in brms and remove the brms specific code in projpred. Do you see any drawback of this approach? The hard-coding could be omitted by only definining the function in brms on an `is.null(cvfun)` if the user doesn't pass their own version.

hsbadr:
> Thank you for opening this issue. I think `projpred` should not have to know about `brmsfit` objects at all. So I would prefer this function to be defined in brms and remove the brms specific code in projpred. Do you see any drawback of this approach? The hard-coding could be omitted by only definining the function in brms on an `is.null(cvfun)` if the user doesn't pass their own version.

It's ok either way, provided that:
1.  All user inputs to `get_refmodel` are respected.
2. The number of arguments to `cvfun` is constant, or the fallback here `if is.null(cvfun)` is supported by `projpred` (currently, this isn't the case in the `workflow` branch).
3. The `refmodel` generated by `brms` is compatible with `projpred` (See https://github.com/stan-dev/projpred/pull/116).

So, either remove all `brmsfit` cases in `projpred` and handle them here or rely on `projpred::init_refmodel`, which is the easier solution as you don't have to change the code here for `projpred` compatibility.

paul-buerkner:
Thanks for your insights. @AlejandroCatalina what is your view on this?

AlejandroCatalina:
I agree with @paul-buerkner , the best course is to remove all `brmsfit` code from `projpred`. I just want to clarify that the reason that code exists is to give support until brms includes all the necessary bits! 

Specifically for `cvfun` support, I think it's best to return it to `init_refmodel` as argument from `get_refmodel.brmsfit` and deal with `NULL` as we currently do. I will gladly remove the hardcoded `cvfun` code from projpred.

R/projpred.R
```diff
@@ -121,11 +121,23 @@ get_refmodel.brmsfit <- function(object, newdata = NULL, resp = NULL,
...
+    cvfun <- function(folds) {
+      cvres <- kfold(
+        object, K = max(folds),
+        save_fits = TRUE, folds = folds
+      )
+      fits <- cvres$fits[, "fit"]
+      return(fits)
+    }
+  }
+
```
AlejandroCatalina: Does `brms::kfold` automatically check `options(mc.cores)` for parallelization? Anyway, we want to have control over that from projpred at least, so that passing `cvfun(..., cores = N)` uses only `N` cores.

hsbadr: 9c21a6340abd6eb7c68e4fe2b06977661761494c should allow `cvfun` to pass extra arguments to `kfold`. I'll defer to @paul-buerkner to make sure that `kfold` accepts `cores` argument and use it if not.

On a related note, the checks in `projpred` can be generic. I'll update https://github.com/stan-dev/projpred/pull/116.

paul-buerkner:
For what purpose do we need the `elapsed_time` thing? I wonder because the feature also needs to work with cmdstanr not only with rstan backend.

AlejandroCatalina:
We are adding running time estimates in projpred@workflow branch. Specifically, the time estimate to run projpred KFold CV depends on how long the model takes to train per fold (which we can estimate from the full training time).

paul-buerkner:
I see. But why is this to be stored in the brmsfit rather than the refmodel object?

AlejandroCatalina:
No reason, honestly. That was OP’s proposal I think. I don't have a very strong opinion against it as we already extend family objects in projpred as well, but I can see why it would be better in refmodel as well, so I’m okay with your opinion.

Best,
Alejandro

hsbadr:
> I see. But why is this to be stored in the brmsfit rather than the refmodel object?

Two reasons: 1) it belongs to the fit, and 2) it aligns well with the current development on `workflow` where the summary object extracts the `fit` from the `refmodel`. I initially added `fit_elapsed_time` slot to the `refmodel`, but that introduced more changes compared to the current implementation.

hsbadr:
To be specific, the difference is in `summary.vselsearch` in https://github.com/stan-dev/projpred/pull/116/commits/3206c839242a450b2885b4b66285ddf73871e303. It's ok for me either way.

paul-buerkner:
The problem of storing inside `fit` is that this may not be possible depending on what `fit` is. For example, if `fit` was an S4 object (as is `stanfit`) then we could neither use `$` not could we create a new slot on the fly. Accordingly, I think elapsed_time needs to be stored inside refmodel in a separate slot even if that leads to a little more verbose code.

If `elapsed_time` is not available, what should the default value be in this case?

hsbadr:
> The problem of storing inside `fit` is that this may not be possible depending on what `fit` is. For example, if `fit` was an S4 object (as is `stanfit`) then we could neither use `$` not could we create a new slot on the fly. Accordingly, I think elapsed_time needs to be stored inside refmodel in a separate slot even if that leads to a little more verbose code.

I think we're storing it in the parent object/list, not in the `stanfit` slot. But, again, it's ok either way. 

> If `elapsed_time` is not available, what should the default value be in this case?

How's that possible? Doesn't the `cmdstanr` store the fit elapsed time or have a method to retrieve it?

paul-buerkner:
Both of my comments go beyond brms. 

Not every model object for which `get_refmodel` exists will be `S3` so the projpred code that works generically should not rely on the ability of the model object to store something in `$elapsed_time`. 

Same with the default of `elapsed_time`. Not every engine will be able to provide this information so we need to agree on a value that `elapsed_time` is set on when not this information is not available.

With regard to `cmdstanr`, I will have to check if/how this info can be extracted. @rok-cesnovar do you know out of your mind where `cmdstanr` stores that model fitting time?

AlejandroCatalina:
I guess if there is no information available is okay to have NA. In the end we just want that to provide a time estimate of how long it’ll take to run projpred under several options, and it’s okay if some information is simply not available.

Best,
Alejandro

paul-buerkner:
I pushed a change to the PR to store the elapsed time in `fit_elapsed_time` again. Feel free to rename this argument to whatever you end up naming it in projpred (I took this name from the snippet you linked above). Let me know once this slot in projpred exists and I will merge this PR here once the new projpred is on CRAN (the brms main branch needs to be releasable at any time and should not depent on dev branches of other packages).

hsbadr:
> I pushed a change to the PR to store the elapsed time in `fit_elapsed_time` again. Feel free to rename this argument to whatever you end up naming it in projpred (I took this name from the snippet you linked above). Let me know once this slot in projpred exists and I will merge this PR here once the new projpred is on CRAN (the brms main branch needs to be releasable at any time and should not depent on dev branches of other packages).

@paul-buerkner @AlejandroCatalina Do we really have to add a `fit_elapsed_time` argument to `init_refmodel`? If that's ok, the default is `fit_elapsed_time = NA` then.

paul-buerkner:
It cannot be a slot in `fit` for the above reasons unfortunately.

paul-buerkner:
Perhaps it makes sense to split this PR. The `cvfun` stuff I can merge into brms right now because it works already with the current projpred as far as I understand. The `elapsed_time` will have to be merged later. @hsbadr would you agree to splitting this PR? That way, we can move more quickly at least on one aspect.

hsbadr:
> It cannot be a slot in `fit` for the above reasons unfortunately.

Ok. https://github.com/stan-dev/projpred/pull/116/commits/ffc0e2ca1aa68ed49204930ae0265fc21a9bd2e1 should do it. We need to update `init_refmodel` doc and merge https://github.com/stan-dev/projpred/pull/11 to `workflow` and update `develop`.

> Perhaps it makes sense to split this PR. The `cvfun` stuff I can merge into brms right now because it works already with the current projpred as far as I understand. The `elapsed_time` will have to be merged later. @hsbadr would you agree to splitting this PR? That way, we can move more quickly at least on one aspect.

Sounds good to me. Will you handle it?

paul-buerkner:
Yes, I will handle it.

AlejandroCatalina:
Just to make it clear, we don't need any further change in `projpred@master` to incorporate `cvfun` into `get_refmodel.brmsfit` now, so that's good to go. 

The `workflow` branch in `projpred` is still very experimental and will take some time before even making it into `develop`, so that part is not in a hurry.

hsbadr:
> Just to make it clear, we don't need any further change in `projpred@master` to incorporate `cvfun` into `get_refmodel.brmsfit` now, so that's good to go.
> 
> The `workflow` branch in `projpred` is still very experimental and will take some time before even making it into `develop`, so that part is not in a hurry.

Agreed. Cleaning up the `brmsfit`-specific code, including `cvfun`, from `projpred` applies to all branches though.

paul-buerkner:
The `cvfun` fix is now in brms master. Thank you all for working on projpred improvements!

PR #1139, opened 2021-04-13
projpred: Pass `fit_elapsed_time` to `init_refmodel`

paul-buerkner:
As discussed in #1136

### DIFF
--- a/R/projpred.R
+++ b/R/projpred.R
@@ -142,7 +142,8 @@ get_refmodel.brmsfit <- function(object, newdata = NULL, resp = NULL,
   args <- nlist(
     object, data, formula, family, folds, dis,
     ref_predfun = NULL, proj_predfun = NULL, div_minimizer = NULL, 
-    cvfun = cvfun, extract_model_data = extract_model_data, ...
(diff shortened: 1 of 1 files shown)


paul-buerkner:
Just a note that `brms:::elapsed_time` now also works with cmdstanr and defaults to `NULL` if no information is provided (not `NA`). I think `NULL` would also be a better default for that information in projpred.

paul-buerkner:
@fweber144 Do you know if this PR is still relevant for something that will be implemented in projpred in the future?

fweber144:
Yes, for projpred's `workflow` branch.

paul-buerkner:
@fweber144 Any updates on this?

fweber144:
I don't think the `workflow` branch will be merged into `master` (because `workflow` is quite outdated now), but that doesn't mean that some of its features couldn't be re-implemented in `master`. Especially the timings (for which this PR is relevant) might be something that @avehtari is interested in, right?

paul-buerkner:
I will close this PR for the time being. It's such a minor change we can do it directly on main whenever this feature makes it into projpred.

PR #1153, opened 2021-05-05
Tests for equivalence of `sratio()` and `cratio()` in case of symmetric distribution functions

fweber144:
This is the PR for enumeration points 4 and 5 of [this comment](https://github.com/paul-buerkner/brms/pull/1137#issuecomment-832596346) (see also [this comment](https://github.com/paul-buerkner/brms/pull/1137#issuecomment-832659592) for an explanation why I commented out the tests for the equivalence of `sratio()` and `cumulative()` in case of the cloglog link).

### DIFF
--- a/R/make_standata.R
+++ b/R/make_standata.R
@@ -112,7 +112,7 @@ make_standata <- function(formula, data, family = gaussian(), prior = NULL,
       tidy_meef(bterms, data), tidy_ranef(bterms, data)
     )
   }
-  structure(out, class = "standata")
+  structure(out, class = c("standata", "list"))
 }
 
 #' Extract data passed to Stan
--- a/R/misc.R
+++ b/R/misc.R
@@ -240,8 +240,8 @@ isNA <- function(x) {
   length(x) == 1L && is.na(x)
 }
 
-is_equal <- function(x, y, ...) {
-  isTRUE(all.equal(x, y, ...))
+is_equal <- function(x, y, check.attributes = FALSE, ...) {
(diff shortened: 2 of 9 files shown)


paul-buerkner:
Thanks for opening this PR! I would prefer these tests to live in `tests/local/tests.models_new.R` (or some new file in `/local` because all the example models are stored and shipped with brms thus further increasing the package size. Would that be possible?

fweber144:
Sure, I'll check that. I wasn't 100% sure if I put everything at the correct (or most appropriate) place.

Another remark: Commit d9266cc is just a leftover fragment from some of my local code. I thought that additional unit test could not harm (especially because it doesn't need a new fit, so it's computationally inexpensive), but if you don't want it, just remove it. In any case, it's not complete (as it only tests the `sratio()` family).

fweber144:
I now moved `brmsfit_example7` (and a copy of `brmsfit_example4`) as well as the corresponding test to `tests/local/tests.models_new.R`. Is it OK this way?

paul-buerkner:
Great! Can you remove the commented out cloglog tests please? I don't really want this to be in the code base at the moment even if commented out.

paul-buerkner:
I removed the commented-out tests and will merge this PR after checks are passing locally.

fweber144:
Thanks! Sorry for not replying—I was just away from my computer.

paul-buerkner:
no worries :-)

Am Do., 6. Mai 2021 um 16:29 Uhr schrieb Frank Weber <
***@***.***>:

PR #1159, opened 2021-05-14
Link functions, inverse-link function for `categorical()`, other preparations for augmented-data projection

fweber144:
This PR:

* adds the link functions requested [here](https://github.com/paul-buerkner/brms/pull/1137#issuecomment-829244236),
* adds the inverse-link as well as the link function for the `categorical()` family,
* performs some other preparations for the augmented-data projection planned for the projpred package,
* performs some minor cleaning of changes introduced by older PRs.

Notes/questions:

1. For the link functions for the `sratio()` and the `cratio()` family, the best solution I could come up with is a recursive one. I hope that's OK.
2. I encountered an issue for the `categorical()` family which I have resolved in [this line](https://github.com/fweber144/brms/blob/a66518e696b803b328a9c0c63a4537e7de9211a2/R/projpred.R#L104). However, I'm not sure whether the `multinomial()` and the `dirichlet()` family (and perhaps other families) need to be added in that `if` condition, too.
3. As mentioned in [these lines](https://github.com/fweber144/brms/blob/a66518e696b803b328a9c0c63a4537e7de9211a2/R/projpred.R#L93-L95), I do not re-define ordinal or categorical families in `get_refmodel.brmsfit()` to preserve their family-specific extra arguments like `refcat` and `thresholds`. Perhaps you want to avoid the re-definition also for other families (except `bernoulli()` which needs to be re-defined as `binomial()`, I guess) or also if `!isTRUE(dot_args$aug_data)` (because the families might gain extra arguments in the future and then the re-definition would be inappropriate)?
4. Currently, `R CMD check` gives the `NOTE`
    ```
    > checking dependencies in R code ... NOTE
    Missing object imported by a ':::' call: ‘projpred:::arr2augmat’
    ```
    That `projpred:::arr2augmat()` function does not exist yet in projpred (only in my augmented-data development branch). I hope this `NOTE` is OK. If not, we could comment the offending line for now and uncomment it as soon as the function exists in projpred.
5. For now, these are my changes I would consider as necessary for the augmented-data projection. In the future, further changes might be necessary but I can't think of any additional ones right now.


paul-buerkner:
Thanks a lot! This is a lot of code so it will take me some time to check all of it. 

Since this PR depends on a develop projpred version, I cannot merge it before this projpred version is on CRAN. So we either have to weight until this point or comment out anything that is not compatible with the current projpred on CRAN.

fweber144:
> Since this PR depends on a develop projpred version, I cannot merge it before this projpred version is on CRAN. So we either have to weight until this point or comment out anything that is not compatible with the current projpred on CRAN.

Then I would prefer to comment out [this line](https://github.com/fweber144/brms/blob/a66518e696b803b328a9c0c63a4537e7de9211a2/R/projpred.R#L137) containing the `projpred:::arr2augmat()` call. Since the ordinal families are currently not supported by projpred's CRAN version (and therefore raise an error), the user won't even get that far as to use that line.

fweber144:
Perhaps it would be safer to not only comment out that line, but to also add a temporary error so I don't forget to uncomment that line later.

paul-buerkner:
I agree.

Frank Weber ***@***.***> schrieb am Sa., 15. Mai 2021, 15:23:

paul-buerkner (approved):
Thank you so much! Looks good. Is there anything you want to change before I merge?

You added quite some new tests, which I appreciate. Just a heads up that I may need to skip some of them on CRAN later on as the CRAN team already complains about long testing times of brms.

paul-buerkner:
About the families: Right now, I overwrite them so they have additional info that projpred needs which brmsfamilies don't need and don't store. So for now, we need to stick to this approach but I am happy to rethink this as we move further with projpred.

fweber144:
Thanks for reviewing. There's nothing I want to change before you merge.

> You added quite some new tests, which I appreciate. Just a heads up that I may need to skip some of them on CRAN later on as the CRAN team already complains about long testing times of brms.

No problem. Even though the tests which are new in this PR shouldn't take long. I guess you were talking of tests with new `brm()` fits introduced by older PRs, right? (#1137 and #1153, if I looked it up correctly.)

paul-buerkner:
You are right, the new tests are actually quite quick so shouldn't be a problem for CRAN. Merging now.

PR #1160, opened 2021-05-16
man: Replace `\pkg{Stan}` with `\proglang{Stan}`

hsbadr:
Minor change.


paul-buerkner:
Thanks!

jgabry:
Just curious, does `\proglang{}` actually do anything? I just tried and it doesn't seem to have any affect on text and it's not listed in https://cran.r-project.org/doc/manuals/R-exts.html#Marking-text as a valid markup command.

paul-buerkner:
Hmm, I haven't checked but remember having seen it in other packages as well. It might as well do nothing. :-D

paul-buerkner:
@hsbadr How did you come up with this markup?

hsbadr:
> @hsbadr How did you come up with this markup?

It's used in `Rmd` files to distinguish between packages and programming languages.

Check [this](https://stat.ethz.ch/pipermail/r-help/2007-November/144810.html):
> If you are free to format your references to packages and functions as you
> please, you might follow the convention used by the Journal of Statistical
> Software.
> 
> you can do this by adding the following to your latex file:
> 
> \newcommand{\pkg}[1]{{\normalfont\fontseries{b}\selectfont #1}}
> \let\proglang=\textsf
> \let\code=\texttt
> 
> then you can refer to R as \proglang{R}, to packages with e.g.,
> \pkg{packagename} and to functions/other code as \code{functionname}.

@jgabry May need to add a new command, something like:
```latex
\newcommand{\proglang}[1]{\textsf{#1}}
```

@paul-buerkner You already use it at [here](https://github.com/paul-buerkner/Bayesian-IRT-paper/blob/ba82294a0ec92f9edc47fb956b8827bdbf272f81/Bayesian-IRT.tex#L13).

paul-buerkner:
I use it in a tex and rmd files converted to PDFs via latex but not in R doc (Rd) files I think.

hsbadr:
> Hmm, I haven't checked but remember having seen it in other packages as well. It might as well do nothing. :-D

It should do what's defined in the new command, but feel free to revert this if you don't prefer it.

paul-buerkner:
I think I would prefer to revert but out of interest, where would you put `\newcommand{\proglang}[1]{\textsf{#1}}` to work in all .Rd files?

jgabry:
Thanks for the explanation @hsbadr.  

> I think I would prefer to revert but out of interest, where would you put `\newcommand{\proglang}[1]{\textsf{#1}}` to work in all .Rd files?

That's what I'm still a bit confused about. Seems like a useful thing to do in a tex file but I don't see how that would impact how Rd files are rendered.

hsbadr:
> I think I would prefer to revert but out of interest, where would you put `\newcommand{\proglang}[1]{\textsf{#1}}` to work in all .Rd files?

> As from R version 3.2.0, User-defined macros may be given in a separate file using `\newcommand` or `\renewcommand`.

It could also be added at the top of the Rd file that uses the new command, but I don't know how to automatically add headers or append macros to the files generated by `roxygen2`.

paul-buerkner:
Thanks, I didn't know that. I don't think it's worth it for us going through this trouble at the moment. `pkg{}` is fine if though not entirely on point.

PR #1164, opened 2021-05-24
Fix CRAN NOTEs

hsbadr:
Here's the list of current NOTEs:
```r
Check: Rd cross-references
Result: NOTE
    Undeclared packages ‘shiny’, ‘gtable’ in Rd xrefs
```
```r
Check: package dependencies
Result: NOTE
    Package suggested but not available for checking: 'cmdstanr'
```
```r
Check: installed package size
Result: NOTE
     installed size is 6.5Mb
     sub-directories of 1Mb or more:
     R 3.1Mb
     doc 2.4Mb
```

The 1st one should be fixed by this PR.

@paul-buerkner Do you really need to suggest `cmdstanr` before it's released to CRAN?
I think the installed package size is hard to resolve.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -58,6 +58,8 @@ Suggests:
     digest,
     diffobj,
     R.rsp,
+    gtable,
+    shiny,
     knitr,
     rmarkdown
 Description: Fit Bayesian generalized (non-)linear multivariate multilevel models


paul-buerkner:
Thanks! linking to cmdstanr is necessarily unfortunately, but CRAN excepts it this way, so it's fine. Do you know where the xrefs to `shiny` and `gtable` are coming from? Perhaps we can also get rid of these links to avoid putting more stuff in suggests? Not a big deal to put stuff there of course, just want to make sure we need it.

hsbadr:
> Do you know where the xrefs to `shiny` and `gtable` are coming from?

`gtable` is referenced at:
https://github.com/paul-buerkner/brms/blob/a06d32f495d148e26bdc2cea0f0f8e3c131c6ff1/R/plot.R#L34

`shiny` is referenced at least at:
https://github.com/paul-buerkner/brms/blob/71fc8e4f0cd3aae9c8169155ca8117870d8f82b2/R/launch_shinystan.R#L13

To fix the NOTE, we could remove the references or suggest the referenced packages. CRAN notes that because documentation links will be broken otherwise (if those packages are referenced but not installed).

hsbadr:
> ```r
> Check: installed package size
> Result: NOTE
>      installed size is 6.5Mb
>      sub-directories of 1Mb or more:
>      R 3.1Mb
>      doc 2.4Mb
> ```

One way to reduce the installed package size, **only if desired**, is to convert it to a project and split the big components into a few modular packages: e.g., `brms` (main package), `brms.families` (custom families), `brms.vis` (visualizations), `brms.shiny`, ... etc. This isn't worth it just for the NOTE fix, but it could be a nice future plan if it's better to maintain each component separately and load the required components with `brms` (similar to `tidyverse`, `tidymodels`).

paul-buerkner:
Thanks. I currently don't want to go down the road of splitting up brms, but it may be something to consider in the future.

PR #1172, opened 2021-06-07
Update brmsformula.R

mbg-unsw:
Update documentation to reflect the fact that mixture models now allow truncation and censoring.


paul-buerkner:
Thanks!

PR #1173, opened 2021-06-07
add test that non-looped non-linear models 

wds15:
This test ensures that predictions for non-looped non-linear models are evaluated in a blocked way with the order of the data-set not being changed.


paul-buerkner:
Perfect, thank you!

PR #1210, opened 2021-08-06
Initial support for the posterior package

paul-buerkner:
This PR adds initial support for the posterior package. There is a lot of code that can/should be moved eventually and this PR should be seen more as a start to get going, with more to follow in the future.


paul-buerkner:
This PR is now basically ready, subject to some points that need discussion (see below). I ended up putting way more into this PR than originally planned so that now all posterior related stuff that does not (yet) break backwards compatibility is included. 

Some discussion points remain, for which I would like to hear a second opinion on. @mjskay would mind taking a look and providing your thoughts? Others' opinions are of course very welcome too!

1. `as_draws.brmsfit` and friends now support all arguments of `subset_draws`, not only `variable`. This clutters the interface a bit but I think it is preferable for the following reason: All the "standard" extractor methods such as `as.data.frame` are now also created using posterior methods internally. However, we cannot pipe them anymore in a subsequent `subset_draws` because they are no longer draws objects, but it would still be highly desirable to enable subsetting draws when using these methods. So for consistency, all posterior and standard extractor methods now internally also call `subset_draws` with full flexibility. Do you think this is an appropriate interface?

2. `plot.brmsfit` and friends used to have a `pars` arguments to only show selected paramters. `pars` is now deprecated in favor of `variable`. Would you agree that  that the name `variable`, as opposed to `variables`, is preferable also in this setting (see also point 3 below).

3. Some post-processing functions such as `posterior_predict` used to have a `subset` argument to selected a subset of posterior draws. `subset` is now deprecated because it was a terrible name and I now use `draw_ids` instead. Using the argument name `draw` would have been more consistent with point 2 above, but I had the feeling that `draw` was not specific enough in the context of a general purpose post-processing functions while `draw_ids` is a quite informative name.

mjskay:
Exciting to see this coming! I'm busy the next two days or so but I will take a look later this week.

paul-buerkner:
Thank you! @mjskay would you have time to check my questions this week? I am planning on submitting the a new brms version to CRAN and of this week, which I think is long overdue, and I would love to include the posterior PR in the release.

mjskay:
Sure, taking a look now. I think points 2 and 3 seem reasonable (or at least I don't have strong opinions on).

For point 1, it seems like a trade-off between internal consistency of the APIs within brms and external consistency with posterior and/or other packages. I'll admit to a bias towards external consistency, so let me first make the case against merging the two interfaces in as_draws_XXX. 

The main disadvantage I see of mixing the subset_draws and as_draws_XXX interfaces together in brms is that it will make users expect it in other packages, which then means other package developers will probably also want to implement it (or, users will be confused if they are moving between brms and another package that does not implement it). So, from the {posterior} perspective, I'd make the decision based on whether or not this is how we want the interface to look, and if we do we might want to make that clear in any guidance we have for package implementors (incidentally, it might be nice for us to put such guidance into a vignette in posterior).

Since it's not clear to me that users would often move between as.data.frame and as_draws_df (rather, I would expect them to move from one to the other if they want to use posterior, but not back and forth), I think I am less worried about users getting confused about whether or not some parameter is as.array / as.data.frame but not as_draws_XXX.

All that said, there's another argument I am more fond of, which is that merging the two interfaces is useful if subsetting prior to converting between types will be faster. It looks like you take some advantage of this fact in the implementation.

Thus, I *think* I can be convinced it's a good idea, but that whatever lessons are learned in doing it we should document in a vignette in posterior describing how we think these interfaces can/should be implemented. I'd prefer something with less implementer burden than the merged interface creates, but given the potentially large speed improvement in big models of subsetting variables prior to type conversion, it's probably a necessary thing.

paul-buerkner:
Thank you very much for your input! Yes, allowing the subsetting of variables will indeed be a huge potential speed up, but subsetting draws etc. may not or only less so. So I will, for now, only allow for subsetting `variable` in `as_draws_XXX` but allow for all subsetting in `as.data.frame` etc. We can easily extend this to `as_draws_XXX` later if we decide that is a good idea to support in general and externally consistent across packages.

mjskay:
Cool, sounds good to me!

PR #1213, opened 2021-08-11
doc updates for custom families

jsocolar:
This is a minor doc update to explain how to pass multiple vint or vreal terms and to further clarify the impact of setting `custom_family(type = )`.  I have not yet run roxygenise.  

While I think this is the only relevant place in the doc for this information, a potential problem is that it is not easily discoverable.  For example `?custom_family` provides a link to the doc for `stanvars`, but there's not trail of links to get through to the doc for `?brmsformula`.

Additionally, there are two things that I'd like to doc but I'm not sure where to put the doc:
* The important distinction between `custom_family(..., vars = vint1)` and `custom_family(..., vars = vint1[n])` that was a stumbling block for both me and @wds15.
* How multiple `vint` or `vreal` terms get matched to the arguments of the lpdf or lpmf.  (Not only am I unsure of where to document this; I am unsure of what the answer even is!)

### DIFF
--- a/R/brmsformula.R
+++ b/R/brmsformula.R
@@ -331,10 +331,12 @@
 #'   For custom families, it is possible to pass an arbitrary number of real and
 #'   integer vectors via the addition terms \code{vreal} and \code{vint},
 #'   respectively. An example is provided in
-#'   \code{vignette('brms_customfamilies')}.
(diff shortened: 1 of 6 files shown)


paul-buerkner:
Thanks! I added some more details to the doc, hopefully answering your questions above. Merging this now.

PR #1234, opened 2021-09-10
`get_refmodel.brmsfit()` docs: ignored arguments

### LINKED ISSUE #1230: `extract_model_data()`: ignored arguments
The `extract_model_data()` function used internally by `get_refmodel.brmsfit()` ignores arguments `wrhs`, `orhs`, and `extract_y`. This is relevant for `projpred:::predict.refmodel`, `projpred::proj_linpred()`, and `projpred::proj_predict()`.

It might be worth discussing this with @AlejandroCatalina, i.e., if we want the behavior of those projpred functions to be the same for `"stanreg"` and `"brmsfit"` fits and, if yes, how that common behavior should look like. But until that is decided, I think it would be good to document this in brms. I can create a PR for this if you want.

fweber144:
This fixes #1230: In the `get_refmodel.brmsfit()` docs, details concerning ignored arguments of `extract_model_data()` are added.

### DIFF
--- a/R/projpred.R
+++ b/R/projpred.R
@@ -16,6 +16,11 @@
 #' @param ... Further arguments passed to 
 #' \code{\link[projpred:get-refmodel]{init_refmodel}}.
 #' 
+#' @details Note that the \code{extract_model_data} function used internally by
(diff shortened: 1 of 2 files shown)


paul-buerkner:
Thank you!

Issue #1237, opened 2021-09-16
validate_car_matrix() binary check removes sparsity


apeterson91:
Line 613 in the [formula-ac.R](https://github.com/apeterson91/brms/blob/master/R/formula-ac.R) file has a line that creates a non-sparse matrix from a previously sparse matrix. This defeats the purpose of using a sparse matrix in the function.  

I'm working on a fix that would keep M sparse but I wanted to document here in case someone more 
familiar with the [Matrix](https://cran.r-project.org/web/packages/Matrix/index.html) package spots an elegant solution first. Reprex below. See also this [documentation](https://stat.ethz.ch/R-manual/R-devel/library/Matrix/html/lsparseMatrix-classes.html) from the Matrix R package.

``` r
##validate_car_matrix <- function(M){ 
## .
## .
## .
## is sparse matrix
M <- Matrix::Matrix(rbinom(n = 100^2,size = 1,p=1E-3),nrow = 100, ncol = 100)
## Creates non sparse matrix 
not_binary <- !(M == 0 | M == 1)

still_sparse <- (M>0)

still_sparse[1:5,1:5]
#> 5 x 5 sparse Matrix of class "lgCMatrix"
#>               
#> [1,] . . . . .
#> [2,] . . . . .
#> [3,] . . . . .
#> [4,] . . . . .
#> [5,] . . . . .
not_binary[1:5,1:5]
#> 5 x 5 Matrix of class "lgeMatrix"
#>       [,1]  [,2]  [,3]  [,4]  [,5]
#> [1,] FALSE FALSE FALSE FALSE FALSE
#> [2,] FALSE FALSE FALSE FALSE FALSE
#> [3,] FALSE FALSE FALSE FALSE FALSE
#> [4,] FALSE FALSE FALSE FALSE FALSE
#> [5,] FALSE FALSE FALSE FALSE FALSE

## } 
```

<sup>Created on 2021-09-15 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

paul-buerkner:
Thank you for posting this issue and working on a fix!

PR #1239, opened 2021-09-16
Fix binary check in validate_car_matrix()

apeterson91:
Fixes this [issue](https://github.com/paul-buerkner/brms/issues/1237).


paul-buerkner:
Thanks! Why did you use unique on the non zero values? Doesn't that change the length of the vector in such that `not_binary` may have smaller length than `M@x`?

apeterson91:
Yes, that's an error on my part. Let me know if you want me to issue a new pull request or if it is easier to edit on your end.

paul-buerkner:
Thank you! pushed the changes directly to master now.

PR #1269, opened 2021-11-25
Corrected passthrough of threads for VI

stuvet:
Variational inference with threading failed with `cmdstanr` since the `threads` argument was misnamed. 

Altered naming of the threads argument for matching algorithms, but left unmatching algorithms to fall through to the later `stop` command. You may prefer to handle the unmatching algorithms here too, or to provide a default via `else`.


paul-buerkner:
Thank you!

PR #1287, opened 2022-01-27
Quick fix for issue #1286 (and other changes for K-fold CV)

fweber144:
This is a "quick" fix for issue #1286 because it resolves the error described in #1286 (due to `allow_new_levels`), but it doesn't resolve the inconsistency between the reference model's linear predictors and the submodels' linear predictors described in stan-dev/projpred#268.

The first three commits (d47829399fb55015ad46ef9f69a575d6a9d740a4, ddcece60c8eda555087e8b4a1e17d2f3d44ffe3f, db89c178406339ae68124357fad50cade8da88d6) are not really related to the `allow_new_levels` issue; I stumbled across them while working on this. To see the issue of attaching brms (solved by commit db89c178406339ae68124357fad50cade8da88d6), consider the following reprex:
```r

# Data --------------------------------------------------------------------

set.seed(457211)
dat <- matrix(rnorm(41 * 2, sd = 0.4),
              ncol = 2,
              dimnames = list(NULL, paste0("X", 1:2)))
dat <- data.frame(dat)
dat$group <- gl(n = 8, k = floor(nrow(dat) / 8), length = nrow(dat),
                labels = paste0("gr", seq_len(8)))
group_icpts_truth <- rnorm(nlevels(dat$group), sd = 6)
group_X1_truth <- rnorm(nlevels(dat$group), sd = 6)
icpt <- -4.2
dat$y <- icpt +
  group_icpts_truth[dat$group] +
  group_X1_truth[dat$group] * dat$X1
dat$y <- rnorm(nrow(dat), mean = dat$y, sd = 4)

# Fit reference model -----------------------------------------------------

options(mc.cores = parallel::detectCores(logical = FALSE))
bfit <- brms::brm(y ~ X1 + X2 + (1 + X1 | group),
                  data = dat,
                  control = list(adapt_delta = 0.9),
                  refresh = 0,
                  seed = 1140350788)

# Refit in K folds --------------------------------------------------------

Kcv <- 2
set.seed(568456) # Needed to make the construction of the CV folds reproducible.
cvfits_grp <- loo::kfold(bfit,
                         K = Kcv,
                         folds = "grouped",
                         group = "group",
                         save_fits = TRUE,
                         seed = 4910396)
```
which reports the following to the console:
```
Fitting model 1 out of 2
Loading required package: Rcpp
Loading 'brms' package (version 2.16.3). Useful instructions
can be found by typing help('brms'). A more detailed introduction
to the package is available through vignette('brms_overview').

[...]

PR #1290, opened 2022-01-31
Fix PR #1287

fweber144:
Sorry, I just realized that I have overseen something while making the switch from `kfold_seed` and `refprd_seed` to `brms_seed` in PR #1287. I did search for remaining occurrences of `kfold_seed` and `refprd_seed`, but that didn't prevent me from overseeing that more of them need to be replaced. And since that `refprd_seed` remnant got swallowed up by `projpred::init_refmodel()`'s ellipsis, it didn't produce an error in brms's `R CMD check`.


paul-buerkner:
Thank you for spotting this issue!

PR #1292, opened 2022-02-02
Support augmented-data projection in projpred

fweber144:
This PR should perform the final changes which are necessary to support the augmented-data projection in projpred. Note that I partly reverted some of my changes in `get_refmodel.brmsfit()` introduced earlier (by PR #1159). The reason is that I found a more elegant way of handling the preparations for the augmented-data projection, namely by using a wrapper function for the user-specified `ref_predfun` within **projpred**. (Btw, this wrapper also handles the subtraction of the offsets, which is why we didn't have to subtract offsets in the brms-specific `ref_predfun` in PR #1287.) Sorry for the changes in `get_refmodel.brmsfit()` introduced earlier which are now partly reverted. But I think the new handling is much better, especially because it requires much less changes to **brms** and because it avoids exporting **projpred** functions which would only be needed by **brms**.

The development version of **projpred** currently does not support the augmented-data projection yet. If you try to create a reference model with the `categorical()` or an ordinal family such as `cumulative()` (in order to perform an augmented-data projection), **projpred** will throw an appropriate error (not only the development version, but also **projpred**'s CRAN version). That's also why I did not create a `NEWS` entry here in this PR.

I noticed that `get_refmodel.brmsfit()` has argument `newdata`. Currently, I can't think of a situation where this would be necessary, but I left it in case there is indeed a reasonable use case. However, if there are no use cases, this argument might confuse users, so then it might be worth removing it.

I noticed that you recently replaced `ilink` by `inv_link` (in all of the code, it seems). Some time ago, I decided to use the suffix `ilink` in projpred, so with this PR, you get a new occurrence of `ilink` which, however, should not be replaced by `inv_link` (otherwise, we would have to change **projpred** code).

### DIFF
--- a/R/distributions.R
+++ b/R/distributions.R
@@ -1960,7 +1960,7 @@ dcategorical <- function(x, eta, log = FALSE) {
   if (length(dim(eta)) != 2L) {
     stop2("eta must be a numeric vector or matrix.")
   }
-  out <- inv_link_categorical(eta, log = log)
+  out <- inv_link_categorical(eta, refcat_obj = NULL, log = log)
   out[, x, drop = FALSE]
 }
 
@@ -1972,10 +1972,11 @@ dcategorical <- function(x, eta, log = FALSE) {
 #   dcategorical()) or an array (S x N x `ncat` or S x N x `ncat - 1` (depending
 #   on `refcat_obj`)) containing the same values as the matrix just described,
 #   but for N observations.
-# @param refcat_obj Either NULL or an object of class "brmsfamily". If NULL, `x`
-#   is not modified at all. If an object of class "brmsfamily", then
-#   insert_refcat() is used to insert values for the reference category into
-#   `x`.
+# @param refcat_obj Either the string "first", an object of class "brmsfamily",
+#   or NULL. If "first", then the first category is used as reference and
+#   corresponding values are inserted into `x`. If an object of class
+#   "brmsfamily", then passed to insert_refcat() which is used to insert values
+#   for the reference category into `x`. If NULL, `x` is not modified at all.
 # @param log Logical (length 1) indicating whether to log the return value.
 # 
 # @return If `x` is a matrix, then a matrix (S x `ncat`, with S denoting the
@@ -1983,8 +1984,10 @@ dcategorical <- function(x, eta, log = FALSE) {
 #   categories) containing the values of the inverse-link function applied to
 #   `x`. If `x` is an array, then an array (S x N x `ncat`) containing the same
 #   values as the matrix just described, but for N observations.
-inv_link_categorical <- function(x, refcat_obj = NULL, log = FALSE) {
-  if (!is.null(refcat_obj)) {
+inv_link_categorical <- function(x, refcat_obj = "first", log = FALSE) {
+  if (identical(refcat_obj, "first")) {
+    x <- insert_refcat(x, family = categorical()) # The link does not matter.
+  } else if (!is.null(refcat_obj)) {
     x <- insert_refcat(x, family = refcat_obj)
   }
   if (log) {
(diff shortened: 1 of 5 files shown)


paul-buerkner:
Thank you! Reviewing this now.

As for the `newdata` argument. This allows to do projpred selection on entirely new data, or subsets of the original data, which may sometimes help with computation time. Or am I overlooking something?

R/projpred.R
```diff
@@ -122,53 +116,38 @@ get_refmodel.brmsfit <- function(object, newdata = NULL, resp = NULL,
...
+    # Also use argument `incl_thres` of posterior_linpred():
+    ref_predfun <- function(fit, newdata = NULL) {
+      # Setting a seed is necessary for reproducible sampling of group-level
+      # effects for new levels:
+      if (exists(".Random.seed", envir = .GlobalEnv)) {
+        rng_state_old <- get(".Random.seed", envir = .GlobalEnv)
+        on.exit(assign(".Random.seed", rng_state_old, envir = .GlobalEnv))
+      }
+      set.seed(refprd_seed)
+      posterior_linpred(fit,
```
paul-buerkner, line 145: here, you don't transpose via `t()`. Is this intentional?

paul-buerkner (commented):
Looks good. Just have two minor comments/questions.

fweber144:
> As for the `newdata` argument. This allows to do projpred selection on entirely new data, or subsets of the original data, which may sometimes help with computation time. Or am I overlooking something?

At first, I also thought this would be `newdata`'s purpose, but wouldn't the reference model have to be refitted if the data changes? As far as I can tell from quickly inspecting the source code of `current_data()`, such a refit does not take place.

R/distributions.R
```diff
@@ -1960,32 +1960,35 @@ dcategorical <- function(x, eta, log = FALSE) {
...
 # @return If `x` is a matrix, then a matrix (S x `ncat`, with S denoting the
 #   number of posterior draws and `ncat` denoting the number of response
 #   categories) containing the values of the inverse-link function applied to
 #   `x`. If `x` is an array, then an array (S x N x `ncat`) containing the same
 #   values as the matrix just described, but for N observations.
-inv_link_categorical <- function(x, insert_refcat_fam = NULL, log = FALSE) {
-  if (!is.null(insert_refcat_fam)) {
-    x <- insert_refcat(x, family = insert_refcat_fam)
+inv_link_categorical <- function(x, refcat_obj = "first", log = FALSE) {
+  if (identical(refcat_obj, "first")) {
```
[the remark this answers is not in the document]
fweber144: Thanks, I'll replace it.

R/projpred.R
fweber144: Yes, I decided to only accept S x N x C_lat arrays (S: number of draws, N: number of observations, C_lat: number of latent response categories, i.e., on linear predictor scale) as output of `ref_predfun` in case of the augmented-data projection. That avoids an unnecessary permutation of the array margins.

fweber144: I documented this in **projpred**, so I users should notice this.

R/distributions.R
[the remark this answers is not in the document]
fweber144: Done (commit 46ebaa6).

paul-buerkner:
The reference model could serve as a well predicting model also for new data so I see no reason not to have a `newdata` argument.

R/projpred.R
paul-buerkner: Thanks!

paul-buerkner:
Thank you for this PR! Looks good to me so merging now.

fweber144:
> The reference model could serve as a well predicting model also for new data

Yes, but if it's only for prediction, then there's the `projpred::predict.refmodel()` method which also has a `newdata` argument. What I'm not sure about is whether the creation of a reference model where fit and data don't match complies with **projpred**'s understanding of a reference model. I have to think about this again and maybe try to construct a use case. Currently, I still think a refit would be necessary, but I might be missing something. In any case, thank you for your feedback!

fweber144:
> Thank you for this PR! Looks good to me so merging now.

Thanks for merging!

PR #1294, opened 2022-02-04
Fix PR #1292

fweber144:
I'm really sorry, but I already found two bugs for the `categorical()` family in my code from PR #1292. This PR fixes these bugs (see the commit messages for details). The new common `ref_predfun` definition also avoids redundancies (in fact, these redundancies introduced the `t()` bug).


paul-buerkner:
Thank you!

Issue #1274, opened 2021-12-19
logistic normal distribution - feature request


bachlaw:
Thank you for all the tremendous work here.

As compositional-data guru John Aitchison has noted, the Dirichlet distribution tends to be useful in theory but not very often in practice, due to its unrealistic assumption that multiple events which together form a simplex somehow have no correlation whatsoever with one another, even when transformed off the sum-to-1 constraints.

One solution to this problem is the logistic normal distribution, whereby the model becomes a multivariate normal regression with the intercepts as logits summing to 1 under a simplex, and the various components then gain a correlation matrix.  In addition to making compositional data analysis much more meaningful, it can offer a workaround for multinomial models that prefer to recognize correlation among outcomes but wish to avoid the dreaded multinomial probit, as long as the modeler is willing to recast the various counts as compositions of a whole, with weights reflecting the number of "trials" for each row. The logistic normal distribution seems to come up most often in connection with latent dirichlet analysis and especially correlated topic modeling, but it has independent and more straightforward usefulness in the compositional data context. (The logistic normal's little brother, the logit-normal, can be a more flexible alternative to the beta distribution.)

As it is, the `mvbind` framework seems fully able to handle the multivariate normal nature of the outcomes, even with weights.  Thus, it seems like it would only require a few lines of additional Stan code to add this capability to brms's existing multivariate model structure, because we are talking about adding transformed parameters, not a whole new density function (https://discourse.mc-stan.org/t/how-do-i-specify-a-logit-normal-distribution/4894). Having said that, I am struggling a bit at how that code would end up looking, much less how it would be coded most efficiently.

So, while recognizing time is limited and there are no doubt plenty of requests for various distributions / families to be supported, I think this one would be a very worthy addition.

paul-buerkner:
I agree with the idea. Technically, for brms, this won't be a multivariate model (via `mvbind`) though but rather be a special univariate model with multi column response, as is dirichlet.

Let's discuss the name of this family. Logistic normal is usually understood as the "univariate" case as is log normal. How about "softmax_normal" or is there another canonical name?

bachlaw:
> I agree with the idea. Technically, for brms, this won't be a multivariate model (via `mvbind`) though but rather be a special univariate model with multi column response, as is dirichlet.
> 
[...]

PR #1295, opened 2022-02-06
Implement the logistic-normal family

paul-buerkner:
This PR implementes feature #1274. It still needs some work on priors, docs, and tests before merging but should be ready soon.


paul-buerkner:
Merging now.

PR #1304, opened 2022-02-27
Canonicalize model code for CmdStan >= 2.29.0

### LINKED ISSUE #1301: New warnings from latest Stan version
Hello, love brms, thanks for putting in all the effort writing and maintaining it. Just wanted to point out that when using the `cmdstanr` backend with the latest version of Stan on a model Stan produces the warning:

```
Declaration
    of arrays by placing brackets after a variable name is deprecated and
    will be removed in Stan 2.32.0. Instead use the array keyword before the
    type. This can be changed automatically using the auto-format flag to
    stanc
```

I'm sure this is already on your RADAR, but just in case, it seems like Stan has changed its preference for how you declare arrays.

rok-cesnovar:
See inline comments for a bit of detail. 

I didn't add any tests yet, will wait if you are fine with this first, I used the following to test this:

```r
library("brms")
data("kidney", package = "brms")
head(kidney, n = 3)

fit1 <- brm(formula = time | cens(censored) ~ age * sex + disease 
            + (1 + age | patient), 
            data = kidney, family = lognormal(),
            backend = "cmdstanr")
```

This produces warnings with 2.29, but not on the PR branch.

Fixes #1301

### DIFF
--- a/R/backends.R
+++ b/R/backends.R
@@ -25,7 +25,7 @@ parse_model <- function(model, backend, ...) {
   require_package("cmdstanr")
   temp_file <- cmdstanr::write_stan_file(model)
   if (cmdstanr::cmdstan_version() >= "2.29.0") {
-    .canonicalize_stan_model(temp_file, overwrite_file = FALSE)
+    .canonicalize_stan_model(temp_file, overwrite_file = TRUE)
   }
   out <- eval_silent(
     cmdstanr::cmdstan_model(temp_file, compile = FALSE, ...),
@@ -98,7 +98,7 @@ compile_model <- function(model, backend, ...) {
   args <- list(...)
   args$stan_file <- cmdstanr::write_stan_file(model)
   if (cmdstanr::cmdstan_version() >= "2.29.0") {
-    .canonicalize_stan_model(args$stan_file, overwrite_file = FALSE)
+    .canonicalize_stan_model(args$stan_file, overwrite_file = TRUE)
   }
   if (use_threading(threads)) {
     args$cpp_options$stan_threads <- TRUE
--- a/R/backends.R
+++ b/R/backends.R
@@ -24,6 +24,9 @@ parse_model <- function(model, backend, ...) {
 .parse_model_cmdstanr <- function(model, silent = 1, ...) {
   require_package("cmdstanr")
   temp_file <- cmdstanr::write_stan_file(model)
+  if (cmdstanr::cmdstan_version() >= "2.29.0") {
+    .canonicalize_stan_model(temp_file, overwrite_file = FALSE)
+  }
   out <- eval_silent(
     cmdstanr::cmdstan_model(temp_file, compile = FALSE, ...),
     type = "message", try = TRUE, silent = silent
@@ -94,6 +97,9 @@ compile_model <- function(model, backend, ...) {
   require_package("cmdstanr")
   args <- list(...)
   args$stan_file <- cmdstanr::write_stan_file(model)
+  if (cmdstanr::cmdstan_version() >= "2.29.0") {
+    .canonicalize_stan_model(args$stan_file, overwrite_file = FALSE)
+  }
   if (use_threading(threads)) {
     args$cpp_options$stan_threads <- TRUE
   }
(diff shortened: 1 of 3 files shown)


R/backends.R
```diff
@@ -610,3 +616,33 @@ repair_stanfit_names <- function(x) {
...
+
+.canonicalize_stan_model <- function(stan_file, overwrite_file = TRUE) {
+  if (os_is_windows()) {
+    stanc_cmd <- "bin/stanc.exe"
+  } else {
+    stanc_cmd <- "bin/stanc"
+  }
+  stanc_flags <- c(
+    "--auto-format",
+    "--canonicalize=deprecations,braces,parentheses"
```
rok-cesnovar, line 628: `braces,parentheses` is optional. This just adds { } for single line if/for, and braces cleans up any extra braces.

R/make_stancode.R
```diff
@@ -307,6 +307,12 @@ make_stancode <- function(formula, data, family = gaussian(),
   if (parse) {
     scode <- parse_model(scode, backend, silent = silent)
   }
+  if (backend == "cmdstanr") {
+    if (cmdstanr::cmdstan_version() >= "2.29.0") {
+      tmp_file <- cmdstanr::write_stan_file(scode)
+      scode <- .canonicalize_stan_model(tmp_file, overwrite_file = FALSE)
```
rok-cesnovar, line 313: We didn't discuss this in the issue, but I added this for make_stancode as well, so that the output code also doesn't produce warnings if cmdstanr is used.

paul-buerkner:
Thank you! The changes look good to me!

rok-cesnovar:
The unit tests dont test for the cmdstanr specifics right? In that case feel free to merge.

paul-buerkner:
I am not sure I understand. Can you elaborate?

rok-cesnovar:
Are there any unit tests in tests/ that run with the cmdstanr backend only?

paul-buerkner:
Yes, severaly, starting at line 2279 in https://github.com/paul-buerkner/brms/blob/master/tests/testthat/tests.make_stancode.R

paul-buerkner:
Perfect, thank you so much! Merging now.

paul-buerkner:
One problem this canonicalizing implies is that we cannot compile these models with rstan anymore afterwards. This will make features such a bridgesampling or moment matching fail if you have been fitting your model with Cmdstanr and Stan >= 2.29.0. Is there something we can do you fix this?

R/backends.R
```diff
@@ -24,6 +24,9 @@ parse_model <- function(model, backend, ...) {
 .parse_model_cmdstanr <- function(model, silent = 1, ...) {
   require_package("cmdstanr")
   temp_file <- cmdstanr::write_stan_file(model)
+  if (cmdstanr::cmdstan_version() >= "2.29.0") {
+    .canonicalize_stan_model(temp_file, overwrite_file = TRUE)
+  }
   out <- eval_silent(
     cmdstanr::cmdstan_model(temp_file, compile = FALSE, ...),
     type = "message", try = TRUE, silent = silent
```
rok-cesnovar, line 32: @paul-buerkner I might be wrong, but I think if this function just returned `model`, that might work as the model is no canonicalized? 
I honestly don't know what `bridgesampling` uses (where it gets the code from), but the idea to fix this would be to only canonicalize right before calling `cmdstan_model()`.

R/backends.R
```diff
@@ -94,6 +97,9 @@ compile_model <- function(model, backend, ...) {
   require_package("cmdstanr")
   args <- list(...)
   args$stan_file <- cmdstanr::write_stan_file(model)
+  if (cmdstanr::cmdstan_version() >= "2.29.0") {
+    .canonicalize_stan_model(args$stan_file, overwrite_file = TRUE)
+  }
```
rok-cesnovar, line 102: If bridgesampling uses the stan_file, we need to also remove this.

R/backends.R
paul-buerkner: Here is a reprex:

```R
fit3 <- brm(
  time | cens(censored) ~ age * sex + disease + (1|patient),
  data = kidney, family = lognormal(), 
  backend = "cmdstanr"
)

bridge_sampler(fit3)
```

The problem is that "brmsfit$model" also has the new array syntax but `add_rstan_model`, which is called for bridgesampling and friends doesn't know that. I have just fixed it on master.

PR #1314, opened 2022-03-11
projpred: Make `extract_model_data()` return a factor `y` if necessary

fweber144:
This fixes a bug in `get_refmodel.brmsfit()` for Bernoulli models having a `factor`-like response:
```r
# Data --------------------------------------------------------------------

data("df_gaussian", package = "projpred")
dat <- data.frame(y = df_gaussian$y, df_gaussian$x)
set.seed(457211)
icpt <- -0.42
dat$y <- rbinom(nrow(dat), size = 1,
                prob = binomial()$linkinv(icpt + 0.8 * dat$X1))
dat$y <- as.factor(paste0("resp", dat$y))

# Fit ---------------------------------------------------------------------

library(brms)
fit <- brm(y ~ X1 + X2 + X3 + X4 + X5,
           data = dat,
           family = bernoulli(),
           control = list(adapt_delta = 0.95,
                          max_treedepth = 15L),
           refresh = 0,
           seed = 1140350788)

# projpred ----------------------------------------------------------------

library(projpred)
refm <- get_refmodel(fit)

d_test_dummy <- list(type = "test",
                     y = refm$y,
                     test_points = seq_along(refm$y),
                     data = refm$fetch_data(),
                     weights = refm$wobs,
                     offset = refm$offset)
vs <- varsel(refm, d_test = d_test_dummy,
             nterms_max = 2, nclusters = 2, nclusters_pred = 3, seed = 123)
## --> Gives:
# Error: New factor levels are not allowed.
# Levels allowed: 'resp0', 'resp1'
# Levels found: '0', '1'
##

cvvs <- cv_varsel(refm, cv_method = "kfold", K = 2,
                  nterms_max = 2, nclusters = 2, nclusters_pred = 3, seed = 123)
## --> Gives:
# Error: New factor levels are not allowed.
# Levels allowed: 'resp0', 'resp1'
# Levels found: '0', '1'
##
```
At first, I tried to set `check_response = FALSE` in the `posterior_linpred()` call which is performed in `ref_predfun()`, but I got the error `formal argument "check_response" matched by multiple actual arguments`. So the changes from this PR are kind of a workaround.

As may be seen from the first of the two commits included here, this fix is not only important for Bernoulli models with a `factor`-like response, but also for the upcoming augmented-data approach in projpred.

### DIFF
--- a/R/projpred.R
+++ b/R/projpred.R
@@ -111,7 +111,8 @@ get_refmodel.brmsfit <- function(object, newdata = NULL, resp = NULL,
   
   # allows to handle additional arguments implicitly
   extract_model_data <- function(object, newdata = NULL, ...) {
-    .extract_model_data(object, newdata = newdata, resp = resp, ...)
+    .extract_model_data(object, newdata = newdata, resp = resp,
+                        aug_data = aug_data, ...)
   }
   
   # The default `ref_predfun` from projpred does not set `allow_new_levels`, so
@@ -193,7 +194,8 @@ get_refmodel.brmsfit <- function(object, newdata = NULL, resp = NULL,
 
 # auxiliary data required in predictions via projpred
 # @return a named list with slots 'y', 'weights', and 'offset'
-.extract_model_data <- function(object, newdata = NULL, resp = NULL, ...) {
+.extract_model_data <- function(object, newdata = NULL, resp = NULL,
+                                aug_data = FALSE, ...) {
   stopifnot(is.brmsfit(object))
   resp <- validate_resp(resp, object, multiple = FALSE)
   family <- family(object, resp = resp)
@@ -208,14 +210,13 @@ get_refmodel.brmsfit <- function(object, newdata = NULL, resp = NULL,
   data <- current_data(object, newdata, resp = resp, check_response = TRUE)
   attr(data, "terms") <- NULL
   y <- unname(model.response(model.frame(respform, data, na.action = na.pass)))
-  # TODO: uncomment as soon as the 'aug_data' flag is implemented
-  # if (is_like_factor(y)) {
-  #   y_lvls <- levels(as.factor(y))
-  #   if (aug_data && !is_equal(y_lvls, family$cats)) {
-  #     stop2("The augmented data approach requires all response categories to ",
-  #           "be present in the data passed to projpred.")
-  #   }
-  # }
+  if (aug_data) {
+    y_lvls <- levels(as.factor(y))
+    if (!is_equal(y_lvls, family$cats)) {
+      stop2("The augmented data approach requires all response categories to ",
+            "be present in the data passed to projpred.")
+    }
+  }
   
   # extract relevant auxiliary data
   # call standata to ensure the correct format of the data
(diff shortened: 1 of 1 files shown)


paul-buerkner:
Thanks! Some of the code you are using is unreliable because brms puts so much stuff on left-hand sides of formulas. I will make some edits to the PR before merging.

fweber144:
Sure, thanks.

paul-buerkner:
I have made some edits to your PR. Can you check and verify that this still does what you want? Also, I commented out the `aug_data` check for now because this argument is not yet implemented in brms to my knowledge. Is that correct?

paul-buerkner:
Edit: I see that aug_data is present. Sorry. Let me fix that real quick.

paul-buerkner:
Done. Should be ready from my side.

fweber144:
Thanks. Unfortunately, one of projpred's unit tests is failing now. The reason is that the `current_data()` call in [this line](https://github.com/fweber144/brms/blob/692a955d38d9632556e1e30adc3dd102e248d2f5/R/projpred.R#L207) doesn't work for multilevel models where `newdata` has new group levels. It seems like we could set `allow_new_levels = TRUE` there (should be passed to `validate_newdata()`), but perhaps we could also revert to my original proposal: Lines https://github.com/paul-buerkner/brms/blob/9a038e537d5fef52e2baf22d9192d70e74f3ae62/R/projpred.R#L97-L100 should ensure that the left-hand side of the formula doesn't contain any terms from `` ?brms::`addition-terms` ``, right? Sorry, those lines where probably the reason why I was using `eval(formula[[2]], data, environment(formula))`, but I completely forgot about that when I read your comment

> Thanks! Some of the code you are using is unreliable because brms puts so much stuff on left-hand sides of formulas. I will make some edits to the PR before merging.

above.

paul-buerkner:
You are right. Good catch, thank you! I am using the `model.frame` approach to ensure that `y` is processed exactly as in other places of brms which I think is safer despite probably equivalent to your `eval` approach in almost all cases. 

Thank you for your work on this PR! Merging now.

fweber144:
Thanks for merging!

Now I guess we should probably also set `allow_new_levels = TRUE` in the `current_data()` call in `get_refmodel.brmsfit()` (line https://github.com/paul-buerkner/brms/blob/57e3da9e954d090f42467f8402c5f9a9b9236a6e/R/projpred.R#L173). In fact, this is only relevant if `get_refmodel.brmsfit()` is called with non-`NULL` `newdata`. As discussed at PR #1292, I can't think of a situation where this would be necessary, but I might be wrong.

Apart from that, I have one question left: Originally, I created `y_lvls` within `get_refmodel.brmsfit()` to ensure that all response levels from the original dataset are used. Now with your changes I was worried that in cases where `newdata` does not contain all levels from the original dataset, the `y` returned by `.extract_model_data()` would have a smaller level set than the response in the original dataset. But it seems like the `current_data()` call in `.extract_model_data()` (lines https://github.com/paul-buerkner/brms/blob/57e3da9e954d090f42467f8402c5f9a9b9236a6e/R/projpred.R#L207-L210) ensures that all levels from the original dataset are also in `data` (and therefore also in `y`), right?

paul-buerkner:
I agree and have added `allow_new_levels = TRUE` accordingly.

Indeed, since `current_data` calls `validate_newdata` internally if newdata is passed, all factor levels are set correctly.

fweber144:
> I agree and have added `allow_new_levels = TRUE` accordingly.

Thanks!

> Indeed, since `current_data` calls `validate_newdata` internally if newdata is passed, all factor levels are set correctly.

Nice!

PR #1319, opened 2022-03-14
Add Beta-binomial family

hdrab127:
Hey, this should add `beta_binomial` as discussed. Will be working on updating the custom families vignette later in the week.

The first commit (https://github.com/paul-buerkner/brms/commit/4dfe0f018e221a8ad9a04f6595c6980dbc2240df) fixes my bug in the recent `zero_inflated_beta_binomial` code where it was calling the wrong `zi` link Stan function, i.e., it was using `bernoulli_logit` for non-distributional models - sorry I missed that! Then adds the previously missing `lccdf`/`lcdf` functions for censored/truncated models.

The second commit (https://github.com/paul-buerkner/brms/commit/39b4b73e0185e8bef1f7138a8f734f686eed8eb0) adds the Beta-binomial family and looks to produce the same results as your custom family `beta_binomial2` definition. This link has a quick example: https://hdrab127.github.io/brms-zibb-example/brms_zeroinflated_betabinomial.html#model-3a-beta-binomial-custom (scroll down for the in-built and `pp_check`/`loo`), I can test a few more different models if it helps?

Edit: sorry about the force push! I was trying to amend the last branch commit to include this pull request number in NEWS but probably wasn't the cleanest approach 🤦‍♂️.


paul-buerkner:
Great! Thank you! Would you like the vignette changes to be part of this PR or handled separately? I don't mind. If you want to make it part of this PR, I would just wait with reviewing it until you say it is complete from your side.

hdrab127:
Part of this PR sounds simplest I guess, thanks 👍

paul-buerkner:
I need to release brms 2.17 to CRAN soon. Would you like this PR to be part of it as well? In this case, we still would have to make the changes to the vignette.

paul-buerkner:
Actually, since I would like to move the release forward today, I will just merge your PR after a few minor edits from my side. We can then just revisit the vignette at a later point if you like. Thank you for working on the beta-binomial feature!

paul-buerkner:
Merging now :-)

hdrab127:
Cool, thanks Paul! 

Sorry dropped off the radar there. I had a good few goes at the log-normal natural-scale custom family for this vignette but kept running into sampling efficiency/convergence issues with that parameterisation. This was trying it on the simplified censored kidney example from the brms overview. I think, at least for that model, sigma is blowing up to try and pull scale larger mu values. Will look more into it and maybe add some comments on the original natural scale feature request.

PR #1321, opened 2022-03-15
Fix workflow failures

andrjohns:
Add `cmdstanr` installation to workflow setup to fix dependency errors

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -39,6 +39,13 @@ jobs:
           http-user-agent: ${{ matrix.config.http-user-agent }}
           use-public-rspm: true
 
+      - name: Install cmdstanr
+        run: |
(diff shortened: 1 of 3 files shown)


andrjohns:
Yep it's just struggling because cmdstanr isn't on CRAN. I'm debugging on my fork (so I don't clog up your notifications!), I've just about got it sorted

andrjohns:
Quick update - I've figured out how to handle the cmdstanr install, which also identified that some of the tests were failing because of the addition of the code canonicaliser in #1304 means that the reference values for some of the `make_stancode` tests were no longer accurate.

I've fixed those issues and now I'm just chasing down some failures while building vignettes on Windows

andrjohns:
@paul-buerkner Sorry about the false start, this is all good to go now

paul-buerkner:
Thank you!

Issue #878, opened 2020-04-16
Are parameter estimates for `cosy` constrained to be positive?


mabarbour:
Are parameter estimates for `cosy` constrained to be positive? Below is a reproducible example with the attached dataset [example_df.xlsx](https://github.com/paul-buerkner/brms/files/4486487/example_df.xlsx).

`brms` code:
`library(brms)`
`summary(brm(Y ~ T1 + T2 + (1 | Group) + cosy(gr = Group), data = example_df, control = list(adapt_delta = 0.95)))`

What I believe is the equivalent code in `nlme`:
`library(nlme)`
`summary(lme(Y ~ T1 + T2,  data = example_df, random = ~ 1 | Group, correlation = corCompSymm(form = ~ 1 | Group)))`

The population-level estimates are virtually the same, but `nlme` estimates a negative value for the `rho` whereas `brms` estimates a positive value. In addition, the lower bound for the `brms` estimate is zero.

paul-buerkner:
Yes, the correlation is bound by zero in brms. Is there any theoretical reason, why a negative value would make sense for a cosy structure? I understand that nlme may find negative estimates but that doesn't (necessarily) mean they are sensible.

mabarbour:
It is my understanding that this correlation will be negative if the variance between groups is smaller than the variance within groups. This happens commonly in split-plot field experiments. One example is given by Nelder 1977:
<img width="685" alt="Screen Shot 2020-04-16 at 3 17 56 PM" src="https://user-images.githubusercontent.com/2720243/79460774-944cd600-7ff5-11ea-943a-953517cc12f1.png">

It has been argued that one way to account for this "excess variance" is to fit a compound symmetry correlation structure ([Ben Bolker on Stackoverflow](https://stackoverflow.com/questions/32407631/mixed-effects-model-with-negative-variances/32407816)). Without the compound symmetry correlation structure, these group-level effects would be estimated as effectively zero in a mixed-model. This occurs in both `nlme` or `lme4`:

`summary(lme(Y ~ T1 + T2, data = example_df, random = ~ 1 | Group))`
 or
`summary(lmer(Y ~ T1 + T2 + (1 | Group), data = example_df))`

To my understanding, this suggests that the residual correlations of the model have been mis-specified, which will affect the uncertainty in the parameter estimates. 

Below are some additional links to previous discussions I've seen on this topic, by people more knowledgeable than me:

[...]

Issue #1094, opened 2021-02-08
Priors: Impose bounds on more classes


hsbadr:
Currently, upper and lower bounds are only allowed for classes `"b"`. In many cases, with strong priors on the ranges, imposing bounds significantly reduces the computation time.

https://github.com/paul-buerkner/brms/blob/e45f64e8e73b87605d7bf39e1eac2b171818fe48/R/priors.R#L19-L22

Moreover, it isn't allowed to impose bounds with QR decomposition.
https://github.com/paul-buerkner/brms/blob/5eca085ec227cf65b5002d719501525f483c2bb9/R/stan-predictor.R#L335

paul-buerkner:
I agree. The prior bounds framework needs to be reworked and made more flexible.

drjhw:
I have the same problem - my model only converges if I set both the 'lb' and 'ub' arguments using set_prior. I need these bounds to constrain the intercept for the half-time in a non-linear model (similar to SSlogis, in package nlme)  to sensible (positive) values. But, setting these bounds for the intercept then prevents the model from estimating coefficients for the variables of interest, because these covariates can lengthen or shorten the half-time. 
I've been hoping to perform these analyses for more than a decade. So, I was excited to find that brms could estimate the basic model, with no covariates - but then frustrated to find that I could not relax the bounds to estimate the covariate effects.

DoktorMike:
I also would love this feature, but currently you can still do it if you use nonlinear formulas instead which forces you to set priors for all of your parameters. It's very flexible but also quite verbose.

paul-buerkner:
Closed via #1324

lnalborczyk:
Hi Paul,

I am echoing the remark of @mpirikL above, it does not seem possible to specify a prior with a lower bound on a specific beta parameter. For instance, `prior(normal(0, 1), class = b, coef = "x", lb = 0)` gives the following error: `Error: Argument 'coef' may not be specified when using boundaries`.

Do you know whether there is a way to get around this issue?

Thank you!

Ladislas

paul-buerkner:
use brms non linear syntax to define two separate linear predictors to
specify different bounds.

Ladislas Nalborczyk ***@***.***> schrieb am Fr., 11. Nov.
2022, 15:06:

lnalborczyk:
Perfect, thanks!

mschubert:
For reference, an example to

> use brms non linear syntax to define two separate linear predictors to specify different bounds

[...]

PR #1324, opened 2022-03-21
Refactor and unify parameter boundary interfaces

paul-buerkner:
This PR refactors and unifies parameter boundary interfaces such that bounadries of all parameters can be changed by the user via the `prior` interface.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -2,8 +2,8 @@ Package: brms
 Encoding: UTF-8
 Type: Package
 Title: Bayesian Regression Models using 'Stan'
-Version: 2.16.11
-Date: 2022-03-07
+Version: 2.16.12
+Date: 2022-03-21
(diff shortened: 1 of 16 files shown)


paul-buerkner:
This should be basically ready except for changing the boundary storage in the `brmsprior` object, where the `bound` column should be replaced by two columns named `lb` and `ub` for increased consistency in internal and external interfaces

paul-buerkner:
While working on this PR, I found some issues with internal design choices specifically related to the selected "base-prior". I left some TODOs at the relevant places and will fix them within this PR. However, I will have to think about the most consistent behavior for a bit, so will set this PR aside for now. 

As a reminder to myself, I also need to `restructure` priors of old models to correctly include all the bounds.

paul-buerkner:
This PR will close issues #878 and #1094

PR #1318, opened 2022-03-13
Trim all trailing whitespace

andrjohns:
This PR simply trims all trailing whitespace present in `brms`. My IDE was automatically trimming whitespace while working on files, which leads to much more significant diffs than necessary.

This PR touches a lot of files, but it's a bit more manageable to review if you use the 'ignore whitespace' filter by adding `?w=1` to the URL of the PR. I'll add a comment with a link once the PR is opened and the link is available.

I'm also happy to split this into multiple PRs in case you want to review in more depth, just let me know!

PR #1337, opened 2022-04-21
Fix PR #1314

fweber144:
In PR #1314, I didn't realize that your changes are not quite equivalent to my original code: The check which is now performed in lines https://github.com/paul-buerkner/brms/blob/6b3c626672a1bf25514da4b7268dedfe7add9de0/R/projpred.R#L218-L224 of `brms:::.extract_model_data()` fails for non-factor responses (of a categorical or ordinal model) if `newdata` does not contain all unique values from the original dataset. That's why I performed that check in `get_refmodel.brmsfit()` based on the original dataset (lines https://github.com/paul-buerkner/brms/blob/55490722db4346fe6a3ed510f4054e584d6f5ca6/R/projpred.R#L119-L122), but in [this comment](https://github.com/paul-buerkner/brms/pull/1314#issuecomment-1065947185) I forgot about non-factor responses. Sorry for that.

With this PR, I decided to omit that check completely as it is error-prone and because I will perform it in a more general way within projpred.

Btw, I saw that PR #1318 trimmed all trailing whitespace, but `brms.Rproj` does not have `StripTrailingWhitespace: Yes`. So when using RStudio's indentation mechanism, the current `brms.Rproj` might introduce new trailing whitespace.

### DIFF
--- a/R/projpred.R
+++ b/R/projpred.R
@@ -214,14 +214,6 @@ get_refmodel.brmsfit <- function(object, newdata = NULL, resp = NULL,
     allow_new_levels = TRUE
   )
   y <- unname(model.response(model.frame(respform, data, na.action = na.pass)))
-  aug_data <- is_categorical(formula) || is_ordinal(formula)
-  if (aug_data) {
-    y_lvls <- levels(as.factor(y))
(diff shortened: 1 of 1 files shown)


paul-buerkner:
Thank you!

PR #1361, opened 2022-05-24
Fix class definition removed from ‘dbplyr’ #issue 1356

munoztd0:
added  Matrix (>= 1.4.0), to dependencies to fix class definition removed class with tidyverse

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -20,7 +20,6 @@ Depends:
     Rcpp (>= 0.12.0),
     Matrix (>= 1.4.0),
     methods
-    Matrix
 Imports:
     rstan (>= 2.19.2),
     ggplot2 (>= 2.0.0),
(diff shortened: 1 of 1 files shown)


paul-buerkner:
Thanks! Much appreciated!

paul-buerkner:
But why does it have to be under depends? Wouldn't imports be enough? Things in depends are automatically loaded into the environment when loading brms which is not ideal as it clutters the environment with a lot of function.

munoztd0:
I totally agree but that must be an issue with tidyverse_conflicts() I'm affraid

paul-buerkner:
I see. Thank you! I will think of that a little bit and see if some other options come up that avoids putting it under depends.

paul-buerkner:
I think this issue was resolved outside of brms. Closing this now. Thank your for your time and effort you put into finding a solution from the brms side!

Issue #1135, opened 2021-04-06
Update with cmdstan backend throws error, or requires recompilation


jflournoy:
If I attempt to `update` a brms object that uses `backend = cmdstan` it either throws the following error:

```
Start sampling
Error: Model not compiled. Try running the compile() method first.
```

or requires recompilation even with trivial changes to the data.

Here is a reproducible example:

```
library(brms)
library(cmdstanr)
cmdstanr::cmdstan_version()
# [1] "2.26.1"
packageVersion('brms')
# [1] ‘2.15.0’

data <- data.frame(y = rnorm(500))
threads <- threading(threads = 4)

compiled <- brms::brm(y ~ 1, data = data, 
                      chains = 1, iter = 100,
                      backend = 'cmdstan',
                      file = 'compiled')

#Clear env and restart R
library(brms)
library(cmdstanr)

newdata <- data.frame(y = rnorm(500, mean = 10))
compiled <- readRDS('compiled.rds')
fit <- update(compiled, newdata = newdata, 
              chains = 1, iter = 1000, warmup = 500)
#Start sampling
#Error: Model not compiled. Try running the compile() method first.

fit2 <- update(compiled, newdata = newdata, 
               chains = 1, iter = 1000, warmup = 500,
               backend = 'cmdstan')
#The desired updates require recompiling the model
#Compiling Stan program...
#Start sampling
```

paul-buerkner:
There are two things going on:

1) The second model needs recompilation because you misspecified the backend. It should be `"cmdstanr"` not `"cmdstan"`. I just pushed a minor fix to github to return an informative error in case of a misspecified backend (or a valid backend when partial matching is possible) instead of a recompilation trigger.

2) cmdstanr seems to be unable to run a model compiled in another R session. @jgabry do you have any idea how to fix this from the brms or cmdstanr side?

jgabry:
Hmm, in general cmdstanr shouldn't need to recompile if it can find the
compiled executable. Is brms doing compilation in a temporary directory? If
so then that would require recompiling if there's a new R session because
it wouldn't be able to find the compiled executable.

>

paul-buerkner:
It seems to write to a tmp-directory indeed. At least this is what str(<CmdStanModel>) tells me. I am not sure I specify that explicitely though somewhere. Here is the code I used to compile cmdstanr model:

```R
# compile Stan model with cmdstanr
# @param model Stan model code
# @return model compiled with cmdstanr
[...]

PR #1374, opened 2022-07-05
Document option `cmdstanr_write_stan_file_dir` for `update.brmsfit()`

fweber144:
This documents how cmdstanr's option `cmdstanr_write_stan_file_dir` can be used to avoid a recompilation when calling `update.brmsfit()` on a cmdstanr-backend fit in a fresh R session, thereby solving #1135 completely (without the need for further changes in cmdstanr).

I wasn't sure how to clean up the files which are currently created permanently by the new example (two `model_<hash>.stan` files and one `model_<hash>(.exe)` executable). Deleting all regexp'ed `^model_.*\\.stan` files and corresponding executables after running the example would be dangerous. Less dangerous alternatives would be to

1. diff the list of files before and after running the new example and delete all files created in between;
2. extend the `if (!file.exists("fit_cmdstanr.rds"))` condition to only run the new example if no `^model_.*\\.stan` files and no such executables exist in the current working directory. Then after running the example, we could simply remove all `^model_.*\\.stan` files and corresponding executables.

What do you think?

**EDIT:** Clarified the wording for cleaning up the files.

### DIFF
--- a/R/update.R
+++ b/R/update.R
@@ -62,7 +62,6 @@
 #'                       file = "fit_cmdstanr")
 #'   upd_cmdstanr <- update(fit_cmdstanr,
 #'                          formula. = rate ~ conc)
-#'   file.remove("fit_cmdstanr.rds")
 #' }
 #' }
 #'
--- a/R/update.R
+++ b/R/update.R
@@ -14,6 +14,13 @@
 #'   arguments to be ignored.
 #' @param ... Other arguments passed to \code{\link{brm}}.
 #'
+#' @details When updating a \code{brmsfit} created with the \pkg{cmdstanr}
+#'   backend in a different \R session, a recompilation will be triggered
+#'   because by default, \pkg{cmdstanr} writes the model executable to a
+#'   temporary directory. To avoid that, set option
+#'   \code{"cmdstanr_write_stan_file_dir"} to a path of your choice before
+#'   creating the original \code{brmsfit} (see section 'Examples' below).
+#'
 #' @examples
 #' \dontrun{
 #' fit1 <- brm(time | cens(censored) ~ age * sex + disease + (1|patient),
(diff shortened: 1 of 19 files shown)


wds15:
Why should there ever be an issue? The hash is **unique** to the model itself. There is only ever an issue if you did change the compile options you want to use - but I am not sure how to deal with that in automation. In fact, I'd argue that if someone changes compile options, then wiping out the cache directory is a very sensible thing to do.

fweber144:
I was thinking about the best way to clean up after the example, so that the state of the working directory is the same after running the example as before. With "dangerous" I meant that a user could have `^model_.*\\.stan` files and corresponding executables from a different cmdstanr model lying around in the working directory.

paul-buerkner:
Aren't the names quite convoluted and long anyway?

fweber144:
Yes, they are. I guess your point is the same as @wds15's, namely that there shouldn't be conflicts in terms of file names. I agree, but my point was rather how to clean up the newly added files.

fweber144:
If you say we don't need to clean them up (that's how it is implemented currently), I'm fine with that, too.

paul-buerkner:
I don't think you need to clean them up, as in removing these files, if the user had told explicitly where they want to put the files. I think they would need to take care of that themselves. Or am I missing your point?

R/update.R
```diff
@@ -34,6 +41,29 @@
...
 #' summary(fit4)
+#'
+#' ## to avoid a recompilation when updating a 'cmdstanr'-backend fit in a fresh
+#' ## R session, set option 'cmdstanr_write_stan_file_dir' before creating the
+#' ## initial 'brmsfit'
+#' ## CAUTION: the following code creates some permanent files (two
+#' ## 'model_<hash>.stan' files and one 'model_<hash>(.exe)' executable) in the
+#' ## current working directory, additionally to file 'fit_cmdstanr.rds' (which
+#' ## is deleted afterwards)
+#' if (!file.exists("fit_cmdstanr.rds")) {
```
paul-buerkner, line 52: Why do you check for the .rds file in this example?

paul-buerkner (commented):
Looks good. I had only one minor question to resolve before we can merge.

paul-buerkner:
Yes, that was exactly my thought.

R/update.R
fweber144: In case the user has such a file lying around in the current working directory (unlikely, but not impossible).

fweber144:
Done. So I think this is ready to be merged.

fweber144:
And sorry, @wds15, for my mistakable wording. I edited the initial comment so it's clearer now.

R/update.R
fweber144: Ok, then I would add `as.numeric(Sys.time())`. Is that ok?

paul-buerkner:
But you still check for the existence of the file. Can this be removed too and instead the name be made less common to avoid name conflicts?

R/update.R
paul-buerkner: Yeah, I think

fweber144:
Yes, I wrote https://github.com/paul-buerkner/brms/pull/1374#issuecomment-1177207550 before I read your comment on that.

R/update.R
fweber144: Oh, taking the time is a bad idea because the example involves a restart of the R session in between. I'll follow your RNG suggestion.

paul-buerkner:
Or something like `fit_example_update_cmdstanr`. Nobody will name their models like that :-D

R/update.R
fweber144: I picked the RNG suggestion now (commit https://github.com/paul-buerkner/brms/pull/1374/commits/89d56d1e3140725619672ba4b1ae4875518e452a).

fweber144:
Should be done now. Please check if it's ok this way.

paul-buerkner:
Looks good. Thank you! Merging now.

Issue #1378, opened 2022-07-11
`kfold()`: Arguments (from `...`?) not passed to `brm()`


fweber144:
This is similar to #1373, but for `kfold.brmsfit()` instead of `update.brmsfit()`. So here it seems like `kfold.brmsfit()` does not pass all arguments from the original `brm()` call to the K fold-specific `brm()` calls (I guess arguments from the ellipsis `...` are missing). For example: argument `adapt_delta` in case of the cmdstanr backend:
```r
library(brms)
options(mc.cores = parallel::detectCores(logical = FALSE))
fit_cmdstanr <- brm(rate ~ conc + state,
                    data = Puromycin,
                    refresh = 0,
                    seed = 11403,
                    backend = "cmdstanr",
                    adapt_delta = 0.9)
unique(sapply(fit_cmdstanr$fit@stan_args, "[[", "delta"))
## Gives: 0.9
kfold_cmdstanr <- kfold(fit_cmdstanr, K = 2, save_fits = TRUE, seed = 789)
fit_cmdstanr_fold1 <- kfold_cmdstanr$fits[1, "fit"]$fit
unique(sapply(fit_cmdstanr_fold1$fit@stan_args, "[[", "delta"))
## Gives: 0.8
```
Specifying `adapt_delta` in the `kfold()` call helps (even though the `...` documentation at `?kfold.brmsfit` only mentions "post-processing functions"):
```r
kfold_cmdstanr_add <- kfold(fit_cmdstanr, K = 2, save_fits = TRUE, seed = 789,
                            adapt_delta = 0.9)
fit_cmdstanr_fold1_add <- kfold_cmdstanr_add$fits[1, "fit"]$fit
unique(sapply(fit_cmdstanr_fold1_add$fit@stan_args, "[[", "delta"))
## Gives: 0.9

```
Tested with brms v2.17.0.

paul-buerkner:
Fixed.

fweber144:
Thanks! I can confirm that the example is working now.

PR #1392, opened 2022-08-12
Fix documentation for `...` in `kfold.brmsfit()`

fweber144:
This fixes the documentation for `...` in `kfold.brmsfit()` because `kfold.brmsfit()` allows to pass arguments to `brm()` (see <https://github.com/paul-buerkner/brms/issues/1378#issue-1300775618>).

Here, I have copied the `...` documentation from `brm_multiple()` instead of inheriting it from `loo.brmsfit()`. But I'm not sure if it actually contains everything that is needed for `...` in `kfold.brmsfit()`. You might have to mention more functions (apart from `brm()`) that can get arguments from `...`.

### DIFF
--- a/R/kfold.R
+++ b/R/kfold.R
@@ -38,6 +38,7 @@
 #' @param future_args A list of further arguments passed to
 #'   \code{\link[future:future]{future}} for additional control over parallel
 #'   execution if activated.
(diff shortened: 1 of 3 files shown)


paul-buerkner:
Thanks!

PR #1396, opened 2022-08-25
Fully delegate canonicalisation to cmdstanr

andrjohns:
This PR updates the Stan model canonicalisation process to be fully delegated to `cmdstanr`, rather than manually calling `stanc`. This is to allow for compatibility with the new WSL functionality in `cmdstanr`, (which is unfortunately not currently compatible with `brms` due to the manual `stanc` call)

Let me know if I can add any other clarifications/changes here!

Cheers,
Andrew

### DIFF
--- a/R/backends.R
+++ b/R/backends.R
@@ -621,10 +621,11 @@ file_refit_options <- function() {
 
 .canonicalize_stan_model <- function(stan_file, overwrite_file = TRUE) {
   cmdstan_mod <- cmdstanr::cmdstan_model(stan_file, compile = FALSE)
-  out <- capture.output(
+  out <- utils::capture.output(
     cmdstan_mod$format(canonicalize = list("deprecations",
                                           "braces",
                                           "parentheses"),
-                      overwrite_file = overwrite_file))
+                      overwrite_file = overwrite_file,
+                      backup = FALSE))
   paste0(out, collapse = "\n")
 }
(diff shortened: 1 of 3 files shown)


R/backends.R
```diff
@@ -620,32 +620,11 @@ file_refit_options <- function() {
...
-      cat(res$stdout, file = stan_file, sep = "\n")
-    }
-  }
-  res$stdout
+  cmdstan_mod <- cmdstanr::cmdstan_model(stan_file, compile = FALSE)
+  out <- capture.output(
+    cmdstan_mod$format(canonicalize = list("deprecations",
+                                          "braces",
+                                          "parentheses"),
+                      overwrite_file = overwrite_file))
```
rok-cesnovar, line 628: I know this all happens in the tmp folder, but set `backup = FALSE`

rok-cesnovar:
The changes look good, I would only [bump cmdstanr](https://github.com/paul-buerkner/brms/blob/master/DESCRIPTION#L48) to 0.5.0 as that is the first version with the `$format()` method.

andrjohns:
> The changes look good, I would only [bump cmdstanr](https://github.com/paul-buerkner/brms/blob/master/DESCRIPTION#L48) to 0.5.0 as that is the first version with the `$format()` method.

Oh good catch, thanks!

paul-buerkner:
Thank you! Is this PR ready to be merged from your side?

andrjohns:
> Thank you! Is this PR ready to be merged from your side?

Yep! This is compatible with the current `cmdstanr` (and the last few versions as well), so it's good to go

paul-buerkner:
Perfect! Merging now.

PR #1400, opened 2022-08-31
A function to read csvs via ```cmdstanr``` and repackage info as a ```stanfit``` object

SimonCMills:
Commit of two files: ```R/read_csv_as_stanfit.R``` that contains the function ```read_csv_as_stanfit()```, and ```tests/testthat/tests.read_csv_as_stanfit.R``` that contains a series of unit tests to check that the output (a stanfit object) is the same as that generated by ```rstan::read_stan_csv()``` for a variety of different sampling types (no warmup, dense warmup, no samples, thinned samples, etc.). You might want to add something that prevents these tests being run on CRAN. 

As a broad overview, ```read_csv_as_stanfit()``` uses ```cmdstanr::read_cmdstan_csv()``` to read csvs in, and then repackages these into a stanfit object. While it also allows for selecting a subset of variables at the read stage, ```stanfit``` methods have issues if you select out specific subsets of a variable (e.g. ```x[1]```). You can create the ```stanfit``` object fine, but it will error when it tries to print it (I think these issues largely arise within ```summary.stanfit```, specifically within the ```rstan:::summary_sim()``` function). On a more positive note, ```stanfit``` methods seem to have no problems so long as you select variables without reference to a specific element (e.g. select ```x```, rather than ```x[1]```), which is pretty helpful! As I've written it, you can't currently select a subset of the sampler diagnostics. 

The function currently fails on the variational inference test: the object looks sensible enough, but isn't identical in the metadata that is stored. I can probably get round to fixing this in the next week or so, but have left it as it is for the time being.

I hope this is vaguely useful. Let me know if there is anything more I can help with!


paul-buerkner:
Thank you a lot! This looks really nice upon first glance! I will check it out in more detail in the next couple of days. 

One more question: I would like to add you as a contributor to brms for this effort. Is that okay for you? And is there anyone else who I should additionally credit for this code?

SimonCMills:
That would be great, if you are happy with that! @jsocolar contributed to the test code.

paul-buerkner:
I have updated the PR a bit. In particular, we can now finally ensure that only relevant variables's draws (instead of all variables' draws) are stored in models fitted with the cmdstanr backend. 

Could you check whether the meta data I added in DESCRIPTION and NEWS is sensible? Once you confirm that (and all checks have passed) I will merge this PR.

SimonCMills:
This all looks good to me! This will be super helpful, thanks for bringing it into brms.

rok-cesnovar:
Under which license is this being contributed. I am wondering because this looks great and I think this might be worth copying to cmdstanr directly at some point.

SimonCMills:
Hmm, I guess these would by default be under brms's licensing. If it is useful, I can quickly upload the two functions I wrote to a separate repository under a BSD-3 license which wouldn't (I don't think) affect licensing for it being brought into brms, but would mean you could just bring it into cmdstanr at a later date without any licensing issues.

paul-buerkner:
Great! any aspect that I contributed to this can of course be reused as
well in cmdstanr under its license. in fact, having it in cmdstanr would
probably be preferable in the long run anyway.

Rok Češnovar ***@***.***> schrieb am Do., 1. Sept. 2022,
17:21:

paul-buerkner:
Merging this now. Thank you again, @SimonCMills, for making this feature possible!

Issue #1307, opened 2022-03-03
Feature Request: AR correlation structure for random effects


davibf11:
Hello,

First and foremost, thank you for all of the excellent work that's done here! `brms` is an excellent package and really helps decrease dependence on working in base `stan`, which is helpful when working with others (as fun as it is to pick through thousands of lines of a complicated model...).

One feature that I think would be really helpful for `brms` to be able to implement would be to allow for AR correlation structures within random effects. This is something that I do frequently in base `stan` when working with data that has groups that are ideal to pool information over, and which also has multiple observations over multiple time periods. 

I've put together an example script (and `stan` code) together [here](https://github.com/davibf11/ar_example_model/tree/main) to give an idea about what I'm talking about. This is different from the `stan` models that I usually use - I modified it to try to use a lot of similar code to what `brms` usually outputs. Essentially, we'd be fitting the following model:

`y ~ 1 + (ar(time, p = 1) | group)`

The correlation structure is achieved using a pretty straightforward spectral decomposition. This allows for each group to have a separate random effect for each time period with the specific correlation structure. I'm sure there's a more intuitive way to set this up in `brms` to allow for users to set different `p` values, but I figured I would put something together to give an idea of what I'm talking about. Let me know if this doesn't make sense, or if there's anything else I can do to clarify!

bachlaw:
This would be a very welcome addition.  `glmTMBB` has a natural implementation of many of these structures which seem to work (exclusively?) at the group level, and anything from the `lme4` group is usually a good start.  https://cran.r-project.org/web/packages/glmmTMB/vignettes/covstruct.html

jackobailey:
Possibly related to this issue https://github.com/paul-buerkner/brms/issues/1025

bachlaw:
As a Bayesian example would probably be more helpful, [INLA ](https://inla.r-inla-download.org/r-inla.org/doc/latent/ar.pdf) accommodates this through their notation:
`y ~ 1 + f(time, replicate=group_idx, model='ar', order = 1)`. However, it is possible this is simply more straightforward to do through a latent Gaussian model.

PR #1409, opened 2022-09-15
AR-correlated random effects; issue #1307

mpdylan:
This PR adds a new parameter `latent` to `ar()`, `ma()`, and `arma()`, with a default value of `FALSE`. The function of this parameter, when set to `TRUE`, is to explicitly parameterize the autocorrelated effects using latent variables, even if the model family has natural residuals and does not require such parameterization. The Stan code generated has a structure borrowed from the latent variable parameterization currently used for model families that do not admit natural residuals, such as Poisson, but with different names for variables to avoid confusion.

This PR also modifies `prepare_predictions_ac()` and other relevant functions to ensure that the autocorrelated random effects are incorporated when generating predictions using `posterior_epred()`. If new data is provided for prediction, it checks to see whether grouping levels for the new data match those in the original data used to fit the model. If no match is found for new grouping levels, it samples a new multivariate normal vector in the same way that the existing latent variable functionality does. If a match is found, it attempts to sample the autocorrelated effect conditional on the observed values, by taking the draws of the standardized latent variables for those data points for which they exist, filling in newly generated standard normal random variates for new data points, and multiplying by the lower Cholesky factor of the covariance matrix to apply the covariance structure. This process needs some work -- currently, it will only work correctly if the new data contains all of the original data, and the new data points for prediction appear contiguously in the data frame following the original data points. In particular, this means that we can't fill in unobserved time points in the middle of a time series.

In order to make the above work, there are some ancillary changes, such as modifying `make_standata()` to include grouping levels for autocorrelated effects when the `latent` flag is set to `TRUE.`

Posterior predictions using `predict()` appear to work at first glance, but I haven't tested them carefully yet.


paul-buerkner:
Thank you for working on this feature! it looks quite good already!

Just a quick initial observation from your text: the way I would like it to
work is the latent = TRUE coincides with the latent residual approach for
non normal families rather than creating redundancy by doing the same thing
twice under two different names. would it be possible to align these two to
be the same? if it takes too much time and effort for you, please let me
know and I will edit your PR accordingly.

Paul

Dylan Murphy ***@***.***> schrieb am Do., 15. Sept. 2022,
16:58:

mpdylan:
Sure, I think that would be simple to revise. I'd also like to put in some more time improving the handling of the time variable in two main ways:

- correctly match times in new data and original data so that `fitted` and `predict` can handle new time points in the middle of a series, not only at the end
- allow multiple observations for the same time point/group

paul-buerkner:
Nice, that all sounds good!

paul-buerkner:
Thank you! Is this PR ready for another round of my review? If so, can you provide me with some test cases that I could run and debug through the code to see what happens and where things may need some more changes?

mpdylan:
I thought it was ready, but I discovered a bug while preparing some test cases to send you, so I'll follow up on this soon.

I have a couple more things I would like to work on, but they may make more sense as a separate PR; e.g., implementing covariance/latent parameterization for some higher-order structures such as AR(2).

paul-buerkner:
thanks! looking forward to the examples! and I agree that the other
features should better go into a separate PR.

Dylan Murphy ***@***.***> schrieb am Fr., 2. Dez. 2022, 00:11:

mpdylan:
After fixing some bugs I think this is ready to be looked at. I've prepared a handful of examples at [mpdylan/brms_test_cases](https://github.com/mpdylan/brms_test_cases) demonstrating the latent variable functionality on time series with gaps or multiple observations on the same time point. This includes one example using synthetic data and two using publicly available data sets; models with and without linear predictors and grouping factors; and, Gaussian, Poisson, and binomial likelihoods.

R/brmsfit-methods.R
```diff
@@ -391,6 +391,83 @@ VarCorr.brmsfit <- function(x, sigma = 1, summary = TRUE, robust = FALSE,
   lapply(tmp, .VarCorr)
 }
 
+#' Extract Latent Autocorrelated Effects
```
paul-buerkner, line 394: This function has very limited application for an exported function. Why do we need it? I would prefer not to have such a function being exported?

R/brmsfit-helpers.R
```diff
@@ -255,7 +255,12 @@ get_cov_matrix_ac <- function(prep, obs = NULL, latent = FALSE) {
   }
   # prepare residual standard deviations
   if (latent) {
-    sigma2 <- as.numeric(prep$ac$sderr)^2
+    if (!is.null(prep$ac$sderr)) {
+      sigma2 <- as.numeric(prep$ac$sderr)^2
+    }
+    if (!is.null(prep$ac$sdacranef)) {
```
paul-buerkner, line 261: is "sdacranef" still needed? It seems to come from an old version of the code.

R/data-helpers.R
```diff
@@ -240,7 +240,10 @@ order_data <- function(data, bterms) {
       tv <- seq_rows(data)
     }
     if (any(duplicated(data.frame(gv, tv)))) {
-      stop2("Time points within groups must be unique.")
+      if (!parameterize_ac_effects(bterms)) {
```
paul-buerkner, line 243: why do we need to distinquish `has_ac_latent_residuals` and `parameterize_ac_effects`? Naively I would think that the former should be enough but I may be mistaken of course.

R/data-predictor.R
```diff
@@ -668,6 +669,66 @@ data_ac <- function(bterms, data, data2, basis = NULL, ...) {
       c(if (N_tg > 1L) begin_tg[2:N_tg], N + 1) - begin_tg
     ))
     out$end_tg <- with(out, begin_tg + nobs_tg - 1)
+    if (parameterize_ac_effects(acef)) {
```
paul-buerkner, line 672: this adds a lot of new data variables to standata. Why can't we reuse all the variables that we already have for ARMA correlation structure? It seems they would also be applicable to the latent version.

R/prepare_predictions.R
```diff
@@ -107,9 +107,13 @@ prepare_predictions.brmsterms <- function(x, draws, sdata, data, ...) {
   for (dp in valid_dpars) {
     dp_regex <- paste0("^", dp, resp, "$")
     if (is.btl(x$dpars[[dp]]) || is.btnl(x$dpars[[dp]])) {
+      old_acdata <- NULL
+      if (parameterize_ac_effects(x$dpars[[dp]])) {
```
paul-buerkner, line 111: why do we need to rerun full `make_standata` here again? in prepare predictions, we already compute `standata` and pass is as variable `sdata`.

R/prepare_predictions.R
```diff
@@ -764,6 +773,145 @@ prepare_predictions_ac <- function(bterms, draws, sdata, oos = NULL,
       }
     }
   }
+  if (parameterize_ac_effects(bterms)) {
```
paul-buerkner, line 776: To make sure everything is in order with the code below, I would need to discuss this code in a meeting I think.

inst/chunks/fun_scale_time_err_t.stan
```diff
@@ -0,0 +1,31 @@
...
+   */
+   vector scale_time_err_t(vector zerr, real sderr, matrix chol_cor,
+                           int[] n_times, int[] begin, int[] end, int[] times) {
+     vector[rows(zerr)] err;
+     int temp_times[max(n_times)];
+     for (i in 1:size(n_times)) {
+       // Set up the indexing vector
+       temp_times[1] = 1;
+       for (j in 1:(n_times[i]-1)) {
+         temp_times[j+1] = temp_times[j] +
```
paul-buerkner, line 22: can you explain to me what the temp_times are doing to make sure I understand correctly?

paul-buerkner (changes requested):
I think this PR looks more and more ready. Thank you for your hard work!

I have added a bunch of questions for discussion. some of them, I may also address myself later but beforehand I would like to understand some of your coding decision better to make sure I don't break stuff.

R/brmsfit-methods.R
mpdylan: I mainly added this in response to a question from a colleague who was experimenting with my fork -- they wanted a way to extract the latent variables analogous to calling `ranef` on a fitted model to extract group-level parameters. I'd be happy to do this in a different way if that would fit better.

R/brmsfit-helpers.R
mpdylan: ah, I think you're right; that was from the earliest version. will remove this.

R/data-helpers.R
mpdylan: We probably don't. I think my intent here was to avoid "overwriting" existing latent variable functionality with the new code.

R/brmsfit-methods.R
paul-buerkner: I would prefer to not have this function as part of this PR as it is too specific. I would like to think through separately if there is a good and general enough function and use-case that would contain this here as a special case. Perhaps we can discuss in a separate issue after this PR has been merged (without the function for now).

R/brmsfit-helpers.R
paul-buerkner: Thanks!

R/data-helpers.R
paul-buerkner: I would like to have the latent feature tightly integrated with the rest of the autocorrelation features and code base. Otherwise, we have multiple things that do basically the same and the branching logic becomes confusing. I am sorry that I am so particular about these things. You can let me know when there are changes that I suggest that you would prefer me to do in this PR. I know it is a bit tedious sometimes.

R/data-predictor.R
paul-buerkner: Thank you! I appreciate you looking into it!

inst/chunks/fun_scale_time_err_t.stan
paul-buerkner: I see. I think Stan will not allow you to edit input variables if I remember correctly, so using something like temp_times (or a shorter name to make lines not too long) is fine. Just wanted to understand what the purpose of this was.

paul-buerkner: Just a side note, that I will eventually want to allow missing time points in all of the correlation structures, so I will likely build onto your implementation in the future and role it out to all the correlation structures.

R/data-predictor.R
mpdylan: Upon reviewing this, I find that several of the existing variables (`begin_tg`, `end_tg`, and `nobs_tg`) are not used in the new formulation, so instead of defining new variables we could just change the definition of these. Furthermore, I think `ac_time` is no longer used anywhere, so I'll just remove it.

I do think `N_latent_err` and `latent_err_idx` are necessary to define the correct number of latent parameters and match them with rows in the data frame, and `max_time_span` is needed to avoid indexing the covariance matrix beyond its range.

R/prepare_predictions.R
mpdylan: The reason for this is that we need some information on the old grouping levels (which isn't present in `sdata` since that is computed using the `newdata`) to make predictions conditional on the original data. However, perhaps it's better to split out a helper function similar to `prepare_predictions_ranef` that packages only the necessary metadata and posterior draws for the predictions requested.

paul-buerkner: I see. brms has it's own process in place for this purpose. See `standata_basis` in `make_standata.R`

R/data-predictor.R
paul-buerkner: Changing the definition of old variables is possible but perhaps not ideal. Why would we need to change the meaning of them for use in the new structure? 

I agree that the three variables you mentioned at the end are likely needed.

mpdylan: The old variables contain the start and end indices of groups in the vector of observations; the new ones contain indices in the vector of latent variables, which will be different indices if there are any time points observed more than once. I think one could extract the same information from the original definition of these variables combined with `latent_err_idx`, eliminating the need to change them, at the cost of adding a little more code to the transformed data block in the generated Stan program, and possibly something to prepare_predictions (though I would have to think about that as there's a lot there already that can probably be simplified).

mpdylan:
Latest commit makes the following changes:
* Remove some of the unnecessary variables from the standata
* Add variables to the basis section of standata and remove the call to make_standata in prepare_predictions
* Simplify predictor_ac (which I think I had just made more complex than necessary, upon reviewing)
* Simplify prepare_predictions_ac by removing some redundant or unnecessary vectors of indices
* Fixed an error that could occur in predictions if newdata was supplied that reproduced some, but not all, of the original training data for a group level
* Remove the outdated sdacranef code and the parameterize_ac_effects, replacing the latter with has_ac_latent_residuals
* Remove the exported ac_latent function
* Produce a warning when generating predictions if the latent parameters were not saved at fit time

R/brmsfit-helpers.R
```diff
@@ -255,7 +255,9 @@ get_cov_matrix_ac <- function(prep, obs = NULL, latent = FALSE) {
   }
   # prepare residual standard deviations
   if (latent) {
-    sigma2 <- as.numeric(prep$ac$sderr)^2
+    if (!is.null(prep$ac$sderr)) {
```
paul-buerkner, line 258: Why should `prep$ac$sderr` ever be NULL in this case? Then sigma would not be defined at all?

R/data-predictor.R
```diff
@@ -663,11 +664,77 @@ data_ac <- function(bterms, data, data2, basis = NULL, ...) {
   if (use_ac_cov_time(acef)) {
     # data for the 'covariance' versions of time-series structures
     out$N_tg <- length(unique(tgroup))
-    out$begin_tg <- as.array(ulapply(unique(tgroup), match, tgroup))
-    out$nobs_tg <- as.array(with(out,
-      c(if (N_tg > 1L) begin_tg[2:N_tg], N + 1) - begin_tg
-    ))
-    out$end_tg <- with(out, begin_tg + nobs_tg - 1)
+    # if using latent variables, add grouping levels
+    if (has_ac_latent_residuals(bterms)) {
```
paul-buerkner, line 668: this looks good to me at first glance, but I would like to review this in a call with you.

R/formula-ac.R
```diff
@@ -580,10 +587,21 @@ use_ac_cov_time <- function(x) {
 
 # does the model need latent residuals for autocor structures?
 has_ac_latent_residuals <- function(bterms) {
-  !has_natural_residuals(bterms) &&
+  (!has_natural_residuals(bterms) ||
+     has_ac_subset(bterms, latent = T)) &&
     (use_ac_cov(bterms) || has_ac_class(bterms, "arma"))
 }
 
+# use explicitly parameterized autocor effects?
```
paul-buerkner, line 595: remember to remove this before we merge if we continue to not need this function.

R/make_standata.R
```diff
@@ -290,6 +290,11 @@ standata_basis_ac <- function(x, data, ...) {
       out$locations <- NA
     }
   }
+  if (has_ac_class(x, "arma")) {
```
paul-buerkner, line 293: you are overwriting the complete "out" variable instead of adding to it here. Please add the new variables to the existing list.

R/formula-ac.R
```diff
@@ -71,11 +71,11 @@ NULL
 #' }
 #'
 #' @export
-arma <- function(time = NA, gr = NA, p = 1, q = 1, cov = FALSE) {
+arma <- function(time = NA, gr = NA, p = 1, q = 1, cov = FALSE, latent = FALSE) {
```
paul-buerkner, line 74: Is argument "latent" documented somewhere already?

R/prepare_predictions.R
```diff
@@ -742,25 +743,146 @@ prepare_predictions_ac <- function(bterms, draws, sdata, oos = NULL,
     out$end_tg <- sdata[[paste0("end_tg", p)]]
   }
   if (has_ac_latent_residuals(bterms)) {
+    # prepare predictions for latent ar/ma models
+    out$level_tg <- sdata[[paste0("level_tg", p)]]
     err_regex <- paste0("^err", p, "\\[")
+    zerr_regex <- paste0("^zerr", p, "\\[")
```
paul-buerkner, line 749: Why do we need "zerr" in addition to "err" here?

R/prepare_predictions.R
```diff
@@ -742,25 +743,146 @@ prepare_predictions_ac <- function(bterms, draws, sdata, oos = NULL,
...
-      out$err <- matrix(nrow = nrow(draws), ncol = length(out$Y))
-      sderr_regex <- paste0("^sderr", p, "$")
-      out$sderr <- prepare_draws(draws, sderr_regex, regex = TRUE)
-      for (i in seq_len(out$N_tg)) {
-        obs <- with(out, begin_tg[i]:end_tg[i])
-        zeros <- rep(0, length(obs))
-        cov <- get_cov_matrix_ac(list(ac = out), obs, latent = TRUE)
-        .err <- function(s) rmulti_normal(1, zeros, Sigma = cov[s, , ])
-        out$err[, obs] <- rblapply(seq_rows(draws), .err)
+      if (new) {
```
paul-buerkner, line 763: Let's review this part during our call.

R/priors.R
```diff
@@ -995,10 +995,15 @@ prior_ac <- function(bterms, def_scale_prior, ...) {
     prior <- prior +
       brmsprior(class = "cosy", ls = px, lb = "0", ub = "1")
   }
+  # changed ac tag
   if (has_ac_latent_residuals(bterms)) {
     prior <- prior +
       brmsprior(def_scale_prior, class = "sderr", ls = px, lb = "0")
   }
+  # if (parameterize_ac_effects(bterms)) {
```
paul-buerkner, line 1003: remember to remove this commented out code

R/stan-predictor.R
```diff
@@ -1249,23 +1250,42 @@ stan_ac <- function(bterms, data, prior, threads, normalize, ...) {
     if (is.btnl(bterms)) {
       stop2(err_msg, " for non-linear models.")
     }
-    str_add(out$par) <- glue(
-      "  vector[N{resp}] zerr{p};  // unscaled residuals\n"
-    )
+    #changed ac effects tag
+    if (has_explicit_time &&
```
paul-buerkner, line 1254: put if statements in one line if possible

R/stan-predictor.R
```diff
@@ -1249,23 +1250,42 @@ stan_ac <- function(bterms, data, prior, threads, normalize, ...) {
...
+    }
     str_add_list(out) <- stan_prior(
       prior, class = "sderr", px = px, suffix = p,
       comment = "SD of residuals", normalize = normalize
     )
-    str_add(out$tpar_def) <- glue(
-      "  vector[N{resp}] err{p};  // actual residuals\n"
-    )
+    # changed ac effects tag
+    if (has_explicit_time &&
```
paul-buerkner, line 1270: see above

R/stan-predictor.R
```diff
@@ -1377,20 +1397,40 @@ stan_ac <- function(bterms, data, prior, threads, normalize, ...) {
...
-    str_add(out$tdata_def) <- glue(
-      "  int max_nobs_tg{p} = max(nobs_tg{p});",
-      "  // maximum dimension of the autocorrelation matrix\n"
-    )
+    # changed ac tag
+    if (has_explicit_time & has_ac_latent_residuals) {
+      str_add(out$data) <- glue(
+        "  int<lower=1> N_latent_err{p};\n",
+        "  int<lower=1> max_time_span{p};\n",
+        # "  int<lower=1> begin_err_gr{p}[N_tg{p}];\n",
```
paul-buerkner, line 1405: remember to remove the commented out code

R/stan-predictor.R
```diff
@@ -1406,16 +1446,34 @@ stan_ac <- function(bterms, data, prior, threads, normalize, ...) {
...
+        "  // compute residual covariance matrix\n",
+        "  chol_cor{p} = cholesky_cor_{cor_fun}({cor_args}, max_nobs_tg{p});\n"
+      )
+    }
+    # changed ac tags
+    if (has_ac_latent_residuals) {
+      if (has_explicit_time) {
+        str_add(out$tpar_comp) <- glue(
+          "  // compute correlated time-series residuals\n",
+          "  err_tp{p} = scale_time_err_t(",
```
paul-buerkner, line 1466: can we give "scale_time_err_t" a more informative suffix than "_t"?

paul-buerkner (changes requested):
I think this PR looks quite good already. Except for the outstanding live code reviews via a call, my comments are minor I think.

paul-buerkner:
@mpdylan Is this PR still something you are working on?

mpdylan:
Hi @paul-buerkner -- I was pulled away from this for some time due to another project, but I'd like to resume working on it. Are there any specific changes to the current code that I should be aware of before getting back to it?

paul-buerkner:
Thank you for coming back to this issue! As far as I can remember, in our last meeting, I recommended changing the code structure to some degree to align with other autocorrelation features I had implemented after you started working on this feature. I still feel sorry for putting extra work on you because of this. Did you already change your PR in this direction? I just want to make sure that we can align different features and structures as much as possible to reduce the overall maintenance burden on brms.

mpdylan:
Yes, I remember that as well. I was beginning to work on that when I had to pause, but I'll review my notes and the current state of my fork this week and see how much is left to be done.

PR #1435, opened 2022-12-07
Unstructured (auto)correlation terms

paul-buerkner:
This PR implements unstructure autocorrelation matrix terms as discussed with @wds15. May I ask you to check out and test this new feature?

Here is a simple example:
```R
library(brms)
fit <- brm(count ~ Trt + unstr(visit, patient), data = epilepsy)
summary(fit)
```


wds15:
Great to see this feature landing in brms... will test these days and let you know my findings.

wds15:
Working all well so far. I tried homogeneous and also figured how to get to the heterogeneous version. Nice. One fail I found is with threading:

image

I am still trying out things.

wds15:
The unstructured works for factor levels... but it seems as if the drop_unused_levels=FALSE option is ignored. See the example where the fixed effects are correctly being sampled for a factor level not in the data (but defined as part of the factor) whereas the unstructured term is missing the respective visit:

image

wds15:
One small implementation note: You seem to pass in index spans over the data with being_tg and end_tg. These indicate start / stop indexes in the data-set per patient. This works fine as is...it's just that I switched to a more compact representation of such a thing in my own code. So instead of doing

begin: 1, 5, 7
end: 4, 6, 9

for coding 3 groups (1-4, 5-6, 7-9) you can have a single slice variable defined as:

slice: 1, 5, 7, 9

then you index group 1 with slice[1] : slice[2]-1 and group i is indexed with slice[I] : slice[i + 1]-1... so slice is an array of length n+1 if you have n groups.

There is no real gain from doing other than nicer data-structures and things are more compact.

paul-buerkner:
thanks for checking it out! Threading is not supposed to work at the moment, and I have added an informative error message. drop_unused_levels now works as expected.

I agree the slice representation would be cleaner. I will add a todo for this and adjust when I have time.

wds15:
Performancewise it really hurts to not vectorise the calls. That is tedious to do in this case as you'd need to treat each pattern of data in a different block (one case for all being present and one for a certain missingness pattern)... but it would be very worth the effort of coding.

As it's written right now we get quite large AD tapes... reduce_sum would split these into smaller chunks leading to better memory locality. So reduce_sum itself should lead to speed-ups is what I'd expect here... but it's not trivial with the index wrangling in this case, of course.

wds15:
and in a few places the comment is not correct:

```
  /* multi-student-t log-PDF for time-series covariance structures
   * in Cholesky parameterization and assuming heterogenous variances
   * Args:
   *   y: response vector
   *   nu: degrees of freedom parameter
   *   mu: mean parameter vector
   *   sigma: scale parameter
   *   chol_cor: cholesky factor of the correlation matrix
   *   se2: square of user defined standard errors
   *     should be set to zero if none are defined
   *   nobs: number of observations in each group
   *   begin: the first observation in each group
   *   end: the last observation in each group
   * Returns:
   *   sum of the log-PDF values of all observations
   */
  real student_t_time_hom_lpdf(vector y, real nu, vector mu, real sigma,
                               matrix chol_cor, data vector se2, int[] nobs,
                               int[] begin, int[] end) {
```

here the homogeneous function is referred to as heterogeneous.

paul-buerkner:
Thanks! I agree reduce_sum could be useful here, but I don't have the time to work on this in the foreseable future, given the complexity of this feature, as you say.

I have fixed the doc typos. Are there more things you want to check or try out before me merging this PR?

wds15:
Björn wanted run some tests these days. I am from my end good for now... looking at the Stan code there are some things which I may try to make things faster, but that can be done later. Maybe I even get my head into reduce_sum stuff, but that's to be seen as the brms code for that isn't too simple as you know.

Unless you want to merge quickly, maybe have Björn test it once more (he intends to couple this with a non-linear model).

paul-buerkner:
Perfect, sounds good. I will wait for Björns feedback then.

wds15:
Uh... now I see why `reduce_sum` will be a mess: You cannot split the data at any data rows. The splits must be aligned with the groupings/visits. Yack.

paul-buerkner:
yes, exactly.

wds15 ***@***.***> schrieb am Do., 8. Dez. 2022, 15:55:

wds15:
Using the idea of re-using the cholesky factor when it is possible speeds up on my mini example things by a factor of 2:

```stan
  real normal_time_hom_flex_lpdf(vector y, vector mu, real sigma, matrix chol_cor,
                                 data vector se2, int[] nobs, int[] begin,
                                 int[] end, int[,] Jtime) {
    int I = size(nobs);
    int has_se = max(se2) > 0;
    vector[I] lp;
    int have_lp[I] = rep_array(0, I);
    matrix[rows(chol_cor), cols(chol_cor)] Cov = sigma^2 * multiply_lower_tri_self_transpose(chol_cor);
    int i = 1;
    while(sum(have_lp) != I) {
      int iobs[nobs[i]] = Jtime[i, 1:nobs[i]];
      matrix[nobs[i], nobs[i]] L;
      if (has_se) {
        // need to add 'se' to the covariance matrix itself
        L = cholesky_decompose(add_diag(Cov, se2[begin[i]:end[i]]));
      } else {
        L = diag_pre_multiply( rep_vector(sigma, nobs[i]), chol_cor[iobs, iobs]);
      }
      lp[i] = multi_normal_cholesky_lpdf(y[begin[i]:end[i]] | mu[begin[i]:end[i]], L);
      have_lp[i] = 1;
      // find all additional cases where we have the same visit pattern
      for(j in i:I) {
        if(is_equal(Jtime[j], Jtime[i]) == 1 && have_lp[j] == 0) {
          have_lp[j] = 1;
          if (has_se) {
            // need to add 'se' to the covariance matrix itself
            L = cholesky_decompose(add_diag(Cov, se2[begin[i]:end[i]]));
          }
          lp[j] = multi_normal_cholesky_lpdf(y[begin[j]:end[j]] | mu[begin[j]:end[j]], L);
        }
      }
      while(have_lp[i] == 1 && i != I) i += 1;
    }
    return sum(lp);
  }
 ```

I hope this code covers the full generality of things. The speedup of 2 is obviously for the case when not using a fixed se... as then you have to do the cholesky decomposition every data row...

**Comment by wds15** [comment]
The above logic can likely go to the other cases as well. And `add_diag` should be used instead of adding diagonal matrices. So instead of

```stan
      if (has_se) {
        // need to add 'se' to the covariance matrix itself
        L = multiply_lower_tri_self_transpose(L);
        L += diag_matrix(se2[begin[i]:end[i]]);
        L = cholesky_decompose(L);
      }
```

it's better to do 

```stan
      if (has_se) {
        // need to add 'se' to the covariance matrix itself
        L = add_diag(multiply_lower_tri_self_transpose(L), se2[begin[i]:end[i]]);
        L = cholesky_decompose(L);
      }
```

This way you avoid putting all those 0's onto the AD stack... which is actually not too much of a problem here, since the se's are data. Still, better to get used to `add_diag`, which is in stan since 2.21... which is ok?

wds15:
On the possibility to parallelise this I did a bit of investigation and here is what I think could work in generality even in brms hopefully. The key issue is that we have to break down things by groups as defined in the data. However, the current pdf functions already only access the (sorted) data via a few index containers. Hence, we can simply subset the index containers and then reuse the existing lpdf and all will work out just fine. I did not know how to modify the partial sum function generated from brms, which is why I introduced in the model block a parallel flavour of the lpdf like so:

```stan
  // likelihood including constants
  if (!prior_only) {
    // initialize linear predictor term
    vector[N] mu = rep_vector(0.0, N);
    mu += Intercept + Xc * b;
    target += normal_time_hom_flex_parallel_lpdf(Y | mu, sigma, Lcortime, se2, nobs_tg, begin_tg, end_tg, Jtime_tg);
  }
```

Then I added these two functions:

```stan
  real normal_time_hom_flex_partial_lpmf(int[] seq, int start, int end,
                                         data vector y, vector mu, real sigma, matrix chol_cor,
                                         data vector se2, int[] nobs_tg, int[] begin_tg,
                                         int[] end_tg, int[,] Jtime_tg) {
    // subset the indexing things to the groups we take the partial
    // sum over
    return normal_time_hom_flex_lpdf(y| mu, sigma, chol_cor, se2, nobs_tg[seq], begin_tg[seq], end_tg[seq], Jtime_tg[seq]);
  }
  real normal_time_hom_flex_parallel_lpdf(data vector y, vector mu, real sigma, matrix chol_cor,
                                          data vector se2, int[] nobs_tg, int[] begin_tg,
                                          int[] end_tg, int[,] Jtime_tg) {
    int Ng = size(nobs_tg);
    int seq[Ng] = sequence(1, Ng);
    return reduce_sum(normal_time_hom_flex_partial_lpmf, seq, 50, y, mu, sigma, chol_cor, se2, nobs_tg, begin_tg, end_tg, Jtime_tg);
  }
```

Note that the `normal_time_hom_flex_lpdf` is not changed at all. The trick is really to pass it the full data while just subsetting the index containers which code how the data is being accessed. From running an example model, I do get the same posterior back as compared to the serial version. For now I also hard-coded the grain size of 50 in here... but ok.

The above approach could possibly even be integrated into brms by way of defining additional parallel variants of the respective unstructured covariance structures lpdfs... which one could likely auto-generate.

Attached is the full Stan model and an example data set for it as json:

[brms-mmrm-dev-rs-special_json.txt](https://github.com/paul-buerkner/brms/files/10206973/brms-mmrm-dev-rs-special_json.txt)
[brms-mmrm-dev-rs-special_stan.txt](https://github.com/paul-buerkner/brms/files/10206974/brms-mmrm-dev-rs-special_stan.txt)

If not in this PR, then maybe this should be filed as an issue for the future?

wds15:
And one q from Björn was if it is possible to have the covariance matrix vary by treatment group? So something like:

```R
library(brms)
fit <- brm(count ~ Trt + unstr(visit | Trt, patient), data = epilepsy)
summary(fit)
```

I can see how I can vary the `sigma` by treatment group and possibly even by visit, but not the correlation matrix.

bjoernholzhauer:
I tried a few things including and things worked as expected, looks good. Within-chain-parallelization would indeed be great. All of the below are "for information only" comments on things I noticed, rather than anything that needs to be addressed for this pull request:

-  I could not figure out how to allow for a different correlation structure for each treatment group. That's something we often do (but I'm not sure that's even specific to this particular pull request, presumably the same with e.g. AR(1)?) If that's not possible, probably something for a separate PR. As @wds15 already pointed out separate SD/Variance per visit (and/or treatment group) is already possible via e.g. `sigma ~ visit` (or more complex things like `visit + treatment` etc. as desired).
- Models specified like `1 + week + dose + BASE + dose:week + BASE:week` worked great, while `0 + week + dose + BASE + dose:week + BASE:week` had big convergence/sampling issues. The issue I had was that I didn't specify priors for the latter model (while a prior is automatically put on the intercept). The moment I put some priors (something like `prior(class="b", student_t(3, 0, 10))`, which for my specific simulated data was a very weak prior), the problems went away. So, may be the case that when one allows so much flexibility, then at least weak priors are very important.
- I tried combining ordinal data over time with this (I treated baseline and week as monotonic effects) and the model struggled 
to sample a bit, which I suspect is more to do with the difficulty of identifying the model well and not an implementation problem.

paul-buerkner:
Thank you for your additional input and work!

I see three new things here:
1) reusing cholesky factors: I think this can be added to the current PR and I will work on this.
2) reduce_sum: highly desireble but non trivial to integrate into the current brms threading code. requires more thought and work. Better to put into a separate PR.
3) autocorrelation parameters varying across some grouping variable. Good idea, but should go into a separate PR to since it should be added for all autocorrelation structures not just UNSTR.

wds15:
Sounds great. I am happy to file new issues for point 2 & 3. Then we can discuss there what's possible / what makes sense in the brms framework.

What about the use of `add_diag`? I recall that you like to support Stan versions below 2.21... for which one would need a workaround. So, we leave that alone for now?

wds15:
Separate functions for `se` stuff makes sense to me, yes. I wanted to really keep the same signature... we should do whatever is easier to maintain going forward.

paul-buerkner:
I was about to change it to `add_diag` but now I think it is safer to not do it for now given that the speed improvement should be minimal in this case. I will leave it be for now, but add a TODO to remind me of changing this once we have a much later rstan version on CRAN that works reliably.

I will see how much effort adding the _se stuff is in a separate function. it will surely simplify the code a lot at the expense of some duplication.

wds15:
Let me ping @bgoodri on the matter of updating rstan on CRAN... Stan is by now diverging by an enormous amount.

wds15:
This version looks great to me.

bjoernholzhauer:
I ended up getting some compiler errors when I tried to combine a non-linear formula with an unstructured covariance matrix. Here's some code to generate the issue I faced:
```
library(tidyverse)
library(brms)

doses <- c(0, 10, 20, 40)
patients_per_arm <- 20
weeks <- c(0,2,4,8,12)
et50 <- 4
ed50 <- 10
emax <- 5

set.seed(123)
sim <- expand_grid(dose=doses, 
            patient=1:patients_per_arm,
            week=weeks) %>%
  mutate(re = rep(rnorm(n=length(doses)*patients_per_arm), 
                  each=length(weeks)),
         USUBJID = paste0(dose,"_",patient),
         AVAL = rnorm(n=length(doses)*patients_per_arm*length(weeks),
                      mean=0, 
                      sd=case_when(
                        week==0~0.5,
                        week==2~0.625,
                        week==4~0.75,
                        week==8~1.0,
                        week==12~1.25)) + re + emax * week^3/(week^3+et50^3) * dose/(dose+ed50) ) %>%
  dplyr::select(-re,-patient)

sim <- sim %>%
  filter(week>0) %>%
  left_join(sim %>%
              filter(week==0) %>%
              rename(BASE=AVAL) %>%
              dplyr::select(USUBJID, BASE),
            by=c("USUBJID")) %>%
  mutate(CHG = AVAL - BASE,
         week = factor(week, levels=weeks),
         dose = factor(dose, levels=doses)) 
```
I can successfully fit a model like this one (essentially I'm fitting something very close to the data generating model, where there is an Emax curve for the treatment curve both across doses and across time with 50% of the maximum effect `Emax` reached after time `ET50` and at dose `ED50` with the steepness of the curve going towards its plateau determined by the parameters `h` and `ht`, respectively):
```
brmfit5 <- brm(
  formula=bf( CHG ~ E0 + Emax * ndose^h/(ndose^h + ED50^h) * nweek^ht/(nweek^ht + ET50^ht),
              nlf(h ~ exp(logh)), nlf(ED50 ~ exp(logED50)),
              nlf(ht ~ exp(loght)), nlf(ET50 ~ exp(logET50)),
              E0 ~ 0 + week + BASE + BASE:week **+ (1|USUBJID)**,
              Emax ~ 1, 
              logED50 ~ 1, logh ~ 1,
              logET50 ~ 1, loght ~ 1,
              nl=TRUE, 
              family=gaussian(), #autocor=~unstr(week, USUBJID), 
              sigma ~ week),
  data = sim %>% mutate(nweek = as.numeric(week), ndose = as.numeric(dose)),
  control = list(adapt_delta=0.99),
  cores=4,
  prior = c(prior(class="b", student_t(3,0,10), nlpar="E0"),
            prior(normal(0,1), nlpar="logh"),
            prior(normal(0,1), nlpar="loght"),
            prior(student_t(3,0,10), nlpar="Emax"),
            prior(normal(2.5,2.5), nlpar="logED50"),
            prior(normal(0.5,5), nlpar="logET50")),
  backend="cmdstanr")
```
However, when I instead ask for an unstructured covariance matrix, I get a compiler error:
```
brmfit6 <- brm(
  formula=bf( CHG ~ E0 + Emax * ndose^h/(ndose^h + ED50^h) * nweek^ht/(nweek^ht + ET50^ht),
              **autocor=~unstr(week, USUBJID),** 
              nlf(h ~ exp(logh)), nlf(ED50 ~ exp(logED50)),
              nlf(ht ~ exp(loght)), nlf(ET50 ~ exp(logET50)),
              E0 ~ 0 + week + BASE + BASE:week,
              Emax ~ 1, 
              logED50 ~ 1, logh ~ 1,
              logET50 ~ 1, loght ~ 1,
              nl=TRUE, 
              family=gaussian(), 
              sigma ~ week),
  data = sim %>% mutate(nweek = as.numeric(week), ndose = as.numeric(dose)),
  control = list(adapt_delta=0.99),
  cores=4,
  prior = c(prior(class="b", student_t(3,0,10), nlpar="E0"),
            prior(normal(0,1), nlpar="logh"),
            prior(normal(0,1), nlpar="loght"),
            prior(student_t(3,0,10), nlpar="Emax"),
            prior(normal(2.5,2.5), nlpar="logED50"),
            prior(normal(0.5,5), nlpar="logET50")),
  backend="cmdstanr")
```
This results in
```
Compiling Stan program...
Semantic error in 'C:/Users/holzhbj1/AppData/Local/Temp/1/RtmpqoJjfg/model-9748687a1a88.stan', line 123, column 14 to column 106:
   -------------------------------------------------
   121:      }
   122:      sigma = exp(sigma);
   123:      target += normal_time_het_flex_lpdf(Y | mu, sigma, Lcortime, se2, nobs_tg, begin_tg, end_tg, Jtime_tg);
                       ^
   124:    }
   125:    // priors including constants
   -------------------------------------------------

A returning function was expected but an undeclared identifier 'normal_time_het_flex_lpdf' was supplied.

mingw32-make.exe: *** [make/program:50: C:\Users\holzhbj1\AppData\Local\Temp\1\RtmpqoJjfg\model-9748687a1a88.hpp] Error 1

Error: An error occured during compilation! See the message above for more information.
```
Apologies, if I messed up my syntax, but I thought this ought to work the way I specified it.

paul-buerkner:
I have now updated the PR to optimize the Stan code (point 1) and moved the se-likelihood to separate functions. This required some code duplication but should leader to easier and safer optimizing of individual cases. @wds15 would you mind going over the Stan code chunks and checking whether those are correct, both in terms of the code optimization and the handling of se? 

@bjoernholzhauer I have also fixed the non-linear issue.

inst/chunks/fun_normal_time_se.stan
```diff
@@ -0,0 +1,106 @@
...
+   */
+  real normal_time_hom_se_lpdf(vector y, vector mu, real sigma, data vector se2,
+                               matrix chol_cor, int[] nobs, int[] begin, int[] end) {
+    int I = size(nobs);
+    vector[I] lp;
+    for (i in 1:I) {
+      matrix[nobs[i], nobs[i]] L;
+      L = sigma * chol_cor[1:nobs[i], 1:nobs[i]];
+      // need to add 'se' to the covariance matrix itself
+      L = multiply_lower_tri_self_transpose(L);
```
wds15, line 24: In this homogeneous case you can pull the sigma * L and the multiply_lower_tri_self_transpose out of the loop. This is always the same. Just subset the respective covariance matrix for each row to what is observed is what is needed per data group.

inst/chunks/fun_normal_time_se.stan
```diff
@@ -0,0 +1,106 @@
...
+  real normal_time_hom_se_flex_lpdf(vector y, vector mu, real sigma, data vector se2,
+                                    matrix chol_cor, int[] nobs, int[] begin,
+                                    int[] end, int[,] Jtime) {
+    int I = size(nobs);
+    vector[I] lp;
+    matrix[rows(chol_cor), cols(chol_cor)] Cor;
+    Cor = multiply_lower_tri_self_transpose(chol_cor);
+    for (i in 1:I) {
+      int iobs[nobs[i]] = Jtime[i, 1:nobs[i]];
+      matrix[nobs[i], nobs[i]] Cov = sigma^2 * Cor[iobs, iobs];
```
wds15, line 76: multiplication with sigma^2 can happen outside of the loop

inst/chunks/fun_scale_time_err.stan
```diff
@@ -21,26 +21,32 @@
...
+     int has_lp[I] = rep_array(0, I);
+     int i = 1;
+     while (sum(has_lp) != I) {
        int iobs[nobs[i]] = Jtime[i, 1:nobs[i]];
-       err[begin[i]:end[i]] =
-         sderr * cholesky_decompose(Cor[iobs, iobs]) * zerr[begin[i]:end[i]];
-     }
-     return err;
-   }
+       matrix[nobs[i], nobs[i]] L = cholesky_decompose(Cor[iobs, iobs]);
```
wds15, line 39: uhh..hurts to read this... maybe we can have an if statement for identifying the case whenever iobs[i] == 1,2,3,...,rows(chol_cor)? In that case you can just take L=chol_cor and can avoid a full cholesky_decompose, which is super expensive and this is probably even a rather common case?

wds15:
Stan code looks good. I left a few more small optimisations as comment. I also opened two more issues for following up on the points we came across in this PR (parallel & by group error structures).

Re-using the cholesky makes my mini examples run substantially faster already. 12s instead of 20s, which is great news for real big problems.

Given that we now process multiple data items at once in the cholesky reuse thing, one could imagine to even go for a vectorised call... but I don't see immediately how to code that efficiently (you need to munge data, which is impossible to do easily of the top of my head). In case there is a good way, we can come back to this.

wds15:
Alright... right after sending off my last comment, I had the idea how we can make use of a fully vectorized call across all grouped observations. The speedup is another 2x for me! So it's really worth doing this.

The problem is that the data is not quite formatted as we need it. While `brms` has things stored in a flat column vector, the vectorized version of the multi normal distributions require that we have things as array of vectors. To reformat things efficiently one has to be very careful so that data stays as data to Stan. The way to achieve that is to do the reformatting in a separate function. This way the separate function gets called with data stuff only as arguments and then we can do the data munging "for free" in case of the data. Here is on of the functions recoded this way:

```stan
  /* grouped data stored linearly in "data" as indexed by begin and end
   * is repacked to be stacked into an array of vectors.
   * end
   */
  vector[] stack_vectors(vector long_data, int n, int[] stack, int[] begin, int[] end) {
    int S = sum(stack);
    int G = size(stack);
    vector[n] stacked[S];
    int i = 1;
    int j = 1;
    while(i != G+1) {
      if(stack[i] == 1) {
        stacked[j] = long_data[begin[i]:end[i]];
        j += 1;
      }
      i += 1;
    }
    return stacked;
  }
  
  /* multi-normal log-PDF for time-series covariance structures
   * in Cholesky parameterization and assuming homogoneous variances
   * allows for flexible correlation matrix subsets
   * Deviating Args:
   *   Jtime: array of time indices per group
   * Returns:
   *   sum of the log-PDF values of all observations
   */
  real normal_time_hom_flex_lpdf(vector y, vector mu, real sigma, matrix chol_cor,
                                 int[] nobs, int[] begin, int[] end, int[,] Jtime) {
    int I = size(nobs);
    //vector[I] lp;
    real lp = 0.0;
    int has_lp[I] = rep_array(0, I);
    int i = 1;
    while (sum(has_lp) != I) {
      int iobs[nobs[i]] = Jtime[i, 1:nobs[i]];
      int lp_terms[I] = rep_array(0, I);
      matrix[nobs[i], nobs[i]] L = diag_pre_multiply(rep_vector(sigma, nobs[i]), chol_cor[iobs, iobs]);
      //lp[i] = multi_normal_cholesky_lpdf(y[begin[i]:end[i]] | mu[begin[i]:end[i]], L);
      has_lp[i] = 1;
      lp_terms[i] = 1;
      // find all additional groups where we have the same timepoints
      for (j in (i+1):I) {
        if (has_lp[j] == 0 && is_equal(Jtime[j], Jtime[i]) == 1) {
          //lp[j] = multi_normal_cholesky_lpdf(y[begin[j]:end[j]] | mu[begin[j]:end[j]], L);
          has_lp[j] = 1;
          lp_terms[j] = 1;
        }
      }
      lp += multi_normal_cholesky_lpdf(stack_vectors(y, nobs[i], lp_terms, begin, end) | stack_vectors(mu, nobs[i], lp_terms, begin, end), L);
      while (has_lp[i] == 1 && i != I) {
        i += 1;
      }
    }
    return lp;
  }
```

The above logic can be applied to all other functions... and again - the simulated fake example speeds up by a whopping factor of 2x.

paul-buerkner:
Thanks for the code review and the vectorization code! I have now made another pass over all the Stan code, made it more consistent, and optimized some further things. @wds15 can you do another round of code review please? 

One thing we need to be very careful of is not to take arbitrary subsets of cholesky factors of correlation matrices. This is only valid if we subset the first J rows:

```R
Cor <- matrix(c(1, 0.5, 0.3, 0.5, 1, 0.1, 0.3, 0.1, 1), 3, 3)
Cor

L <- t(chol(Cor))
L %*% t(L)  # = Cor

L_i <- L[1:2, 1:2]
L_i %*% t(L_i)  # = Cor[1:2, 1:2]

L_j <- L[c(2, 3), c(2, 3)]
L_j %*% t(L_j)  # != Cor[c(2, 3), c(2, 3)]
```

That is, patterns such as `chol_cor[iobs, iobs]` are not valid in general.

wds15:
Thanks for including these latest bits as well...sorry for throwing these things one at a time at you...but this thing develops. Will check the stan code now.

paul-buerkner:
No worries, we are continuously making the implementation better so that makes me :-)

wds15:
There was a mistake in my `is_equal`... here it's corrected:

```stan
int is_equal(int[] a, int[] b) {
    int n_a = size(a);
    // int n_b = size(a); // WRONG
    int n_b = size(b);
    if(n_a != n_b) return 0;
    for(i in 1:n_a) {
      if(a[i] != b[i])
        return 0;
    }
    return 1;
  }
```

inst/chunks/fun_is_equal.stan
```diff
@@ -0,0 +1,15 @@
+  // are two 1D integer arrays equal?
+  int is_equal(int[] a, int[] b) {
+    int n_a = size(a);
+    int n_b = size(a);
```
wds15, line 4: `int n_b = size(b);` is correct. Sorry for that.

inst/chunks/fun_normal_time.stan
```diff
@@ -1,71 +1,140 @@
...
+   */
+  real normal_time_het_flex_lpdf(vector y, vector mu, vector sigma, matrix chol_cor,
+                                 int[] nobs, int[] begin, int[] end, int[,] Jtime) {
+    real lp = 0.0;
+    int I = size(nobs);
+    int has_lp[I] = rep_array(0, I);
+    int i = 1;
+    matrix[rows(chol_cor), cols(chol_cor)] Cor;
+    Cor = multiply_lower_tri_self_transpose(chol_cor);
+    while (sum(has_lp) != I) {
```
wds15, line 112: we could avoid the `sum(has_lp)` check here, which is a sum over all groups. Instead we could count i to I+1 and abort if that value is reached... not a huge time saver though as we likely won't apply this to huge data, but we can avoid the repeated sum over `has_lp` (which we still need to flag which terms have been collected into `lp` as we go).

inst/chunks/fun_normal_time_se.stan
```diff
@@ -0,0 +1,101 @@
...
+                                    int[] end, int[,] Jtime) {
+    real lp = 0.0;
+    int I = size(nobs);
+    matrix[rows(chol_cor), cols(chol_cor)] Cov;
+    Cov = multiply_lower_tri_self_transpose(sigma * chol_cor);
+    for (i in 1:I) {
+      int iobs[nobs[i]] = Jtime[i, 1:nobs[i]];
+      matrix[nobs[i], nobs[i]] Cov_i = Cov[iobs, iobs];
+      Cov_i += diag_matrix(se2[begin[i]:end[i]]);
+      lp += multi_normal_lpdf(y[begin[i]:end[i]] | mu[begin[i]:end[i]], Cov_i);
```
wds15, line 72: I think here it's better to use the old pattern to store for each item the lp into an array and then call `sum` on the vector. If you write it with `lp+=`, then you trigger a sum operator to land on the AD tree each time you do it... that's avoided by the vectored thing. So if we know that we have `I` items, then I'd pre-allocate and then sum over it whereas in the vectored code this isn't such a big deal to worry about it.

inst/chunks/fun_scale_time_err.stan
```diff
@@ -13,8 +14,48 @@
                          int[] nobs, int[] begin, int[] end) {
      vector[rows(zerr)] err;
      for (i in 1:size(nobs)) {
-       err[begin[i]:end[i]] =
-         sderr * chol_cor[1:nobs[i], 1:nobs[i]] * zerr[begin[i]:end[i]];
+       matrix[nobs[i], nobs[i]] L_i;
+       L_i = sderr * chol_cor[1:nobs[i], 1:nobs[i]];
```
wds15, line 18: don't we need to recalculate the cholesky here if we subset??

wds15:
Really nice that you can now recycle the cholesky also for the heterogeneous case. That's great (the most common case we will use). I left a few more comments and I'd recommend to go back to the vector for `lp` and then sum over it whenever we cannot vectorise things. That should be better in terms of the AD tape usage.

I realised that these improvements are also beneficial to the other correlation models (cosy, AR)... nice.

inst/chunks/fun_is_equal.stan
paul-buerkner: Thanks.

inst/chunks/fun_normal_time_se.stan
paul-buerkner: Thanks, will revert.

inst/chunks/fun_normal_time.stan
```diff
@@ -1,71 +1,140 @@
...
+      if (is_equal(iobs, sequence(1, rows(chol_cor)))) {
+        // all timepoints are present in this group
+        L_i = chol_cor;
+      } else {
+        // arbitrary subsets cannot be taken on chol_cor directly
+        L_i = cholesky_decompose(Cor[iobs, iobs]);
       }
-      lp[i] = multi_normal_cholesky_lpdf(
-        y[begin[i]:end[i]] | mu[begin[i]:end[i]], L
+      L_i = sigma * L_i;
```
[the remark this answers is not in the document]
paul-buerkner: Good point. Will change.

inst/chunks/fun_normal_time.stan
paul-buerkner: Does Stan evaluate `y` of `x && y` is `x`already is FALSE?

paul-buerkner: Asked differently, can I safely use 

```Stan
while (i <= I && has_lp[i] == 1) {
        i += 1;
      }
```

as this will mean i becoming I+1 at some point

wds15:
Here is a better and more detailed derivation of how to remove a row&column from a Cholesky: https://normalsplines.blogspot.com/2019/02/algorithms-for-updating-cholesky.html

It's worthwhile to do, since we go from n^3 to n^2 effort (n being the dimension).

inst/chunks/fun_scale_time_err.stan
paul-buerkner: To my understanding, taking the first J row/col subset of a cholesksy factor is valid. Or am I mistaken?

wds15: I see. Yes, that should work.

inst/chunks/fun_normal_time.stan
wds15: 1) yes, that's correct, see https://mc-stan.org/docs/2_25/functions-reference/logical-functions.html#boolean-operator-short-circuiting

2) yes, this logic should work.

paul-buerkner: Tried it out and seems to work.

paul-buerkner:
Thank you again! Having this cholesky factor subsetting efficiently in Stan math would be lovely! And I am glad if I don't have to maintain it inside of brms :-D

I have now incorporated all the remaining comments from above. The code seems to be really fast now. 

Any other things to change before I merge?

wds15:
and I checked: Eigen supports these rank updates to the Cholesky factors out of the box. So it really belongs to Stan-math to do these.

For me - I am good with this implementation now. It great to have it and it indeed got quite a bit faster, which is great. Can't wait to see this land on our production systems! Björn will run some of his examples once more and leave a note here once these are through. I think these are great stress tests for the code - so let's just wait for his note here once done.

paul-buerkner:
Sounds good!

bjoernholzhauer:
I tried everything, again, I had previously tried and it all works nicely now. Looks good to me.

wds15:
Great! Then this can go in from my view.

paul-buerkner:
Great, Thank you for all your help! Merging now.

Issue #1429, opened 2022-11-25
Feature request: adding a hurdle cumulative family


sjwild:
I would like to know if there is any interest in adding a hurdle cumulative-type model. In a lot of survey research, for example, you will have a 5-point Likert scale along with a "Don't know" response option. A hurdle cumulative-type model would allow joint modeling of the "Don't know" responses and the Likert scale.

I have an example using brms and then modifying the resulting Stan code [here](https://github.com/sjwild/hurdle_ordered). (Note I did this using an "identity" link, which I will obviously fix).

I'm happy to initiate a pull request if there is interest.

paul-buerkner:
This could be an interesting family indeed. How much effort is it to first code it up as a brms `custom_family`? That way, it is already easily useable by other people and from there we can see if we want to support it as a built-in family too?

paul-buerkner:
Exciting, thank you! About the nlevels issue, I will think of a nice way to tell brms about the nlevels-2 requirement such that the workaround via an empty brmsfit object and stan code editing is no longer necessary.

sjwild:
That'd be awesome. I _think_ you can do it by modifying the `extract_nthres()` function [here](https://github.com/paul-buerkner/brms/blob/b83c0efaab0e2e9c7ab033635447acd5f3d6fc59/R/data-response.R#L604) to test the `family` argument in `brmsformula`. If `family` is either `hurdle_cumulative` or `ord_beta` (or whatever they are going to be called) then it would generate the correct number of thresholds.

Not sure how that might work if you do something like `y | thres(...) ~ .` Maybe it would be possible to specify an error message or something.

paul-buerkner:
I think, we could add another family `special` that tells how the number of categories relates to the number of thresholds. The default would be `-1` but we could flexibly set it to other values for new families. Once you are done with the custom family except for this point, let me know and I will see how to best implement this new special.

sjwild:
Works for me. I will let you know once I have it finalized.

paul-buerkner:
That looks good! Do you think you can prepare a PR implementing this as a built-in family? Asked differently, what would you need form my side to prepare such a PR?

sjwild:
I'll take a look through a few of the closed PRs to see how it was handled for other families. The one issue that will still need to be addressed is adding another `special` the allows us to modify the number of thresholds vs categories, so that we there are two fewer thresholds than categories.

paul-buerkner:
[...]

PR #1448, opened 2023-01-14
add hurdle_cumulative family to brms

sjwild:
This pull request implements the hurdle_cumulative family discussed in [issue 1429](https://github.com/paul-buerkner/brms/issues/1429).

I mostly followed the code for the ordinal families. I tried to follow [pull request 1310](https://github.com/paul-buerkner/brms/pull/1310), as it seem fairly complete. 

To demonstrate the family works, I produce a short [Rmd document](https://github.com/sjwild/hurdle_ordered/blob/main/test_supported_hurdle_cumulative_family.Rmd). You can see the html output [here](https://htmlpreview.github.io/?raw.githubusercontent.com/sjwild/hurdle_ordered/main/test_supported_hurdle_cumulative_family.html).

To get one of the supported link functions to work, it was necessary to edit `fun_cauchit.stan`. The function expects a `vector`, but the model block uses a for loop, so it is getting a `real` instead. This produces errors. 

One other thing comes up for `posterior_predict_hurdle_cumulative()` To get the proper proportion of values for `hu`, it's necessary to have `theta < hu`, which seems to run counter to other hurdle and zero-inflated families. 

Let me know if there are any changes I need to make or further tests/docs. 

This has been a fun learning experience, and I want to keep contributing to brms, so any advice/etc is helpful!

Steve


paul-buerkner:
Thanks! Upon a quick look, the PR already looks quite good. I will make some edits mostly to align the code structure with the rest of the brms code base.

Specific comments:

Changing the `cauchit.stan` functions that way will break other code that expects this to be a vector. I will check what can be done about it. In the worst case, I will just disallow cauchit link for hurdle_cumulative for now.

> One other thing comes up for posterior_predict_hurdle_cumulative() To get the proper proportion of values for hu, it's necessary to have theta < hu, which seems to run counter to other hurdle and zero-inflated families.

I am afraid I don't understand. Can you clarify a bit please?

paul-buerkner:
Ah, I think I understand the theta < hu now. You mean that the names are awkwardly chosen in previous hurdle models and it should actually be the other way round (changing theta and hu names), right? If so, I agree. Will change it accordingly in this PR:

paul-buerkner:
Thanks for offering to fix the naming conventions! I will let you know once I am done with my minor cleaning of the code.

As for the cauchit link, indeed this currently does not work as it should for some families apparently. But then, not many people seem to use it anyway :-D I will solve this separately but for now will revert the change.

paul-buerkner:
Okay, I have done some minor cleaning to your code and also fixed the awkward naming conventions in posterior_predict functions. Tests etc. pass for me. Is there anything you want to change or add before I merge?

Thank you for providing such a high quality PR! It was a pleasure reviewing this one :-)

sjwild:
No problem!

There is nothing I want to add to this particular one right now. If there is demand I'd like to add the ability to specify a hurdle value other than 0, because for survey data such responses are often coded as 99 or something similar. As well if there's enough demand I'd like to add a zero (or other level) inflated ordinal family, but there's a few issues to think through, including how to properly specify the level so it works for both ordered factor and integer responses, properly accounting for the other link functions, etc.

paul-buerkner:
Update: The cauchit link issue is now also fixed.

sjwild:
Awesome! Now I shall never use the cauchit link again :-p

PR #1451, opened 2023-02-06
Latent projection

fweber144:
This PR adds support for projpred's latent projection which has just been merged into projpred's `master` branch (see stan-dev/projpred#372).

### DIFF
--- a/R/projpred.R
+++ b/R/projpred.R
@@ -204,53 +204,17 @@ get_refmodel.brmsfit <- function(object, newdata = NULL, resp = NULL,
     }
   } else if (latent) {
     if (family$family == "cumulative") {
-      stopifnot(!is.null(family$cats))
-      draws_mat <- as_draws_matrix(object)
-      thres_regex <- paste0("^b", usc(combine_prefix(bterms)), "_Intercept\\[")
-      thres_draws <- prepare_draws(draws_mat, variable = thres_regex,
-                                   regex = TRUE)
-      if (ncol(thres_draws) > length(family$cats) - 1L) {
-        stop2("Currently, projpred does not support group-specific thresholds ",
-              "(argument `gr` of resp_thres()).")
-      }
-      # Note: Currently, `disc` should always be constantly 1 because
-      # distributional models are not allowed here.
-      disc_regex <- paste0("^", "disc", resp, "$")
-      disc_draws <- prepare_draws(draws_mat, variable = disc_regex,
-                                  regex = TRUE)
-      latent_ilink_tmp <- function(lpreds, cl_ref,
-                                   wdraws_ref = rep(1, length(cl_ref))) {
-        thres_agg <- projpred::cl_agg(thres_draws, cl = cl_ref,
-                                      wdraws = wdraws_ref)
-        disc_agg <- projpred::cl_agg(disc_draws, cl = cl_ref,
-                                     wdraws = wdraws_ref)
-        disc_agg <- as.vector(disc_agg)
-        lpreds_thres <- apply(thres_agg, 2, function(thres_agg_c) {
-          # Notes on dimensionalities (with S_agg = `nrow(lpreds)`):
-          # * `disc_agg` is a vector of length S_agg (because `disc` is not
-          #   predicted here),
-          # * `thres_agg` is S_agg x C_lat (with C_lat = `ncats - 1L` =
-          #   `nthres`) and thus `thres_agg_c` is a vector of length S_agg,
-          # * `lpreds` is S_agg x N (with N denoting the number of (possibly
-          #   new) observations (not necessarily the original number of
-          #   observations)).
-          disc_agg * (thres_agg_c - lpreds)
-        }, simplify = FALSE)
-        # Coerce to an S_agg x N x C_lat array:
-        lpreds_thres <- do.call(abind, c(lpreds_thres, rev.along = 0))
-        # Transform to response space, yielding an S_agg x N x C_cat array:
-        return(inv_link_cumulative(lpreds_thres, link = family$link))
-      }
-      c(args) <- list(latent_ilink = latent_ilink_tmp)
-      # Free up some memory:
-      rm(draws_mat)
+      c(args) <- list(
+        latent_ilink = get_latent_ilink_cumul(object = object, family = family,
+                                              bterms = bterms, resp = resp)
+      )
     }
-    # TODO: Add response-scale support for more families: For response-scale
-    # support, they all need a specific `latent_ilink` function; some families
-    # (those for which the response can be numeric) also require specific
-    # `latent_ll_oscale` and `latent_ppd_oscale` functions. The binomial family
-    # (and thereby also the brms::bernoulli() family) has response-scale support
-    # implemented natively in projpred.
+    # TODO: If requested by users, add response-scale support for more families:
+    # For response-scale support, they all need a specific `latent_ilink`
+    # function; some families (those for which the response can be numeric) also
+    # require specific `latent_ll_oscale` and `latent_ppd_oscale` functions. The
+    # binomial family (and thereby also the brms::bernoulli() family) has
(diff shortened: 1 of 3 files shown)


paul-buerkner:
Thanks! When will the new projpred be on CRAN? I want to keep the brms master branch CRAN releasable at any time and would not want to merge this PR if it prevented releasing brms for now.

fweber144:
> When will the new projpred be on CRAN?

I don't know yet, but the brms changes suggested here don't make brms dependent on the projpred version including the latent projection. Of course, that updated projpred version is needed for the latent projection, but all other projpred features are not affected. So even after merging this PR here, an older projpred version which excludes the latent projection can be used (for all features other than the latent projection). In particular, I was able to run brms's `R CMD check` with the current CRAN version of projpred. So concerning your comment

> I want to keep the brms master branch CRAN releasable at any time and would not want to merge this PR yet if it prevented releasing brms for now.

I think this PR should not prevent brms from being released on CRAN.

This is kind of a chicken-or-egg problem: Within projpred, we need to insert the brms version which supports the latent projection (currently, this is highlighted by a `TODO` comment in projpred, see https://github.com/stan-dev/projpred/pull/372#discussion_r1093152256). On the other hand, as you said, it would be desirable to have the latent projection in projpred's CRAN version before merging this brms PR here. So I thought the best solution would be to merge this brms PR and afterwards insert the current (GitHub) version number of brms into projpred. Even if brms would be published on CRAN before the projpred CRAN version includes the latent projection, users who might want to use the latent projection are instructed to use a projpred version > 2.3.0 (the current CRAN version) in the brms `NEWS` file (see https://github.com/paul-buerkner/brms/pull/1451/files#diff-51920e95310ebfbc1ae31709f3b95f89afffbf4f1a6e38e8b2b406e2fb6197ea), so they should realize that they need projpred's GitHub version (until the latent projection is included in a projpred CRAN version). Would you prefer to add such an instruction (to use a projpred version > 2.3.0) also to the documentation of argument `latent`?

Some technical details: The combination of this PR with projpred <= 2.3.0 is possible because `projpred::init_refmodel()`'s ellipsis (`...`, also exists in projpred <= 2.3.0) swallows up argument `latent` and any other arguments related to the latent projection that might be passed to `get_refmodel.brmsfit()`. So if users specify argument `latent` or other arguments related to it, these arguments simply become ineffective.

paul-buerkner:
Makes sense. I will review this PR and then merge.

fweber144:
Thanks! In any case, I'd like to have the latent projection in projpred's CRAN version soon, so it shouldn't take too long until it is available there.

R/projpred.R
```diff
@@ -13,8 +13,13 @@
 #' (see \code{\link[projpred:get_refmodel]{get_refmodel}} for details).
 #' If \code{NULL} (the default), \code{cvfun} is defined internally
 #' based on \code{\link{kfold.brmsfit}}.
+#' @param dis Passed to argument \code{dis} of
```
paul-buerkner, line 16: In which cases would this have to be added manually?

R/projpred.R
```diff
@@ -47,7 +52,8 @@
 #' plot(cv_vs)
 #' }
 get_refmodel.brmsfit <- function(object, newdata = NULL, resp = NULL,
-                                 cvfun = NULL, brms_seed = NULL, ...) {
+                                 cvfun = NULL, dis = NULL, brms_seed = NULL,
```
paul-buerkner, line 55: I prefer seed being kind of last. Can we do `latent, dis, brms_seed, ...` order?

R/projpred.R
```diff
@@ -194,6 +202,55 @@ get_refmodel.brmsfit <- function(object, newdata = NULL, resp = NULL,
         augdat_args_ilink = list(link = family$link)
       )
     }
+  } else if (latent) {
+    if (family$family == "cumulative") {
```
paul-buerkner, line 206: this looks like a lot of code overhead just for the cumulative family. Why is this necessary? 

And if it is necessary, can we perhaps move it to its own function to keep the code here a bit cleaner? I would like special cases to be stored away tidily rather than being around like this in the main function.

paul-buerkner (changes requested):
I have some small comments to address before merging.

R/projpred.R
fweber144: For the latent projection with the probit or the logit link, projpred can use an internal default (constantly 1 and 1.6, respectively), but for other links, there is no such internal default. Perhaps users can come up with some reasonable `dis` value(s) in the future, so it's better to keep this flexible. Apart from that, `dis` is an argument of `init_refmodel()` and previously, it wasn't possible to specify it when using `get_refmodel.brmsfit()`, so already from a conceptual point of view, I think it makes sense to keep this flexible.

R/projpred.R
fweber144: Sure. Perhaps `dis, latent, brms_seed, ...` since `dis` is a long-standing argument of `init_refmodel()` and `latent` is a new one?

R/projpred.R
fweber144: It's necessary to define argument `latent_ilink` which is a new argument within projpred, see [here](https://github.com/stan-dev/projpred/blob/543a7c3a93fcff12b2cf5fe3982323592ae4c729/R/extend_family.R#L26-L32) and [here](https://github.com/stan-dev/projpred/blob/543a7c3a93fcff12b2cf5fe3982323592ae4c729/R/extend_family.R#L131-L154). For the cumulative family, `latent_ilink` is basically `inv_link_cumulative()`, but the draws can be clustered or thinned (that's why I used `S_agg` in the inline comments).

fweber144: > And if it is necessary, can we perhaps move it to its own function to keep the code here a bit cleaner?

Yes, I can try to define `latent_ilink_tmp` outside of `get_refmodel.brmsfit()`. That will require handing over several objects (which is the reason why I didn't define it outside of `get_refmodel.brmsfit()` until now). I'll check how many and which objects are concerned.

R/projpred.R
paul-buerkner: I see. Makes sense.

R/projpred.R
paul-buerkner: okay.

R/projpred.R
paul-buerkner: Thank you! Perhaps just passing the brms object (and some projpred stuff) would be sufficient?

fweber144: Done (commit https://github.com/paul-buerkner/brms/pull/1451/commits/0ebe5ef7c87171835603cc268876f3abdd9cf11f). Could you check whether my changes are ok?

> Perhaps just passing the brms object (and some projpred stuff) would be sufficient?

Unfortunately not. At least `resp` is required, too. And I also passed `family` and `bterms` to the new `get_latent_ilink_cumul()` function, for efficiency.

R/projpred.R
fweber144: Done (commit https://github.com/paul-buerkner/brms/pull/1451/commits/c6268cdc206fa3ef1be82a1582dcaeb453b3b749).

R/projpred.R
paul-buerkner: I made some minor changes mainly due to my taste of code appearence. Can you check that I didn't break anything. Once I got your okay, I will merge this PR.

fweber144: Looks good, shouldn't break anything. I just noticed that you also removed the `rm(draws_mat)` statement. I guess this can lead to unnecessary memory usage (because `draws_mat` will continue to live in the environment of the returned function), but if you say this extra memory usage is ok, I'm fine with merging.

fweber144:
> ah I see. makes sense. I had overlooked this please add it back with a
comment for future me.

Done (commit fd090b8c).

paul-buerkner (approved):
looks good now. thank you!

fweber144:
No problem. Thanks for merging!

Issue #1452, opened 2023-02-07
Feature: Cure models


yananlong:
**Mixture cure models** are used in survival analysis when not all subjects are expected to experience the event(s) under investigation. As reviewed in [Amico and Van Keilegom (2018)](https://www.annualreviews.org/doi/full/10.1146/annurev-statistics-031017-100101), there exist two broad classes of cure models: *mixture cure models* and *promotion time/latent activation cure models*. For the former, we assume that the overall (population) survival function is a mixture of event (uncured) and cured subjects: $ S(t|x,z) = [1-p(z)] + p(z) S_u(t|x), $ where the event survival function ($ S_u $) can be parametric (e.g. Weibull) or non-parametric (Cox PH).

An example of mixture cure models can be found [here](https://github.com/yananlong/brms). I was only able to do log-normal and Weibull since log-logistic lccdf is [not available](https://github.com/stan-dev/math/issues/2657) in the stan math library, even though is more commonly used because the posterior has a closed form.

spinkney:
Is there any reason you can't just write the `log_logistic_lccdf` function yourself? It is easily written in Stan as
```stan
  real log_logistic_lccdf(real x, real alpha, real beta) {
    if (alpha <= 0) {
      reject("alpha must be > 0!");
    }
    if (beta <= 0) {
      reject("beta must be > 0!");
    }
    if (x < 0) {
      reject("x >= 0!");
    }
    
    return -log1p(x / alpha)^beta;
  }
```

PR #1453, opened 2023-02-07
Mixture cure models (Weibull and log-normal)

yananlong:
See [this issue](https://github.com/paul-buerkner/brms/issues/1452)


paul-buerkner:
Thank you! This looks like a very highy quality PR! I will go through it in the next couple of days.

yananlong:
Thanks for your prompt reply! I've actually found a couple of typos that have been fixed in a newer commit.

R/posterior_epred.R
```diff
@@ -485,6 +485,16 @@ posterior_epred_hurdle_lognormal <- function(prep) {
   with(prep$dpars, exp(mu + sigma^2 / 2) * (1 - hu))
 }
 
+posterior_epred_mixcure_lognormal <- function(prep) {
```
paul-buerkner, line 488: Why is it not possible to compute the expected value for these families?

R/stan-likelihood.R
```diff
@@ -889,6 +889,25 @@ stan_log_lik_hurdle_cumulative <- function(bterms, resp = "", mix = "",
...
 }
 
+stan_log_lik_mixcure_lognormal <- function(bterms, resp = "", mix = "", ...) {
+  p <- stan_log_lik_dpars(bterms, TRUE, resp, mix)
+  usc_logit <- stan_log_lik_dpar_usc_logit("inc", bterms)
+  lpdf <- paste0("mixcure_lognormal", usc_logit)
+  sdist(lpdf, p$mu, p$sigma, p$inc)
+}
+
+stan_log_lik_mixcure_weibull <- function(bterms, resp = "", mix = "", ...) {
```
paul-buerkner, line 899: It seems like `reqn` should always be true given the stan code of this model. If true, should the code below be changed or simplified?

inst/chunks/fun_mixcure_lognormal.stan
```diff
@@ -0,0 +1,30 @@
+/* mixcure lognormal (AFT) log-PDF of a single response
+ * identity parameterization of the incidence part
+ */
+real mixcure_lognormal_lpdf(real y, real mu, real sigma, real inc) {
```
paul-buerkner, line 4: can you add some empty lines between function definitions in both stan files? And add a little more doc at the top?

tests/local/setup_tests_local.R
```diff
@@ -1,5 +1,6 @@
 library(testthat)
 library(brms)
+library(mixcure)
```
paul-buerkner, line 3: Do we really need to load this package here?

tests/local/tests.models-6.R
```diff
@@ -0,0 +1,46 @@
...
+        control = list(adapt_delta = 0.825),
+    )
+    print(fit2, digits = 3)
+    print(prior_summary(fit2))
+    expect_range(posterior_summary(fit2)["b_Intercept", "Estimate"], 4.9, 5.1)
+    expect_range(posterior_summary(fit2)["b_inc_Intercept", "Estimate"], 0.95, 1.05)
+    
+    fit1 <- add_criterion(fit1, criterion = "loo" , moment_match = TRUE)
+    fit2 <- add_criterion(fit2, criterion = "loo" , moment_match = TRUE)
+    loo_compare(fit1, fit1)
```
paul-buerkner, line 45: I suggest adding additional tests to see if loo results are in a sensible range. Also please add tests for the posterior_predict method here (and for posterior_epred if implemented).

paul-buerkner (changes requested):
Overall, high quality PR! I have just a few comments and suggestions. What is more, I think it would make sense to add info about the mixcure models in the brms_families vignette.

.gitignore
```diff
@@ -9,3 +9,4 @@ tests/local/models_0.8.0.Rda
 tests/local/models_0.10.0.Rda
 tests/local/models_1.2.0.Rda
 tests/local/Rplots.pdf
+.ipynb*
```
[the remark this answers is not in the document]
yananlong: Sorry about that, I use jupyter lab for editing so I added that line to avoid adding the checkpoints files. Will fix this.

R/distributions.R
```diff
@@ -2052,6 +2052,107 @@ phurdle_lognormal <- function(q, mu, sigma, hu, lower.tail = TRUE,
...
+        1, -inc,
+        inc + do_call(
+            cdf,
+            c(list(q), pars, lower.tail = FALSE, log.p = TRUE)
+        )
+    ))
+    # take the limits of the distribution into account
+    out <- ifelse(q < lb, 0, out)
+    out <- ifelse(q > ub, -Inf, out)
+    if (lower.tail) {
```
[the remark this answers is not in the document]
yananlong: The convention is to use 2 spaces for indentation, correct?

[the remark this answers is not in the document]
paul-buerkner: Yes but what about the if statements afterwards?

paul-buerkner:
@yananlong Is this PR still something you are working on?

arnaudmgh:
Hello @yananlong and @paul-buerkner, 

I started modeling Lymphoma survival data using brms with the Weibull family. However, my posterior predictive survival curve decreases more or less like an exponential decay, the data curve can start very step and flattens at some point (i.e. high hazard followed by short hazard). I added a formula for the shape parameter `bf(time | cens(censor) ~ ..., shape ~ 1)` but it does not help fit much better.

I suspect I could fit the data better with a cure model. Is it something ready to use in brms? Or, should I pull this branch and try to install it in R?

Any guidance would be much appreciated. 

Best
Arnaud

paul-buerkner:
I will close this PR for now due to inactivity but please feel free to reopen it when you should find the time to work on it again.

yananlong:
Hi Paul and Arnaud,

Sorry I have shelved this project for now. I will probably return to it in November.

PR #1455, opened 2023-02-13
Error for response-specific formulas (`categorical()` family)

fweber144:
This PR adds an error message in case of response-specific formulas for the `categorical()` family (supported only by the augmented-data projection).

Reason: Users might be tempted to try something like this:
```r
set.seed(856824715)

data(VA, package = "MASS")
levels(VA$cell) <- paste0("lvl", levels(VA$cell))
offs_mat <- matrix(rnorm(nrow(VA) * nlevels(VA$cell)),
                   nrow = nrow(VA))
VA <- cbind(VA, offs = offs_mat)
VA[["offs.1"]] <- NULL

bfit <- brms::brm(
  formula = brms::brmsformula(
    cell ~ 1,
    mulvl2 ~ treat + age + Karn + offset(offs.2),
    mulvl3 ~ treat + age + Karn + offset(offs.3),
    mulvl4 ~ treat + age + Karn + offset(offs.4)
  ),
  data = VA,
  family = brms::categorical(),
  refresh = 0,
  seed = 255696126
)

library(projpred)
refm <- get_refmodel(bfit)
```
Previously, this code ran through with the resulting `refm$formula` being `cell ~ 1`. With this PR, the `get_refmodel()` call throws an error.

### DIFF
--- a/R/projpred.R
+++ b/R/projpred.R
@@ -94,6 +94,10 @@ get_refmodel.brmsfit <- function(object, newdata = NULL, resp = NULL,
   bterms <- brmsterms(formula)
   if (length(bterms$dpars) > 1L && !conv_cats_dpars(family)) {
     stop2("Projpred does not support distributional models.")
+  } else if (length(bterms$dpars) > 1L && conv_cats_dpars(family) &&
+             length(formula$pforms)) {
+    stop2("Projpred does not support response-specific formulas for the ",
+          family$family, " family.")
   }
   if (length(bterms$nlpars) > 0L) {
     stop2("Projpred does not support non-linear models.")


paul-buerkner:
Thanks! Why isn't it enough to check for `conv_cats_dpars` alone?

fweber144:
Because a `categorical()` family can have a single formula (which is the case which mustn't be aborted, because in that case, we can apply the augmented-data projection).

fweber144:
The problem is only a `categorical()` family with response-specific formulas.

paul-buerkner:
But why should categorical ever have only a single formula?

fweber144:
I was thinking of a model like the following:
```r

data(VA, package = "MASS")
levels(VA$cell) <- paste0("lvl", levels(VA$cell))

bfit <- brms::brm(
  formula = cell ~ treat + age + Karn,
  data = VA,
  family = brms::categorical(),
  file  = "bfit_single_fml",
  refresh = 0,
  seed = 255696126
)

```
Doesn't such a model make sense?

paul-buerkner:
Ah, I see. But this will still apply the same formula to all the mu parameters and in this sense then have multipled (copied) formulas. Would that work with regular projpred still? If yes, then I see why you additional check for `formula$pforms`.

fweber144:
Yes, that doesn't break code for the single-formula `categorical()` case. Specifically, I tried projpred for the single-formula model from https://github.com/paul-buerkner/brms/pull/1455#issuecomment-1427783098 and I also ran projpred's `R CMD check` with the brms version from this PR (and it succeeded).

The reason is that `formula$pforms` has length zero in the single-formula case. However, I'm not sure whether that's intended from brms's side or whether there is a better way to check for response-specific formulas. It's just the best solution I was able to come up with.

paul-buerkner:
Makes sense. Let me think about it for a second before I merge.

paul-buerkner:
Merged after minor edits.

fweber144:
Thanks!

PR #1456, opened 2023-02-13
Check projpred version for latent projection

fweber144:
I realized that in #1451, I could have checked the projpred version that is required for the latent projection directly in the code. Sorry, I have implemented this now (and also added corresponding documentation at argument `latent`; #1451 did so only in the `NEWS` entry).


paul-buerkner:
Good point. Brms has the `require_package` function for this purpose.

fweber144:
Thanks, I switched to `require_package()` (commit 75cedce6bcbd35b6d647747b4c7d7fca76c869a2). Since `require_package()` uses `<` instead of `<=`, I consistently switched to projpred version 2.4.0 as cutoff instead of 2.3.0 (on CRAN, there is no projpred version between 2.3.0 and 2.4.0).

paul-buerkner:
Perfect, thank you!

PR #1459, opened 2023-02-14
Fix #1457 (projpred `newdata` requiring more variables than necessary)

### LINKED ISSUE #1457: Unnecessary variables required in `newdata` for predictions in projpred
When performing predictions with new data in projpred, brms reference models require `newdata` to contain the response (and in fact *all* variables from the original dataset), which is not necessary after projecting onto a smaller model:
```r
library(brms)
ref_fit <- brm(
  formula = rate ~ conc + state,
  data = Puromycin,
  backend = "cmdstanr",
  seed = 2052109,
  refresh = 0
)

library(projpred)
soltrms <- "conc"
prj <- project(ref_fit, solution_terms = soltrms, ndraws = 40)
Puromycin_new <- as.data.frame(sapply(Puromycin[soltrms], quantile))
prj_linpred <- proj_linpred(prj, newdata = Puromycin_new)
```
with the last line throwing
```
Error: Response variables must be specified in 'newdata'.
Missing variables: 'rate'
```
(Tested with projpred 2.4.0 and brms's GitHub version, commit https://github.com/paul-buerkner/brms/commit/b31fcee8db3f4d04635f4ad25f0290367277a915.)

This also came up in [The Stan Forums](https://discourse.mc-stan.org/t/projpred-model-with-horseshoe-prior-does-not-converge/29935) and seems to be a long-standing bug, probably related to the `extract_model_data()` function defined in brms and passed to `projpred::init_refmodel()` (the  `extract_model_data` argument was introduced in projpred 2.0.0, so quite some time ago). I will try to fix this and then create a PR.

fweber144:
This fixes issue #1457. See the commit messages for details. If you don't want to increase the version number, you can revert that of course.

I have also added a `TODO` comment in [this line](https://github.com/fweber144/brms/blob/79d55b1861cd252d832913d286eb7ec77952661b/R/projpred.R#L268) because I experienced a slight inconsistency while testing this fix: If *weights* (i.e., from `resp_weights()`) are missing in `newdata` (and listed in `req_vars`), no error is triggered, but such an error is triggered if *numbers of trials* (i.e., from `resp_trials()`) are missing in `newdata` (and listed in `req_vars`). Already from a conceptual point of view, I think it would be desirable to also throw an error if weights are missing. Tell me if you need a reprex for this (but I think this would be a different issue and furthermore, this is not projpred-related).

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -2,8 +2,8 @@ Package: brms
 Encoding: UTF-8
 Type: Package
 Title: Bayesian Regression Models using 'Stan'
-Version: 2.18.7
-Date: 2023-01-17
+Version: 2.18.8
+Date: 2023-02-14
 Authors@R:
     c(person("Paul-Christian", "Bürkner", email = "paul.buerkner@gmail.com",
              role = c("aut", "cre")),
--- a/NEWS.md
+++ b/NEWS.md
@@ -18,6 +18,8 @@ via the new argument `ndraws_point_estimate`.
 * Fix a Stan syntax error in threaded models with `lasso` priors. (#1427)
 * Fix Stan compilation issues for some of the more special 
 link functions such as `cauchit` or `softplus`.
+* Fix a bug for predictions in **projpred**, previously requiring more variables
+in `newdata` than necessary. (#1457, #1459)
 
 
 # brms 2.18.0
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,7 +11,7 @@ in post-processing methods that require a compiled Stan model.
 * Extend control over the `point_estimate` feature in `prepare_predictions`
 via the new argument `ndraws_point_estimate`.
 * Add support for the latent projection available in **projpred** versions >=
-2.4.0.
+2.4.0. (#1451)
 
 ### Bug Fixes
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,8 +10,8 @@ via the `hurdle_cumulative` family thanks to Stephen Wild. (#1448)
 in post-processing methods that require a compiled Stan model.
 * Extend control over the `point_estimate` feature in `prepare_predictions`
 via the new argument `ndraws_point_estimate`.
-* Add support for the latent projection available in **projpred** versions >=
-2.4.0. (#1451)
+* Add support for the latent projection available in 
+**projpred** versions >= 2.4.0. (#1451)
 
 ### Bug Fixes
 
--- a/R/projpred.R
+++ b/R/projpred.R
@@ -237,16 +237,14 @@ get_refmodel.brmsfit <- function(object, newdata = NULL, resp = NULL,
     formula <- formula$forms[[resp]]
   }
   bterms <- brmsterms(formula)
+  y <- NULL
   if (extract_y) {
-    respform <- bterms$respform
     data <- current_data(
       object, newdata, resp = resp, check_response = TRUE,
       allow_new_levels = TRUE, req_vars = character()
     )
-    y <- unname(model.response(model.frame(respform, data,
-                                           na.action = na.pass)))
-  } else {
-    y <- NULL
+    y <- model.response(model.frame(bterms$respform, data, na.action = na.pass))
+    y <- unname(y)
   }
 
   # extract relevant auxiliary data (offsets and weights (or numbers of trials))
--- a/R/projpred.R
+++ b/R/projpred.R
@@ -267,10 +267,11 @@ get_refmodel.brmsfit <- function(object, newdata = NULL, resp = NULL,
   sdata <- do_call(standata, args)
 
   usc_resp <- usc(resp)
+  N <- sdata[[paste0("N", usc_resp)]]
   weights <- as.vector(sdata[[paste0("weights", usc_resp)]])
   trials <- as.vector(sdata[[paste0("trials", usc_resp)]])
   if (is_binary(formula)) {
-    trials <- rep(1, sdata[["N"]])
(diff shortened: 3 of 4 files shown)


paul-buerkner:
Thanks! Why should weights be required if the response is not? They are only relevant for log_lik, which requires the response as well. Posterior_predict for example does not make use of weights.

fweber144:
> Thanks! Why should weights be required if the response is not? They are only relevant for log_lik, which requires the response as well. Posterior_predict for example does not make use of weights.

I was also thinking about whether the requirement of weights could be tied to the requirement of the response, but projpred uses the weights in `proj_predict()` (the equivalent to `posterior_predict()`) in case of a binomial family (where these weights can be regarded as the numbers of trials, but projpred doesn't differentiate these).

Yesterday, this made me think about observation weights in `proj_predict()` in general, so I had a discussion with @avehtari because of this. I asked him whether it wouldn't it make sense for `proj_predict()` to take observation weights into account also for the Poisson and the Gaussian family, for example. (Although admittedly, in practice, I have never seen observation weights in families other than binomial.) The idea for implementing this is that a weighted dataset would need to be de-aggregated so that instead of a $S_{\text{prj}} \times N$ matrix (with $N$ denoting the number of weighted observations), `proj_predict()` returns an $S_{\text{prj}} \times \tilde{N}$ matrix, with $\tilde{N}$ denoting the number of de-aggregated observations. Things get a little trickier in case of non-integer observation weights. In that case, a solution could be to sample de-aggregated observations with probabilities proportional to the weights. @avehtari suggested stratified sampling for this, as implemented in the posterior package, for example. But even then, non-integer $\tilde{N}$ remains a problem that I'm not sure yet how to solve (simply round $\tilde{N}$ to the next integer?).

I think incorporating observation weights in `proj_predict()` and `posterior_predict()` makes sense, especially if you think of `bayesplot::ppd_dens_overlay()`. The kernel density estimate across observations that is used there should be influenced by observation weights, I think.

But honestly, I'm not very keen to implementing this in projpred soon, as this is a very special case and probably hard to implement in a sensible way in terms of the UI (for example, how to allow users to identify which columns of the $S_{\text{prj}} \times \tilde{N}$ matrix belong to which columns of the $S_{\text{prj}} \times N$ matrix?). For now, I might throw an error or a warning in projpred when using `proj_predict()` in case of observation weights (apart from the binomial family because things seem to be correct there). But as mentioned above, `proj_predict()` in the binomial case prevents us from tying the requirement of weights to the requirement of the response.

fweber144:
I have to add that for `bayesplot::ppd_dens_overlay()`, the original response values (`y`) would also have to be de-aggregated in the same way as for the predictions.

fweber144:
I guess that `rstantools::posterior_predict()` and corresponding rstanarm methods would also need to be changed if we really want to go in this direction.

paul-buerkner:
Well actually, I am not sure I want to change this stuff right now and I will have to think about whether this makes any sense at all. I will merge the PR and we can revisit the weight stuff later.

fweber144:
Thanks! Revisiting the weights stuff later is ok for me.

stainesa:
There is one important use case for weights - surveys. This is pretty much what I do all the time, and almost every analysis is intended either to make inferences about a wider population, or to do predictions, based on a smaller survey, for a larger, simpler dataset from that population.

If I've understood this discussion, I could simply put extra rows into my data, specifically weight<i> copies of each original row for every row observed, where weight<i> is the weight of the original row. The error in rounding to integer weights in my world would be ignorable.

Is this reasonable? It does mess up the sample size horrendously, and might give a very false idea of precision, so I would not embark on it lightly.

fweber144:
What is your response family? And do you want to use `proj_linpred()`, `proj_predict()`, or `predict.refmodel()`?

stainesa:
At the moment most of my own work is logistic regression, with the binomial family, and I pretty much only use the linpred outputs. I study relatively uncommon events (2% to 10% roughly, and the posterior predict functions seems to use a cutoff of 0.5 for decision, which is reasonable, but is wrong for my situation. I'm predicting population level prevalences of these rare events, in my case between hospitals over time. The current work is focused on surveillance. Individual predicted risks are not of interest, and in any case arewildly imprecise.

What I see when I feed one of my stan_glm model fits to get_refmodel() is this :-
` ref_fit_PU <- get_refmodel(fit_PU)
Error in get_refmodel.stanreg(fit_PU) : 
  In case of the binomial family, projpred cannot handle observation weights (apart from the numbers of trials).`

The model itself is fitted as :-
`fit <- stan_glm(make_formula(df.train),
                 data = df.train,
                 family = binomial(link = "logit"),
                 weights = Weights,
                 prior = t_prior, prior_intercept = t_prior ) `

Weights is a vector of numeric sampling weights, reflecting the deliberate and substantial oversampling of one category of patient on the training data.

fweber144:
> the posterior predict functions seems to use a cutoff of 0.5 for decision

`posterior_predict()` draws from the posterior predictive distribution and `proj_predict()` does the same, just for a predictive distribution based on a projected posterior. So there should be no cutoff involved when using these two functions. A cutoff (of indeed 0.5) is involved only when using the `"acc"` (= `"pctcorr"`) performance statistic in `plot.vsel()`, `summary.vsel()`, or `suggest_size()`.

For the binomial family you are using, there should be no extra work necessary to obtain the appropriate output from functions such as `proj_linpred()` and `proj_predict()`. What I don't fully understand is why `get_refmodel.stanreg()` does not work in your case. What exactly is returned by `make_formula(df.train)`?

PR #1478, opened 2023-04-05
Add compress parameter to brm and write_brmsfit

mawilson1234:
RDS objects containing brmsfit objects can be quite large depending on the model. Currently, brms uses saveRDS's default compression, gzip, to save models to disk. However, saveRDS also supports xz and bzip2. I found that simply using xz instead of gzip can result in a dramatic reduction in needed disk space, tested by reading in a saved model and just reexporting it with `saveRDS(object, file, compress = "xz")`. While the write process is noticeably slower, I tested with two models and observed these reductions in file size:

- 141 MB -> 73 MB
- 395 MB -> 205 MB

This reduced the model sizes to a little over half of the original size, which seems like a pretty good savings.

This pull request adds a `compress` parameter to `brm` and `write_brmsfit` that gets passed along to `saveRDS` to allow the user to specify the compression algorithm to use. The default value is the current default for `saveRDS`, which is `gzip`.

### DIFF
--- a/R/brm.R
+++ b/R/brm.R
@@ -208,8 +208,8 @@
 #'   files won't be overwritten, you have to manually remove the file in order
 #'   to refit and save the model under an existing file name. The file name
 #'   is stored in the \code{brmsfit} object for later usage.
-#' @param compress A character string, one of \code{"gzip"}, \code{"xz"}, or
-#'   \code{"bzip2"} (default \code{"gzip"}). If the \code{file} parameter is
+#' @param compress A character string, one of the compression algorithms supported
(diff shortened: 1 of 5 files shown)


paul-buerkner:
Thanks! Would you mind adding this to `brm_multiple` too?

mawilson1234:
Done! I haven't actually used `brm_multiple` myself, but I think I got it right.

paul-buerkner:
Thanks. I renamed the argument to `file_compress` to be more clear about its use. Will merge now.

PR #1492, opened 2023-05-04
Support global shrinkage priors

paul-buerkner:
This PR adds support for the application of the horseshoe and R2D2 prior across all terms of an additive predictor term (i.e., for all terms of a single formula). Here is an example:

```R
set.seed(2563)
dat <- epilepsy
dat$x1 <- sample(1:4, nrow(dat), TRUE)
dat$x2 <- sample(1:10, nrow(dat), TRUE)
dat$x3 <- sample(1:10, nrow(dat), TRUE)

bprior <- prior(R2D2(), class = b) +
  prior(R2D2(main = FALSE), class = sd) +
  prior(R2D2(main = FALSE), class = sds) +
  prior(R2D2(main = FALSE), class = sdgp) +
  prior(R2D2(main = FALSE), class = ar) +
  prior(R2D2(main = FALSE), class = ma)

bform <- bf(count ~ Trt * Base + Age + mo(x1) + (1|patient) + gp(x2) + s(x3) +
              arma(p = 2, q = 2, gr = patient))

fit <- brm(bform, data = dat, prior = bprior, chains = 1)
```


paul-buerkner (commented):
Quickly going over all changes again and it seems to look good.

paul-buerkner:
Checks are looking good (the one that fails fails for other reasons). Merging now.

PR #1521, opened 2023-07-04
Allow data validation to be skipped to improve `posterior_predict` performance with `newdata`

bmfazio:
When `posterior_predict` is called with `newdata`, `validate_newdata` will be called and this represents a non-trivial amount of the total execution time. Advanced users may want to skip this step when they know their data already has the proper structure.

As there are different kinds of validation functions in the package, the proposed argument name is `skip_validate_data`.

```r
library(brms)
fit <- brm(time | cens(censored) ~ age + sex + (1 + age || patient),
           data = kidney, family = "exponential", init = "0",
           backend = "cmdstanr")

microbenchmark::microbenchmark(
  posterior_predict(fit, newdata = fit$data),
  posterior_predict(fit, newdata = fit$data, skip_validate_data = TRUE)
)
# Unit: milliseconds
#                                          expr      min       lq     mean   median       uq      max neval
#                             posterior_predict 228.1443 252.5301 274.1673 266.1105 277.0043 570.7657   100
#  posterior_predict(skip_validate_data = TRUE) 130.8804 149.5136 158.3647 155.1540 163.1465 414.9703   100
```

The change is made for both `current_data` and `current_data2` and should work without issue when passed via `...` from `standata` and `pp_check` calls.

Two calls of `current_data` in `get_refmodel` had `...` added so the argument can also be used there.

There is one `current_data2` call left without passing `...` as it's inside `validate_newdata` so it'd be pointless there.

### DIFF
--- a/R/data-helpers.R
+++ b/R/data-helpers.R
@@ -618,11 +618,12 @@ validate_newdata2 <- function(newdata2, object, ...) {
 }
 
 # extract the current data
-current_data <- function(object, newdata = NULL, skip_validate_data = FALSE, ...) {
+current_data <- function(object, newdata = NULL, skip_validation = FALSE, ...) {
   stopifnot(is.brmsfit(object))
+  skip_validation <- as_one_logical(skip_validation)
   if (is.null(newdata)) {
     data <- object$data
-  } else if(skip_validate_data) {
+  } else if (skip_validation) {
     data <- newdata
   } else {
     data <- validate_newdata(newdata, object = object, ...)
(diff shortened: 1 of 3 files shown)


paul-buerkner:
Thanks! Is there a reason to not call it skip_validate? I liked that name in our discussion today.

bmfazio:
Named and style tweaked as requested.

Also future reminder that the argument may be reused or renamed depending on how we want it to interact with other types of non-data validation steps.

paul-buerkner:
great. Merging now.

Issue #1222, opened 2021-09-01
projpred: `Suggests:` instead of `Imports:`?


fweber144:
Could projpred be moved from `Imports:` to `Suggests:`? Currently,
```r
library(brms)
library(projpred)
detach("package:projpred", unload = TRUE)
```
produces the warning
```
‘projpred’ namespace cannot be unloaded:
  namespace ‘projpred’ is imported by ‘brms’ so cannot be unloaded
```
which does not occur for emmeans, for example (which is in `Suggests:`):
```r
library(brms)
library(emmeans)
detach("package:emmeans", unload = TRUE)
```
If you want, I can create a PR for that.

paul-buerkner:
I think emmeans uses some magic so that methods can be registered to the
right generics even when a package is in suggests and then loaded when
needed. I would also prefer projpred under suggests. so if you can check
what emmeans does and make sure that works in projpred as well that would
be really nice!

Frank Weber ***@***.***> schrieb am Mi., 1. Sept. 2021, 09:17:

fweber144:
I experimented a bit with it. Simply moving projpred from `Imports:` to `Suggests:` does indeed not remove that warning. Unfortunately, using the emmeans way in projpred would require some work for which I currently don't have the time. I'll create an issue on projpred's issue tracker so this can be considered in the future. But for now, I think I would simply create a PR here for brms which moves projpred from `Imports:` to `Suggests:`. Are you ok with that?

paul-buerkner:
I doubt just moving to suggests will work. you can try it out and then run
cmd check. I assume it will complain.

Frank Weber ***@***.***> schrieb am Mi., 1. Sept. 2021, 10:39:

fweber144:
Yes, you're right. I forgot the imports of projpred functions in the `NAMESPACE` file. I'll see how to deal with it.

fweber144:
There seem to be several ways around that, see e.g. [here](https://stackoverflow.com/questions/61482561/whats-the-preferred-means-for-defining-an-s3-method-in-an-r-package-without-int) and [here](https://community.rstudio.com/t/write-s3-methods-for-generics-from-other-packages-without-importing-that-package/8884). The downside would be a breaking change because users would have to load or attach projpred themselves if they want to use it. But on the other hand, I think that makes sense. Not only from a conceptual perspective, but also because currently, users might not even notice to be using projpred. In terms of citations for projpred, that would be disadvantageous.

[...]

PR #1525, opened 2023-07-06
Modernize registration of S3 method for projpred generic

fweber144:
I'm not sure if you are willing to increase the required R version to >= 3.6.0 (until now: >= 3.5.0), but if you are, then this PR provides a more modern way to register the S3 method `get_refmodel.brmsfit()` for the `projpred::get_refmodel()` generic.

This was already discussed in #1222, but at that time (more precisely, at https://github.com/paul-buerkner/brms/issues/1222#issuecomment-910276807), I didn't realize that you don't need to call `library(projpred)` before documenting via `devtools::document()` if you can live with having `get_refmodel.brmsfit(` in `man/get_refmodel.brmsfit.Rd` (as is the case already now) instead of `\method{get_refmodel}{brmsfit}(` (the latter would render the typical `## S3 method for class 'brmsfit'` header in the "Usage" section of HTML and PDF documentation, but as I said, we don't have that header at the moment either). In other words, in #1222, I didn't realize that calling `library(projpred)` before documenting only provides a minor cosmetic improvement compared to the current state.

Another reason why I would switch to this more modern S3 registration is that in https://github.com/r-lib/vctrs/blob/master/R/register-s3.R, the code for `s3_register()` has changed, so it should probably be updated if you don't want to adopt the more modern scheme. However, when copying the new `vctrs::s3_register()` code, you have to keep in mind that this would require at least one new helper function (`vctrs:::.rlang_s3_register_compat()`), which increases the amount of code that has to be copied to and maintained in brms. This also shows that keeping the current S3 registration up to date is kind of tedious, because you regularly have to check `vctrs::s3_register()` for updates.

I have tested this PR locally with the following code (where I restarted the R session between each section):
```r

# Attach ------------------------------------------------------------------

library(brms)
ref_fit <- brm(
  formula = rate ~ conc + state,
  data = Puromycin,
  backend = "cmdstanr",
  seed = 2052109,
  refresh = 0
)

library(projpred)
prj <- project(ref_fit, solution_terms = "conc", ndraws = 40)
prjmat <- as.matrix(prj)

# Load --------------------------------------------------------------------

ref_fit <- brms::brm(
  formula = rate ~ conc + state,
  data = Puromycin,
  backend = "cmdstanr",
  seed = 2052109,
  refresh = 0
)

prj <- projpred::project(ref_fit, solution_terms = "conc", ndraws = 40)
prjmat <- as.matrix(prj)

# Attach brms, load projpred ----------------------------------------------

library(brms)
ref_fit <- brm(
  formula = rate ~ conc + state,
  data = Puromycin,
  backend = "cmdstanr",
  seed = 2052109,
  refresh = 0
)

prj <- projpred::project(ref_fit, solution_terms = "conc", ndraws = 40)
prjmat <- as.matrix(prj)

# Load brms, attach projpred ----------------------------------------------

ref_fit <- brms::brm(
  formula = rate ~ conc + state,
  data = Puromycin,
  backend = "cmdstanr",
  seed = 2052109,
  refresh = 0
)

library(projpred)
prj <- project(ref_fit, solution_terms = "conc", ndraws = 40)
prjmat <- as.matrix(prj)
```
as well as with a couple of hands-on examples as well as within projpred's unit tests (and I have also run `R CMD check` for brms, of course).


paul-buerkner:
Thank you! Yes, I would prefer not to load projpred before documenting. I will check out the PR now.

paul-buerkner:
Looks good. Merging now.

PR #1548, opened 2023-10-10
Update translation of silent argument to fix cmdstanr breakage

andrjohns:
Updates the `silent` -> `show_messages`,`show_exceptions` translation for `cmdstanr` to the following mapping:

- `silent=0` suppress nothing (`show_messages=TRUE, show_exceptions=TRUE`)
- `silent=1` suppress only exceptions (`show_messages=TRUE, show_exceptions=FALSE`)
- `silent=2` suppress messages and exceptions (`show_messages=FALSE, show_exceptions = FALSE`)

### DIFF
--- a/R/backends.R
+++ b/R/backends.R
@@ -278,7 +278,8 @@ fit_model <- function(model, backend, ...) {
       iter_warmup = warmup,
       chains, thin,
       parallel_chains = cores,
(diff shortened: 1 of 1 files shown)


paul-buerkner:
thank you!

jgabry:
Thanks!

PR #1559, opened 2023-11-09
Special casing for lp_approx__

bmfazio:
- Replace lp_approx__ with log_g__
- Assign then return stanfit object instead of implicit return


paul-buerkner:
Thank you!

PR #1561, opened 2023-11-10
Add khat summary to cmdstanr-based variational stanfit

bmfazio:


paul-buerkner:
Is there a way without calling rstan summary? This is not really something I want to call because it can be very slow for large models.

paul-buerkner:
No worries it not. Then we just leave this issue unaddressed.

bmfazio:
Sadly not. If one waits for the rstan code to run as usual, it will add the summaries incompletely. So if pre-adding them is not an option, I can only think of some hacky setup that would detect when the local environment gets updated and adds the missing variable in response to the event.

paul-buerkner:
Okay. Let's leave it unaddressed then. Thank you for your efforts!

PR #1566, opened 2023-11-14
Export psis function

sims1253:
I wanted to do this since forever and finally got around to it. I am not sure if there is a more elegant solution to exporting without a version without the dot without change all the internal calls to `.psis`. 
Let me know if anything needs change.


paul-buerkner:
Looks but, but why have both .psis and psis.brmsfit now? Can you check where .psis is applied and then just change it to psis accordingly. Or is there something that's blocking that change?

sims1253:
You use the prepented dot to `do_call` the criterion functions in https://github.com/paul-buerkner/brms/blob/fe7023f59182b90953c5e5cbe8dbfb13f5ae329b/R/loo.R#L247

So you either have to rename all criterion functions (potentially also exporting them), handle the extra case for psis not having a prepended dot any more or keep `.psis` as the wrapper.

paul-buerkner:
ah makes sense. So lets just leave it like this then. I will merge this soon.

paul-buerkner:
Merging now.

PR #1571, opened 2023-12-11
Add deprecation info to as.mcmc doc

avehtari:
Based on Stan Discourse, deprecation info was missing from as.mcmc doc https://discourse.mc-stan.org/t/as-mcmc-function-deprecated/33507


paul-buerkner:
Thanks!

PR #1578, opened 2024-01-05
Add more pp_check types in aps_types

avehtari:
Add more pp_check types in aps_types, that by default should use all draws


paul-buerkner:
Thanks!

Issue #1617, opened 2024-03-11
abstract function for reading cmdstanr csv output into a fit object for brms and export it


venpopov:
Currently when the cmdstanr backend is used, after the sampling is done, the following lines of .fit_model_cmdstanr() read in the output files, and do necessary repairs and transformation to make an object to put in the fit slot of the brmsfit object:

```r
.fit_model_cmdstanr(....
  ## .... previous lines
  # not all metadata is not stored by read_csv_as_stanfit
  metadata <- cmdstanr::read_cmdstan_csv(
    out$output_files(), variables = "", sampler_diagnostics = ""
  )
  # ensure that only relevant variables are read from CSV
  variables <- repair_variable_names(metadata$metadata$variables)
  variables <- unique(sub("\\[.+", "", variables))
  variables <- setdiff(variables, exclude)
  # temp fix for cmdstanr not recognizing the variable names it produces  #1473
  variables <- ifelse(variables == "lp_approx__", "log_g__", variables)
  # transform into stanfit object for consistent output structure
  out <- read_csv_as_stanfit(out$output_files(), variables = variables)
  out <- repair_stanfit(out)
  # allow updating the model without recompilation
  attributes(out)$CmdStanModel <- model
  attributes(out)$metadata <- metadata
  out
}
```

Would you be willing to abstract these lines into a separate function and export it so that it can be used externally? It could look something like this (naming is just an example - currently misleading since the output is just the fit slot of a brmsfit object):

```r
#' @export
cdmstan_to_brmsfit <- function(files, exclude = "") {
  # not all metadata is stored by read_csv_as_stanfit
  metadata <- cmdstanr::read_cmdstan_csv(
    files, variables = "", sampler_diagnostics = ""
  )
  # ensure that only relevant variables are read from CSV
  variables <- repair_variable_names(metadata$metadata$variables)
  variables <- unique(sub("\\[.+", "", variables))
  variables <- setdiff(variables, exclude)
  # temp fix for cmdstanr not recognizing the variable names it produces  #1473
  variables <- ifelse(variables == "lp_approx__", "log_g__", variables)
  # transform into stanfit object for consistent output structure
  out <- read_csv_as_stanfit(files, variables = variables)
  out <- repair_stanfit(out)
  # allow updating the model without recompilation
  attributes(out)$CmdStanModel <- model
  attributes(out)$metadata <- metadata
  out
 }
 ```

and then the existing .fit_model_cmdstanr will be

```r
# previous lines here
[...]

PR #1618, opened 2024-03-12
Abstract function for reading cmdstanr csv output into a fit object for brms and export it

venpopov:
This PR implements the suggestion in #1617

A new exported function read_csv_as_brms_stanfit() can take cmdstan csv files and a cmdstanr compiled model (optional) and produce a stanfit object formated consistently with the $fit slot of brmsfit objects. The functionality was previously internally part of .fit_model_cmdstanr().

In this version the code is almost 1:1 as what it was before (I added a require_package('cmdstanr') only) and could be merged as is. However, I have some suggestions for improvement, which I will leave as comments in the code and you can decide whether to keep it as is, or if I should make those changes.


R/backends.R
venpopov, line 1: seems I can't leave comments on unchanged lines. 

Lines 757-759 of read_csv_as_stanfit() have:

```r
  fnames_oi <- colnames(samples)

  colnames(samples) <- gsub("\\[", ".", colnames(samples))
  colnames(samples) <- gsub("\\]", "", colnames(samples))
  colnames(samples) <- gsub("\\,", ".", colnames(samples))
```

which I guess was done for consistency with the rstan::read_stan_csv function, but this introduces the problem that needs to then be fixed by repair_stanfit():

```r
    # stanfit may have renamed dimension suffixes (#1218)
    if (length(x@sim$samples[[i]]) == length(x@sim$fnames_oi)) {
      names(x@sim$samples[[i]]) <- x@sim$fnames_oi
    }
 ```

so maybe we can just delete those lines?

File: R/backends.R [hunk 1]
```diff
@@ -957,3 +944,56 @@ read_csv_as_stanfit <- function(files, variables = NULL,
...
+#' # feed the Stan model back into brms
+#' fit <- brm(count ~ Trt, data = epilepsy, empty = TRUE, backend = 'cmdstanr')
+#' fit$fit <- read_csv_as_brms_stanfit(stanfit$output_files(), model = mod)
+#' fit <- rename_pars(fit)
+#' summary(fit)
+#' }
+read_csv_as_brms_stanfit <- function(files, model, exclude = "") {
+  # not all metadata is stored by read_csv_as_stanfit
+  require_package("cmdstanr")
+  metadata <- cmdstanr::read_cmdstan_csv(
```
venpopov, line 981: reading the metadata manually like this introduces unnecessary overhead. this information is available in only a slightly different format from `out$metadata()`.  This information is used to select the variables, which are then renamed and combined with exclude, before being passed to read_csv_as_stanfit. Which are then renamed in the same way in[ that function as well.](https://github.com/paul-buerkner/brms/blob/7b0adcd75e68bef8546c439880ba606bed8cb29b/R/backends.R#L659), so this is unnecessary.

I think the whole thing can be simplified into a single function with less duplication of code and less unnecessary reading of files. I may be missing why this is done this way, so I'll make another version in a separate branch and you can decide which one to keep

venpopov: Here's the alternative version: https://github.com/paul-buerkner/brms/pull/1619

Issue #1218, opened 2021-08-20
displaying models fitted pre 2.16 in brms 2.16 - variables missing in the draws object


fusaroli:
After upgrading to brms 2.16, when trying to open saved models fitted with previous versions (e.g. 2.15), either by calling model or summary(model), i get "Error: The following variables are missing in the draws object: " followed by a long list of model parameters (e.g. Cor_1_1[1,1]).

The models look otherwise fine in the environment pane in rstudio.

Is there a basic conversion to be done to the models to comply with the new "draws" hegemony, which I have missed?

paul-buerkner:
This is not great as I assume a lot of people will have this problem. Can you send me such a model than complains this way?

fusaroli:
sorry, was away from the computer. I seem to only have relatively complex models at hand, but here it's a gaussian one, so the issue shouldn't just be e.g. in mixture models: 
- https://www.dropbox.com/s/iawss4unif9094b/ChildMLUPred0_m.rds?dl=0

paul-buerkner:
Thank you! The problem is actually caused by (what is think might by a bug in) `rstan::read_stan_csv` when reading in CSVs created by cmdstanr. As such this will only happen for models fitted with the cmdstanr backend. The reason this became apparent only now is the use of the posterior package which is more strict about variable names. This is automatically fixed for newly fitted models with brms 2.16.0 and I just pushed a fix to github that does the same for old models.

@jgabry, @bgoodri, @hsbadr: It seems as if `rstan::read_stan_csv` accidently renames parameter names in `stanfit@sim$samples` causing, for example, dimension brackets etc., to be replaced by a `.` as a result of an implicit call to `make.names` I assume. 

I am now working around this issue in brms via the following helper function:

```R
# repairs parameter names of stanfit objects as rstan::read_stan_csv 
# seems to accidentally rename dimension suffixes
repair_stanfit_names <- function(x) {
  stopifnot(is.stanfit(x))
  for (i in seq_along(x@sim$samples)) {
    names(x@sim$samples[[i]]) <- x@sim$fnames_oi
  }
  x
}
```

fusaroli:
awesome, thanks. this now works on a bunch of different models, but there's still at least one issue.
If I try to load zero inflated models, I get 

"Error: Duplicate variable names are not allowed in draws objects.
The following variable names are duplicates:
{'prior_sd_ChildID'}"

This happens for a range of zi models (gaussian, poisson and lognormal),

[...]

PR #1619, opened 2024-03-12
Read csv as stanfit v2 (alternative version)

venpopov:
An alternative version of #1618, as I noted in my comment [here](https://github.com/paul-buerkner/brms/pull/1618#discussion_r1520674645)

Rather than having a new function, here I slightly modify the existing read_csv_as_stanfit() function and export it instead. I find this option a bit cleaner, slightly faster and likely easier to maintain.

- combines read_csv_as_stanfit  and read_csv_as_brms_stanfit into one function
- eliminates redundant reading metadata from file
- remove redundant renaming
- remove renaming samples columns that otherwise needs to be fixed (https://github.com/paul-buerkner/brms/issues/1218)

The main things to note
- previously there was a separate [read_cmdstan_csv() call that only reads in the metadata](https://github.com/paul-buerkner/brms/blob/15c0754ba6e7d1032cf2bee40ad83e990ac09d72/R/backends.R#L294). This is unnecessary, since all of that information is already contained in the call to read_cmdstan_csv() in [read_csv_as_stanfit()](https://github.com/paul-buerkner/brms/blob/15c0754ba6e7d1032cf2bee40ad83e990ac09d72/R/backends.R#L660). This speeds up the import by ~10-20% in my testing.
- I removed these lines:

```r
  colnames(samples) <- gsub("\\[", ".", colnames(samples))
  colnames(samples) <- gsub("\\]", "", colnames(samples))
  colnames(samples) <- gsub("\\,", ".", colnames(samples))
```

as they introduce the issue that then needs to be solved by repair_stanfit()

Feel free to go with either #1618 or this.


paul-buerkner:
Looks good, thank you! I made some minor edits and will now merge the PR.

venpopov:
Cool, thank you, this will be super useful for me! And thanks for adding me as a contributor.

PR #1624, opened 2024-03-18
update doc and loo recommendations

### LINKED ISSUE #1606: Update readme and messages to take into account latest loo release
- Update README to show the new loo output with 3 Pareto-k categories
- If the inference is run with less than about 2200 stored draws, k_threshold<0.7. If khats exceed this threshold, the diagnostic message could be modified to recommend in such cases first to run more iterations to get more than 2200 posterior draws. With more than 2200 stored draws the diagnostic message recommending moment_matching or reloo can stay the same

avehtari:
Small updates to doc. I decided to keep the k_threshold argument default unchanged (0.7), as here it specifically controls for which observations to run moment_matching or reloo, but loo recommendations has a new recommendation if the sample size specific threshold is <0.7 and there are Pareto k's exceeding that threshold. Unfortunately, `ps_khat_threshold()` is not yet exported in `loo` or `posterior` so had to copy that.

EDIT:fixes #1606

### DIFF
--- a/R/loo.R
+++ b/R/loo.R
@@ -679,18 +679,17 @@ recommend_loo_options <- function(loo, k_threshold = 0.7, moment_match = FALSE,
     model_name <- ""
   }
   ndraws <- dim(loo)[1] %||% Inf
-  n <- length(loo::pareto_k_ids(loo, threshold = k_threshold))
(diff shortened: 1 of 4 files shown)


paul-buerkner:
Thanks! Merging now.

PR #1632, opened 2024-03-23
change default prior for von_mises mu tan_half link

### LINKED ISSUE #1631: default prior for mu in von_mises with tan_half link
The current default prior for mu in the von_mises distribution is the same as for every other distribution: 

`student_t(3, 0, 2.5)`

By default the von_mises() family uses a tan_half link for mu. This implies that the prior on the constrained space for mu is really weird and quite informative:

```
mu_unconstrained <- rstudent_t(1000000, 3, 0, 2.5)
mu <- 2 * atan(mu_unconstrained)
hist(mu, breaks=60, freq=F)
```

image

I woudl expect that the default prior should either be uniform on the constrained space, or ccentered around 0, rather than a highly informative bimodal prior.

If we want to have a uniform prior over -pi to pi on the constrained scale with the tan_half link, this is quite easy: we just need a student_t(1, 0, 1) prior on the unconstrained space. This is because if we assume a uniform prior over -pi to pi, and apply the inv_tan_half transformation, the jacobian adjustment gives us exactly the likelihood of a student T with 1 degree of freedom:

$$ f(x) = \frac{1}{\\pi (1+x^2)}$$

or simulated:

```
mu_unconstrained <- rstudent_t(1000000, 1, 0, 1)
mu <- 2 * atan(mu_unconstrained)
hist(mu, breaks=60, freq=F)
```

image

Since you mentioned [here](https://github.com/paul-buerkner/brms/issues/1614#issuecomment-2006051851) that now family-specific priors are possible, I think this would be worth changing.

venpopov:
Closes #1631 by placing a student_t(1, 0, 1) prior on mu in the von_mises family when a tan_half link is used. This corresponds to a uniform prior over the constrained scale from -pi to pi.

The family-specific prior feature you implemented is really neat and made this super easy!

For what is worth, in my testing the prior change only had a meaningfull effects for very low numbers of observations and low precision (kappa) - otherwise results were pretty much identical.

This only affects the prior for the tan-half link - should the identity link prior be left as is? the current prior has a slight bump at 0, but its otherwise unimodal and weakly informative.

### DIFF
--- a/R/family-lists.R
+++ b/R/family-lists.R
@@ -350,7 +350,14 @@
     ybounds = c(-pi, pi), closed = c(TRUE, TRUE),
     ad = c("weights", "subset", "cens", "trunc", "mi", "index"),
     include = c("fun_tan_half.stan", "fun_von_mises.stan"),
-    normalized = ""
+    normalized = "",
+    # experimental use of default priors stored in families #1614
+    prior = function(dpar, link = "identity", ...) {
+      if (dpar == "mu" && link == "tan_half") {
+        return("student_t(1, 0, 1)")
+      }
+      NULL
+    }
   )
 }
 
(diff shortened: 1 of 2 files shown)


paul-buerkner:
Thanks! Merging now.

PR #1639, opened 2024-04-01
Fix p limits for some q* distribution functions; fix qgen_extreme_value

### LINKED ISSUE #1637: "Error: 'p' must contain probabilities in (0,1)" in qstudent_t(1, df = 3), qskew_normal and a few others
I was trying to visualize some distributions defined in brms, for example the student_t, via `ggdist` and the `distributional` package. For some distributions it works fine:

```r
library(ggplot2)
library(ggdist)
library(distributional)
library(brms)
dist1 <- dist_wrap('shifted_lnorm', package = 'brms')
ggplot() +
  stat_halfeye(aes(xdist = dist1))
```

image

but for others such as student_t, skew_normal, there is an error:

``` r
dist1 <- dist_wrap('student_t', package = 'brms', df = 3)
ggplot() +
  stat_halfeye(aes(xdist = dist1))
#> Error in `layer_slabinterval()`:
#> ! Problem while computing aesthetics.
#> ℹ Error occurred in the 1st layer.
#> Caused by error:
#> ! 'p' must contain probabilities in (0,1)
```

<sup>Created on 2024-03-29 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

This is because for some distributions (but not all), the code [here](https://github.com/paul-buerkner/brms/blob/c9146b5c1199b7df6cb4c6512692c44dd0b69da2/R/distributions.R#L2555) gives an error if p=0 or p=1 in a quantile function. Is there a particular reason to reject these values rather than returning +- Inf, as for example stats::qt(p=1) does?

venpopov:
Closes #1637 

This allows quantiles of 0 and 1 to be used for qstudent_t, qskew_normal, qfrechet, qgen_extreme_value and qasym_alplace.

Check if you like it the way I did. In addition, rather than throwing an error when p <0 or >1, the functions produce NaN like all other q* type distribution functions. It probably rarely matters, but I think the consistency is nice. The check itself is still necessary, because some distributions like qfrechet would produce non-meaningful but real values for p > 1.

I added a couple of new tests to check the indended results with p at or beyond the limits

## Incorrect output of qgen_extreme_value

While working on the above, I noticed that qgen_extreme_value produces incorrect results when xi != 0. There was no test for qgen_extreme_value, so I initially added the following test (similar to what was there for qfrechet):

```r
  n <- 10
  x <- rgamma(n, 10, 3)
  mu <- rnorm(n)
  res <- pgen_extreme_value(x, mu = mu, sigma = 1:n, xi = 3)
  q <- qgen_extreme_value(res, mu = mu, sigma = 1:n, xi = 3)
  expect_equal(x, q)
#> Error: `x` not equal to `q`.
#> 10/10 mismatches (average diff: 2.06)
#> [1] 3.385 -  0.296 == 3.090
#> [2] 2.194 - -0.797 == 2.990
#> [3] 2.563 -  2.489 == 0.074
#> [4] 2.340 -  1.075 == 1.265
#> [5] 3.846 - -0.274 == 4.119
#> [6] 0.751 -  0.243 == 0.508
#> [7] 2.213 -  1.127 == 1.086
#> [8] 4.624 -  2.251 == 2.373
#> [9] 4.628 -  2.405 == 2.223
#> ...
```

This [line](https://github.com/venpopov/brms-modified/blob/c9146b5c1199b7df6cb4c6512692c44dd0b69da2/R/distributions.R#L1031) was responsible, so I fixed it and now the new test passes. Let me know if you prefer to have this fix in a separate PR.


paul-buerkner:
Thank you! Looks good! Merging now.

PR #1653, opened 2024-05-13
Implement the brmsframe layer

paul-buerkner:
This PR implements `brmsframe` objects, which store a combination of formula and data information, thus reducing the need for repeated evaluation of several data-processing functions across the brms codebase. Although this PR changes a lot of code, there should not be any user-facing changes.

TODOs:
- [x] update NEWS
- [x] run all local tests
- [ ] review the changes on github


paul-buerkner (commented):
After going through the PR on github again, I think it looks good. Will merge now.

PR #1661, opened 2024-05-24
remove importMethodsFrom rstan summary

venpopov:
This PR removes a single @importMethodsFrom rstan summary statement. This import seems to be unused, but it is causing the loading time of `brms` to be ~ 2.5 seconds (on my machine), because loading rstan is slow. Removing this import speeds up the loading time of `brms` to ~ 0.6 seconds:

Before:

```{r}
start <- Sys.time()
library(brms)
Sys.time() - start
#> Time difference of 2.574536 secs
```

After:

```{r}
start <- Sys.time()
library(brms)
Sys.time() - start
#> Time difference of 0.616188 secs
```

All tests and check() pass and I couldn't find any place it's used, so it seems to be leftover code


paul-buerkner:
Thanks! Merging now!

Issue #1633, opened 2024-03-23
use von_mises_lpdf after next stan update


venpopov:
Currently the von_mises() family is not vectorized if kappa is predicted, because the von_mises_lpdf is numerically unstable for large kappas, and you use the normal approximation for kappa > 100, which requires and if else check.

I have submitted a [PR in stan math library ](https://github.com/stan-dev/math/pull/3036) that improves the stability of the von_mises_lpdf and which will make the von_mises_lpdf2 unneccessary, as I describe in the [issue here](https://github.com/stan-dev/math/issues/3035).

Once that is accepted and released, you could  simplify the von mises family, which my tests show it speeds it up, and also avoids scale=0 warning when 1/sqrt(kappa) passed to normal_lpdf is 0. Some tests described [here at the end](https://github.com/stan-dev/math/issues/3035). 

It's probably going to be a while until the next release and also that will work just for the `cmdstar` backend until rstan catches up, but I'm nothing it here for when the time is right.

of2:
+1 boosting this - thanks for fixing this issue in Stan @venpopov! Keen for it to be updated in BRMS - after seeing your threads on the Stan forums + the pull request here https://github.com/stan-dev/math/pull/3036 I was able to resolve some hiccups in my models by editing the stan code generated by BRMS to use von_mises_lpdf instead of von_mises2_lpdf

paul-buerkner:
Let's just change it now :-)

PR #1678, opened 2024-08-13
remove outdated include statement

venpopov:
After the change for #1633 I got an error that file "fun_von_mises.stan" could not be found. There was a left over include statement in the .family_von_mises() function. Removing it fixes the problem.


paul-buerkner:
thanks!

PR #1687, opened 2024-09-16
Use random effects as predictors in formula via 're' terms

paul-buerkner:


mvuorre:
(Apologies if this is not the right place to comment/ask, happy to move elsewhere.) 

This is a great feature, and its syntax (especially `re(..., resp = "", nlpar = "")`) suggests to me that one can now share parameters across outcomes in multivariate models. One of the tests[^1] already does this by predicting `x` from by-gr random intercepts of `y`. 

1. Since this sharing is now possible for the "random" effects, can it also be somehow done for the "fixed" effects? 
2. Moreover, how would one go about extending the example in that test[^1] to a nonlinear multivariate model? (That is, referring to the `re()` terms by name.) If I understand correctly, the syntax allows for it but it is not clear how one would have to name the parameters to accomplish it.
3. Most concretely, it would be awesome if this came with more examples (and/or was incorporated into a vignette).

Thanks!

[^1]: https://github.com/paul-buerkner/brms/blob/b5f83cbde65c0503580645c3cd172a39da07dc8d/tests/testthat/tests.stancode.R#L1141

paul-buerkner:
1. This will only be possible with brms 3.0

2. and 3. This PR is part of a WIP paper of ours. We will post more examples on this when the paper is out.

paul-buerkner:
Just a note to add NEWS entry before merging.

PR #1698, opened 2024-10-28
brms support for XBX regression

ikosmidis:
Hi @paul-buerkner,

I hope all is well. This PR provides (hopefully) full support for XBX regression through the `xbetax` **rbms** family.  XBX regression is defined in the recent preprint of mine with @zeileis 

> Kosmidis I, [Zeileis A](https://www.zeileis.org/) (2024). Extended-support beta regression for $[0, 1]$ responses.
> [arXiv: 2409.07233](https://arxiv.org/abs/2409.07233)

Specifically, with covariates $x_i$ the likelihood of XBX regression can be defined as
```math
\begin{aligned}
Z_i | X_i = x_i & \sim \text{Beta}(\mu_i \phi_i, (1 - \mu_i) \phi_i) \\
Y^*_i &= (1 + 2 U_i) Z_i - U_i \\
Y_i & = \max(\min(Y^*_i, 1), 0) \\ 
U_i & \sim_\text{ind.} \text{Exponential}(1/\nu) \\
\end{aligned}
```
or, equivalently, and as implemented in the PR
```math
\begin{aligned}
Y_i | X_i = x_i, U_i = u_i & \sim \text{XB}(\mu_i, \phi_i, u_i) \\
U_i & \sim_{\rm ind.} \text{Exponential}(1/\nu) \\
\end{aligned}
```
with $\rm XB$, the extended-support beta distribution as defined in the manuscript. Th $u_i$'s are exceedance parameters, as they determine how much the support of the beta distributon should be extended at the left and right of $(0, 1)$, before censoring, and taking a continuous mixture.

The **betareg** R package already provides the model above, using Gauss-Laguerre quadrature approximation to the log-likelihood.

The approach taken for the **brms** implementation is to use instead $log(u_i) \sim_{\rm ind.} {\rm Normal}(\mu, \sigma^2)$, which was relatively easy to implement with the default priors. In fact, we considered LogNormal when developing the theory, but in our frequentist treatment, we decided in favour of the one extra parameter that the exponential distribution requires, and there was not much evidence for LogNormal in our preliminary experiments.

For example, with the PR we can now do
```
if (requireNamespace("betareg", quietly = TRUE)) {
  data("LossAversion", package = "betareg")
  LossAversion$id <- 1:nrow(LossAversion)
  hmc_xbx <- brm(
      bf(invest ~ grade * (arrangement + age) + male,
           phi ~ arrangement + male + grade,
           u ~ 1 || id),
      data = LossAversion,
      family = "xbetax")
}
```
The new example in `?brm` compares the above HMC fit with the maximum likelihood fit we get from **betareg** with exponential distributions; the fits, up to expectation, are essentially the same. 

The Additions section below includes a list of the additions.

## Questions

1. Is there any way to allow for exponential distributions with mean $\nu$ for the exceedance parameters, allowing for a flat prior in $\nu$?
2. Is there a way to have `u ~ 1 || id`, where `id` is the observation `id`, as the default for `xbetax` families? Correlated varying effects, more complicating grouping for `u` or fixed effects there, while easily accommodated by the facilities in **brms** and fits may be obtainable by HMC, are overkill and are not recommended because both `u` and `phi` control the response variance; the model may end up being formally not identifiable.
3. Is there any way to implement family-specific errors if some conditions about the data are met? e.g. if no zero / one observations are in the responses, then throw a warning and propose the user fits a `"beta"` family; trying `"xbetax"` with flat priors on `u` or the parameters of its distribution can be a challenge for HMC in that case, and there is not much to gain compared to a `"beta"` family fit.

## Notes

- The name `xbetax` for the family (instead of XBX) is used to be close to the`dist = "xbetax"` argument **betareg** uses

- I added **betareg** and **distributions3** for the `XBeta()` distribution and manipulation of it, respectively in Suggests. THese could be coded in **brms** but adding the suggests avoids defining the distributions in multiple places in CRAN (and makes support easier!).

- 👏👏👏👏 for **brms**. Fantastic codebase, and glad I engaged with it!

## Additions

+ DESCRIPTION
  - Suggests now has **betareg** and **distributions3** for the `XBeta()` distribution and manipulation of it, respectively

+ `R/families.R`
  - `brmsfamily()` now supports `link_u`, where u is the exceedance parameter of the extended-support beta distribution
  - `links_dpars()` now has `u = c("log")`
  - `family_bounds.brmsterms()`: `"xbetax" is now in `beta_families`
  - Documentation updates

+ Added `inst/chunks/fun_xbetax.stan`

+ `R/brm.R`
  - Documentation updates
  - Added XBX regression example

+ `R/family-lists.R`
  - Added `.family_xbetax()`

+ `R/log_lik.R`
  - Added `log_lik_xbetax()`

+ `R/posterior_epred.R`
  - Added `posterior_epred_xbetax()`

+ `R/posterior_predict.R`
  - Added `posterior_predict_xbetax()`

+ `R/stan-likelihood.R`
  - Added `stan_log_lik_xbetax()`

+ `R/priors.R`
  - `def_dpar_prior()` now has `u = "gamma(0.01, 0.01)"` if link is identity and `u = "student_t(3, 0, 2.5)"` otherwise
  - `dpar_bounds()` now has `u = list(lb = "0", ub = "")`

+ `R/stan-predictor.R`
  - `stan_dpar_comments()` now has `u = "exceedance parameter"`


paul-buerkner:
Thank you! This looks really nice! I will take a closer look at the PR in the next few days. 

About your questions:


>  1. Is there any way to allow for exponential distributions with mean ν for the exceedance parameters, allowing for a flat prior in ν ?

I don't fully understand. Can you elaborate?


> 2.  Is there a way to have u ~ 1 || id, where id is the observation id, as the default for xbetax families? Correlated varying effects, more complicating grouping for u or fixed effects there, while easily accommodated by the facilities in brms and fits may be obtainable by HMC, are overkill and are not recommended because both u and phi control the response variance; the model may end up being formally not identifiable.

This is unforunately not possible as it would require too much special case coding. It has to remain the responsibility of the user.

> 3. Is there any way to implement family-specific errors if some conditions about the data are met? e.g. if no zero / one observations are in the responses, then throw a warning and propose the user fits a "beta" family; trying "xbetax" with flat priors on u or the parameters of its distribution can be a challenge for HMC in that case, and there is not much to gain compared to a "beta" family fit.

Good question. I will think about a good way to achieve this.

One more quick point: I don't think that such a detailed example of a cool but still niche model should be placed within the `?brm` doc. I am not sure if brms has a good place for such detailed examples. Perhaps in the `?brmsfamily` page. In any case, I don't think the betareg comparison is needed there, which would also allow you to drop the betareg suggests I believe.

ikosmidis:
> Thank you! This looks really nice! I will take a closer look at the PR in the next few days.
> 
> About your questions:
> 
> > 1. Is there any way to allow for exponential distributions with mean ν for the exceedance parameters, allowing for a flat prior in ν ?
> 
> I don't fully understand. Can you elaborate?
> 

The model, as we defined it (see above) has

```math
\begin{aligned}
Y_i | X_i = x_i, U_i = u_i & \sim \text{XB}(\mu_i, \phi_i, u_i) \\
U_i & \sim_{\rm ind.} \text{Exponential}(1/\nu) \\
\end{aligned}
```

However, instead of $U_i \sim_\text{ind.} {\rm Exponential}(1 / \nu)$ (mean $\nu$), I implemented $log(u_i) \sim_{\rm ind.} {\rm Normal}(\mu, \sigma^2)$ in **brms**, by allowing $U_i$ to be independent varying effects per observation and using the default priors for varying effects. We have considered log-normal for $U_i$ in the past, but, in our frequentist treatment we went for Exponential, which introduces only a single unknown parameter to the model. I wonder whether there is a way to (e.g. not going through varying effects?) to define the family to have $U_i$ independent with an $\text{Exponential}(1/\nu)$ prior, and $\nu$ to have a flat prior.

> > 2. Is there a way to have u ~ 1 || id, where id is the observation id, as the default for xbetax families? Correlated varying effects, more complicating grouping for u or fixed effects there, while easily accommodated by the facilities in brms and fits may be obtainable by HMC, are overkill and are not recommended because both u and phi control the response variance; the model may end up being formally not identifiable.
> 
> This is unforunately not possible as it would require too much special case coding. It has to remain the responsibility of the user.

Understood. This is not a big issue with detailed family-specific documentation.

> 
> > 3. Is there any way to implement family-specific errors if some conditions about the data are met? e.g. if no zero / one observations are in the responses, then throw a warning and propose the user fits a "beta" family; trying "xbetax" with flat priors on u or the parameters of its distribution can be a challenge for HMC in that case, and there is not much to gain compared to a "beta" family fit.
> 
> Good question. I will think about a good way to achieve this.

Thanks. This can also be useful to allow for checking that any required-by-the-family packages are attached in the namespace. 

> 
> One more quick point: I don't think that such a detailed example of a cool but still niche model should be placed within the `?brm` doc. I am not sure if brms has a good place for such detailed examples. Perhaps in the `?brmsfamily` page. In any case, I don't think the betareg comparison is needed there, which would also allow you to drop the betareg suggests I believe.

I had considered moving them to `?brmsfamily`, but the existing examples were mostly about defining family objects rather than fitting them, and I did not want to contaminate them :). Happy to put it there. Another option would be to have family vignettes. This would allow the space for more detailed discussion about certain families and their quirks. Happy to produce either.

**betareg** is in suggests because it provides the `XBeta()` function that defines the extended-beta distribution; then the methods of the **distributions3** R package are used to compute log-likelihoods, predictions, etc. I just pushed a version which uses only methods in **betereg** (`dxbeta()`, `rxbeta()`) and a copy of `betareg:::mean_xbeta()`, which we can avoid if we export the latter in a future release of **betareg**. So **distributions3** is now not necessary, but **betareg** in Suggests is.

Best regards
-Ioannis

paul-buerkner:
Thanks!

1. I see. I don't think there is a pretty way to do it. And even the ugly ones that come to mind (injecting Stan code via `stanvar` are probably not going to work super well. Does the regular normal random effects version work well enough?

About the suggested packages: Okay, makes sense.

I will now review the PR.

inst/chunks/fun_xbetax.stan
```diff
@@ -0,0 +1,32 @@
+  /* Extended support beta log-PDF of a single response
+   * Args:
+   *   y: the response value
+   *   mu: mean parameter of the beta distribution
+   *   phi: precision parameter of the beta distribution
+   *   u: the exceedance parameter
+   * Returns:
+   *   a scalar to be added to the log posterior
+   */
+   real xbetax_lpdf(real y, real mu, real phi, real u) {
```
paul-buerkner, line 10: can these functions also be vectorized and if yes, would that speed of the sampling speed? The latest versions of Stan allow function overloading also for user-defined functions so you can define the same function multiple times just with different argument signatures.

It is totally alright not to have a vectorized version. I just wanted to ask to be sure.

paul-buerkner (changes requested):
I think a lot of this PR already looks good and I hope my comments help you to make it merge ready.

Please also add tests for the new functionality. In particular, in addition to some basic tests (see the tests folder), you could add what is your current detailed example in the local tests folder (in tests/local/tests-models-5.R). There it would be great because we had direct validation against betareg even.

ikosmidis:
> Thanks!
> 
> 1. I see. I don't think there is a pretty way to do it. And even the ugly ones that come to mind (injecting Stan code via `stanvar` are probably not going to work super well. Does the regular normal random effects version work well enough?

Yes. This is fine; both exponential and log normal give essentially the same fit; I would rather avoid injecting Stan code. Another advantage of the current specification is that the "xbetax" family directly benefits from future enhancements/features in terms, etc., and the user can experiment with other specifications for the exceedances $U_i$ (though the latter may be asking for trouble for some data sets).

> 
> About the suggested packages: Okay, makes sense.
> 
> I will now review the PR.

Great stuff; thanks for doing that so promptly!

I will start preparing a vignette just in case; it can either go in the **brms** vignette, or I can publish it on my page for future reference (and testing).

paul-buerkner:
I think a vignette would be great! I don't think it can go into brms itself though because CRAN already complains about package size and every vignette contributes to it. So I had to stop adding new vignettes to the package unfortunately.

R/brm.R
```diff
@@ -412,6 +412,56 @@
 #' summary(fit7)
 #' conditional_effects(fit7)
 #'
+#' # Fit an extended-support beta mixture model (Kosmidis & Zeileis,
```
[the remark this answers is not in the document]
ikosmidis: Done in [d706c6d](https://github.com/paul-buerkner/brms/pull/1698/commits/d706c6d29e67cd8cc6e55ff59dd0a204821c5aa3)

R/families.R
```diff
@@ -197,6 +219,7 @@ brmsfamily <- function(family, link = NULL, link_sigma = "log",
                        link_bias = "logit", link_xi = "log1p",
                        link_alpha = "identity",
                        link_quantile = "logit",
+                       link_u = "log",
```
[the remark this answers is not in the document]
ikosmidis: I understand. I saw sigma, but I thought I would stay on the same side as `sigma = def_scale_prior(x)` in `def_par_prior()`. 

Ideally, I would like to have a parameter with `"gamma(0.01, 0.01)"` in `def_par_prior()` if link is `"identity"` and `"student_t(3, 0, 2.5)"` otherwise. For example, I tried with `kappa` and things go a bit off due to the `def_par_prior()` specification (I think),

Such specifications are present for "shape" and "phi", but both names are, unfortunately, quite loaded in a beta regression context, and I'd rather avoid them to avoid conflicts in notation. 

Is it OK to keep the extra parameter for now (perhaps with another greek letter name, say `epsilon` or a specific one like `exceedance` (or `u` as is in our paper), and then revisit in brms 3.0 (that'd be a great feature; I looked for it, when I was putting `xbetax` together :))? Or perhaps there is another way?

I will wait for your input before I continue, as, I think, the name of the parameter is important both for the codebase and the documentation. 

Many thanks,
Ioannis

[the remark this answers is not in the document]
paul-buerkner: You can already achieve family-specific priors for dpars via the `prior` slot in the `.family_name` function (quite a new feature). See `.family_negbinomial` in `family-lists.R` as an example,. With this feature, could you perhaps use the existing `sigma` or `kappa`?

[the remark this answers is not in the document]
ikosmidis: Cool! Now implemented and seems to work as before - commit [641fbd7](https://github.com/paul-buerkner/brms/pull/1698/commits/641fbd726fbaecf27770ac978896bf32f8775379)

R/posterior_epred.R
```diff
@@ -900,3 +900,21 @@ pp_expect <- function(object, ...) {
...
+    d <- (1 + 2 * nu)
+    q0 <- nu / d
+    q1 <- (1 + nu) / d
+    t3 <- pbeta(q1, a, b)
+    t1 <- d * mu * (pbeta(q1, a + 1, b) - pbeta(q0, a + 1, b))
+    t2 <- nu * (t3 - pbeta(q0, a, b))
+    1 + t1 - t2 - t3
+}
+
+posterior_epred_xbetax <- function(prep) {
```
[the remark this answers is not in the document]
ikosmidis: This was a remnant of using **distributions3**. **betareg::dxbeta()** and **betareg::rxbeta()** respect input dimensions, so there is no need for such manipulations now - commit [641fbd7](https://github.com/paul-buerkner/brms/pull/1698/commits/641fbd726fbaecf27770ac978896bf32f8775379)

R/posterior_predict.R
```diff
@@ -1062,3 +1062,21 @@ check_discrete_trunc_bounds <- function(x, lb = NULL, ub = NULL, thres = 0.01) {
   }
   round(x)
 }
+
+get_xbetax <- function(i, prep) {
+    if (is.numeric(i)) {
+        mu <- brms::get_dpar(prep, "mu", i = i)
+        phi <- brms::get_dpar(prep, "phi", i = i)
+        u <- brms::get_dpar(prep, "u", i = i)
+    } else {
```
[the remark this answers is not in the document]
ikosmidis: Again not necessary in the new codebase - commit [641fbd7](https://github.com/paul-buerkner/brms/pull/1698/commits/641fbd726fbaecf27770ac978896bf32f8775379)

R/posterior_predict.R
```diff
@@ -1062,3 +1062,21 @@ check_discrete_trunc_bounds <- function(x, lb = NULL, ub = NULL, thres = 0.01) {
...
+        mu <- brms::get_dpar(prep, "mu")
+        phi <- brms::get_dpar(prep, "phi")
+        u <- brms::get_dpar(prep, "u")
+    }
+    list(mu = c(mu), phi = c(phi), u = c(u))
```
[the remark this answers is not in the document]
ikosmidis: not necessary in the new codebase - commit [641fbd7](https://github.com/paul-buerkner/brms/pull/1698/commits/641fbd726fbaecf27770ac978896bf32f8775379)

R/family-lists.R
```diff
@@ -611,3 +611,20 @@
...
+  list(
+    links = c(
+      "logit", "probit", "probit_approx", "cloglog",
+      "cauchit", "softit", "identity", "log"
+    ),
+    dpars = c("mu", "phi", "u"),
+    type = "real",
+    ybounds = c(0, 1),
+    closed = c(TRUE, TRUE),
+    ad = c("weights", "subset", "cens", "trunc", "index"),
```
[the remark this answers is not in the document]
ikosmidis: That was from copy-and-paste when I set the family up. No need for "cens"/"trunc" options. Now gone - commit [641fbd7](https://github.com/paul-buerkner/brms/pull/1698/commits/641fbd726fbaecf27770ac978896bf32f8775379)

R/posterior_epred.R
```diff
@@ -900,3 +900,21 @@ pp_expect <- function(object, ...) {
            "Please use 'posterior_epred' instead.")
   UseMethod("posterior_epred")
 }
+
+## From betareg
```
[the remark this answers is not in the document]
ikosmidis: Done - commit https://github.com/paul-buerkner/brms/pull/1698/commits/4f561ee9b6748d72a2146388474af60077de4731

R/posterior_predict.R
```diff
@@ -1062,3 +1062,21 @@ check_discrete_trunc_bounds <- function(x, lb = NULL, ub = NULL, thres = 0.01) {
...
+}
+
+posterior_predict_xbetax <- function(i, prep, ...) {
```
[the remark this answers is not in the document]
ikosmidis: No reason really; I suspect `rconitnuous()` allows for more options that `r*`, so I'm now using that - commit [d8b7e8c](https://github.com/paul-buerkner/brms/pull/1698/commits/d8b7e8cc54c2ec981c3452f90c8f721540ae0d9c)

inst/chunks/fun_xbetax.stan
ikosmidis: Happy to give that a shot and run some comparisons. Do you know of any vectorized implementation already in **brms** or elsewhere that I can look at?

paul-buerkner: Looking at it again, I think we do not need a vectorized version in this case. I saw you already started and I am sorry I caused you extra work. The reason is the getting speed improvement out of the vectorized version requires a bit more effort since we need to group the y into those that fall into the different categories defined by the if statements. Kind of the same thing as for zero-inflated models, which are also not vectorized for now. In order to safe your time, I would say that the non-vectorized version alone is good enough.

R/posterior_predict.R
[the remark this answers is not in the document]
paul-buerkner: Thanks!

inst/chunks/fun_xbetax.stan
ikosmidis: No worries. I'll give it another shot and if all fails we default to the scalar version.

paul-buerkner: okay! to be clear: we need the scalar version in any case. an additional vectoriced version may just be helpful to speed up sampling. but this will only be beneficial if the lpdf statements are victorized that is called at the same time for multiple y.

ikosmidis: Sure. Just for info. How is the choice between vectorized and scalar being made internally? I currently, simply have

```r
stan_log_lik_xbetax <- function(bterms, ...) {
    p <- stan_log_lik_dpars(bterms, reqn = TRUE)
    sdist("xbetax", p$mu, p$phi, p$kappa, vec = FALSE)
}
```

for scalar and set `reqn = FALSE` and `vec = TRUE` for vectorized. There must be a more direct way, from the interface, perhaps?

paul-buerkner: Suppose you have build a vectorized version vectorized only over mu but not over phi and kappa. Then you `reqn = TRUE` only if phi or kappa is predicted or some addition terms are used (like weights), which prevents vectorization from being valid. At the same time, vec = TRUE is only valid if you only predict mu (and keep phi and kappy constant) because you have only build a vectorized version for this case. 

Again, I believe my suggestion to make a vectorized version was not a good once since your family would likely require a bit of work to benefit from vectorization.

ikosmidis: I think I got a vectorized version in all mu, phi, kappa in commit [31193c8](https://github.com/paul-buerkner/brms/pull/1698/commits/31193c867a6368c2fa2b8b9c34e49a26522a74f3).

Would
```
stan_log_lik_xbetax <- function(bterms, ...) {
    p <- stan_log_lik_dpars(bterms, reqn = FALSE)
    sdist("xbetax", p$mu, p$phi, p$kappa, vec = TRUE)
}
```
be enough to use it? I see

```
R> stan_log_lik_dpars(bterms, reqn = FALSE)
$mu
[1] "mu"

$phi
[1] "phi"

$kappa
[1] "kappa"
```

and 

```
sdist("xbetax", p$mu, p$phi, p$kappa, vec = TRUE)
$dist
[1] "xbetax"

$args
[1] "mu, phi, kappa"

$vec
[1] TRUE

$shift
[1] ""

attr(,"class")
[1] "sdist"
```

and in the output of `stancode()` we have `target += xbetax_lpdf(Y | mu, phi, kappa)`

Does that do it?

P.S.: I might need to check for only ones, only zeros or both ones and zeros but that's easy to do now

wds15:
Instead of vignettes you can have so called articles. The articles go only on the pkgdown generated web site and not on CRAN. For RBesT (https://opensource.nibr.com/RBesT/), I have dropped but just one vignette and moved the rest to be an article.

inst/chunks/fun_xbetax.stan
paul-buerkner: I think the vectorized version looks good. Now the problem is that this will only work if all 3 parameters are predicted, which will often not be the case. So I would argue you vectorieze (only) the most common case which is, I assume only mu being predicted. Can you check if that leads to relevant speed-ups? 

I will fix the reqn myself before I merge.

paul-buerkner:
Thanks @wds15 that could also be a good option, even for brms more generally.

inst/chunks/fun_xbetax.stan
ikosmidis: Wit all three parameters predicted, and a single run, I get:

Vectorized:
```
Chain 1:  Elapsed Time: 232.716 seconds (Warm-up)
Chain 1:                50.756 seconds (Sampling)
Chain 1:                283.472 seconds (Total)
```

Scalar:
```
Chain 1:  Elapsed Time: 253.758 seconds (Warm-up)
Chain 1:                108.226 seconds (Sampling)
Chain 1:                361.984 seconds (Total)
```

which looks good. I will experiment with only one or two parameters being predicted, and report back

zeileis:
I wanted to recommend the same. :+1: We have used the same idea with a slightly different technical setup for the [colorspace](https://CRAN.R-project.org/package=colorspace) package:

- There is a bunch of Rmd vignettes in the [vignettes folder](https://github.com/r-forge/colorspace/tree/master/pkg/colorspace/vignettes) of the package.
- But all except one are in [.Rbuildignore](https://github.com/r-forge/colorspace/blob/master/pkg/colorspace/.Rbuildignore#L15-L24) so that they are not included in the package itself (and hence not on CRAN).
- The introductory vignette contains links to all other vignettes and is used as the only Rmd [CRAN vignette](https://CRAN.R-project.org/web/packages/colorspace/vignettes/colorspace.html) and the [Get started](http://colorspace.R-Forge.R-project.org/articles/colorspace.html) page on the pkgdown site.

paul-buerkner:
Let me know when you think this PR is ready for another round of review.

inst/chunks/fun_xbetax.stan
ikosmidis: `inst/chunks/fun_xbetax.stan` now has all eight versions of `xbeta_lpdf`, depending on whether `mu`, `phi`, `kappa` are real or vector. So, given the right call, STAN should figure out which one to use. 

What I could not get my head around from the limited time I had to experiment is how **brms** decides which version to use.  For some reason, the default `stan_log_lik_adj(bterms)` did not do it, but I admit I did not dig further.

For now, I forced the use of the non-vectorized version in the commit for now, having

```
stan_log_lik_xbetax <- function(bterms, ...) {
    p <- stan_log_lik_dpars(bterms, reqn = TRUE)
    sdist("xbetax", p$mu, p$phi, p$kappa, vec = FALSE)
}
```
See, https://github.com/paul-buerkner/brms/pull/1698/commits/6524911fec13e06162e3f9c00b72d9e4324665c7

I'm looking forward to seeing how that choice can be made automatically, when you merge.

R/families.R
```diff
@@ -113,26 +113,72 @@
...
 #'
-#'   Below, we list all possible links for each family.
-#'   The first link mentioned for each family is the default.
-#'   \itemize{
-#'   \item{Families \code{gaussian}, \code{student}, \code{skew_normal},
-#'   \code{exgaussian}, \code{asym_laplace}, and \code{gen_extreme_value}
-#'   support the links (as names) \code{identity}, \code{log}, \code{inverse},
-#'   and \code{softplus}.}
+#'
+#'   \item{Family \code{xbetax} allows to estimate extended-support
```
paul-buerkner, line 118: Now a bit long perhaps (sorry). I will adjust this a bit before merging.

R/family-lists.R
```diff
@@ -611,3 +611,31 @@
...
+        ybounds = c(0, 1),
+        closed = c(TRUE, TRUE),
+        ad = c("weights", "subset", "index"),
+        include = "fun_xbetax.stan",
+        prior = function(dpar, link = "identity", ...) {
+        if (dpar == "kappa") {
+            if (link == "identity") {
+                return("gamma(0.01, 0.01)")
+            }
+            else {
```
paul-buerkner, line 632: else should be on the same line as the if closing bracket }

inst/chunks/fun_xbetax.stan
```diff
@@ -0,0 +1,118 @@
...
+
+     real ll_zer = beta_proportion_lcdf(kappa[zer] ./ (1 + 2 * kappa[zer]) | mu[zer], phi[zer]);
+     real ll_one = beta_proportion_lccdf((1 + kappa[one]) ./ (1 + 2 * kappa[one]) | mu[one], phi[one]);
+     real ll_oth = beta_proportion_lpdf((y[oth] + kappa[oth]) ./ (1 + 2 * kappa[oth]) | mu[oth], phi[oth]) - sum(log(1 + 2 * kappa[oth]));
+     real ll = ll_zer + ll_one + ll_oth;
+     return ll;	
+   }
+
+   real xbetax_lpdf(vector y, vector mu, real phi, vector kappa) {
+      vector[1] phiv;
```
paul-buerkner, line 64: I don't think this will work as you expect. You also need to handle the fact that phi for example is not of length 1 not of length N. I believe your fully vectorized version expects all vectors to be of the same length as y. This issue holds for all these partial vector versions.

R/stan-likelihood.R
```diff
@@ -1401,3 +1401,11 @@ tp <- function(wsp = 2) {
   wsp <- collapse(rep(" ", wsp))
   paste0(wsp, "target += ")
 }
+
+## FIXME: IK, 30 Oct 2024: what should the default values of reqn and
+## vec be in order to be benefitting from vectorization when possible?
+stan_log_lik_xbetax <- function(bterms, ...) {
+    p <- stan_log_lik_dpars(bterms, reqn = TRUE)
+    sdist("xbetax", p$mu, p$phi, p$kappa, vec = FALSE)
```
paul-buerkner, line 1409: if you have all vectorized versions really implemented and want to test them, you can set reqn = FALSE and vec = TRUE. At least for your testing this will work.

paul-buerkner (changes requested):
Thanks! The PR has become much better! Some issues remain which I have commented on.

R/family-lists.R
ikosmidis: commit https://github.com/paul-buerkner/brms/pull/1698/commits/3f3f8e9d3ba9af3a0a2caea0f4151d0045aaa046

R/families.R
ikosmidis: 😄 OK. Happy to take a look at the edited version before merging

inst/chunks/fun_xbetax.stan
ikosmidis: OK. I see now. The vectorized versions in commit [e8b3352](https://github.com/paul-buerkner/brms/pull/1698/commits/e8b3352389c3949eb46e9c7f166b0157d1247080) should now work as expected. I tested things around with the `LossAversion` data (the example I had in `?brm` and will appear in the vignette) and despite it being relatively small we can already see speedups.

Here is a comparison in elapsed times between vectorized and non-vectorized implementations
Screenshot 2024-11-01 at 12 12 32

where the x axis is what components are passed to `brm`, based on 20 replications per model with `init = 0`, `chains = 1`, `iter = 2000`.

The fits between vectorized and non-vectorized implementations are the same.

I can also share the script that reproduces the figure and carries out the tests on the `brmsfit` objects.  If you want that what is the right place to push it (`tests/local`?)?.

R/stan-likelihood.R
ikosmidis: See https://github.com/paul-buerkner/brms/pull/1698#discussion_r1824060648_

inst/chunks/fun_xbetax.stan
ikosmidis: One thing I wanted to ask: is there any value in having vectorized functions with real `mu`, e.g.

```
xbetax_lpdf(vector y, real mu, vector phi, real kappa)
```
? These are not difficult to setup, but I am not sure if `rbm` would ever pass real `mu` to the compiled STAN code

paul-buerkner: Thank you! I will now do some edits to the PR and let you know for a final check before merging.

About your question: mu will indeed always be a vector in brms so no need to have a vectorized real mu implementation.

paul-buerkner:
As I am editing your PR I decided to substantially shorten your description in the ?brmsfamily doc. You can write a bit more (but please not so many) details in the brmsfamilies vignette.

ikosmidis:
> As I am editing your PR I decided to substantially shorten your description in the ?brmsfamily doc. You can write a bit more (but please not so many) details in the brmsfamilies vignette.

I have added a bit more details; mainly to advise users when `xbeta` is better to use than hurdle beta families.

ikosmidis:
> I have made some edits to the PR. I think it should basically be ready now.
> 
> I noticed that the posterior_predict test was still missing. Was this intentional?

Added some tests, too.

ikosmidis:
All looks good to me now, and my local tests pass after your edits. 

Just a few things that I would like to keep on the todo list, related to the discussion above:

1. Is there any way to implement family-specific errors if some conditions about the data are met? e.g. if no zero / one observations are in the responses, then throw a warning and propose the user fits a "beta" family; trying `xbeta` or the parameters of its distribution can be a challenge for HMC in that case, and there is not much to gain compared to a "beta" family fit. Similarly, if 0/1's are detected with `beta` family, an error can inform the user to use `xbeta` or the one/two-part hurdle models instead, directing them to the documentation for details.

2. I will prepare a blog post for my web page and release it when the new version is on CRAN. I'd be happy to convert it to a vignette when/if family-specific vignettes come to the package pages / CRAN.

I'm happy for this to be merged. 

Thanks for the input and for being open to having  the`xbeta` family in **rbms**.

paul-buerkner:
Thank you!

1. I think it makes sense to open a new issue for that as I have to implement the mechanism allowing for such errors and messages.
2. Thank you for writing a vignette!

I will now run a few more checks and then merge if they pass.

paul-buerkner:
Unrelated question: Could a similar strategy that you employ also be used for lower-bounded families such as, say, Gamma or lognormal? This way, one may be able to avoid hurdle_gamma or hurdle_lognormal as well if the aim is to model a single rather than multiple different responses processes.

paul-buerkner:
checks fail online for unrelated reasons, but pass locally. So i am going to merge this PR.

Thank you for contributing to brms and providing this amazing new feature!

ikosmidis:
> Regarding the applicability to other lower-bounded families: Yes, in principle, this is doable and worth trying out.
> 
> I think it is not always quite as appealing as in the beta case where we can leverage results for the four-parameter beta distribution and also get the connection to the normal distribution.
> 
> For the gamma distribution, I also know that the idea of a censored shifted gamma distribution has been proposed by [Baran & Nemoda 2016, Environmetrics](https://doi.org/10.1002/env.239). So this is similar in spirit but without the continuous mixture of the shift (corresponding to our exceedance parameter) for shrinking it twoards zero.

Yes, indeed. Beta is convenient because we can extend the support, and we establish results linking the resulting distribution to the normal linear model.

However, nothing prevents trying the same methodology to other distributions with bounded support. For example, with Gamma or Log-normal, one could subtract $\kappa$ from a Gamma-distributed random variable and link that to varying/population effects. I guess this can be implemented in full generality in **brms**. Still, identifiability issues must be examined carefully on an ad-hoc basis, especially for location-scale families with location and scale specifications.

Issue #1705, opened 2024-11-15
Feature request: Change default `brm(iter = 2000)` to `brm(iter = getOption("brms.iter", 2000)`


mvuorre:
A small quality of life improvement request:

I frequently run a bunch of models and don't initially want to wait for 2000 iterations. It would be faster if I could set a global option (e.g.) `options(brms.iter = 200)` and have `brm()` use that option. If the default for that option was 2000, it would default to current behavior. Does this make sense? If so I am happy to submit a PR.

paul-buerkner:
sounds good!

Matti Vuorre ***@***.***> schrieb am Fr., 15. Nov. 2024,
08:48:

mvuorre:
Implemented in https://github.com/paul-buerkner/brms/commit/cbd7a2612930d2e85130984d276b95ffa8ecb292, closing.

PR #1707, opened 2024-11-18
Use brms.iter option (#1705)

mvuorre:
Changes the default `brm(iter = 2000)` to `brm(iter = getOption("brms.iter", 2000))` (#1705) & revises associated documentation.

I tested this locally using examples in README.Rmd, but did not add tests because other similar arguments in `brm()` are not tested either.

Going forward it might make sense to think about using `options()` for e.g. `brm(control = <x>)` and others.


paul-buerkner:
Looks good, thank you!

PR #1714, opened 2024-11-27
fix misplaced paren in .loo_R2

avehtari:
There is a misplaced paren in .loo_R2 causing wrong results. 

I first noticed discrepancy between manually computed point estimate and the distribution, then discrepancy compared to `bayesboot` and after 30mins of staring found the misplaced paren.


paul-buerkner:
Thank you for finding and fixing this!

PR #1717, opened 2024-12-12
Add the brms_seed to loo_R2

### LINKED ISSUE #1713: loo_R2: add seed argument
loo_R2 used Bayesian bootstrap for quantifying the uncertainty. To be able to quantify the corresponding uncertainty when comparing two models, the same random seed should be used in Bayesian bootstrap

With seed argument and the following code
```
loo_r2_diff <- loo_R2(fit1, re_formula=NA, summary=FALSE, seed=1) - loo_R2(fit2, re_formula=NA, summary=FALSE, seed=1)
```
`loo_r2_diff` would contain Bayesian bootstrap draws describing the uncertainty in model comparison (without common seed the result would be not useful).

The relevant part in `.loo_R2()´ would be to wrap the call `rexp()` with a local seed that has been passed as one of the arguments
```
  withr::with_seed(SEED, {
    exp_draws <- matrix(rexp(S * N, rate = 1), nrow = S, ncol = N)
  })         
```
I used `withr::with_seed()` as an easy illustration, but there are ways to do the same without adding new dependency.

mcol:
This allows to fix the seed in `loo_R2` and restore the RNG state when leaving the function. The implementation and the choice of the argument name follow what is already done in `get_refmodel.brmsfit`.

Fixes  #1713.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@
 
 * Fit extended-support Beta models via family `xbeta` 
 thanks to Ioannis Kosmidis. (#1698)
+* Add the `brms_seed` argument to `loo_R2` thanks to
+Marco Colombo. (#1713)
 
 ### Bug Fixes
 
--- a/R/loo_predict.R
+++ b/R/loo_predict.R
@@ -213,7 +213,8 @@ E_loo_value <- function(x, psis_object, type = "mean", probs = 0.5) {
 #' @export loo_R2
 #' @export
 loo_R2.brmsfit <- function(object, resp = NULL, summary = TRUE,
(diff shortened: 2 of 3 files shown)


paul-buerkner:
Cool, thank you! Is there a reason you call the argument `brms_seed` rather than just `seed`?

mcol:
Not particularly, I followed what was done in `get_refmodel.brmsfit`. I can change it simply to `seed` if you prefer.

paul-buerkner:
yes I would prefer that. thank you!

Marco Colombo ***@***.***> schrieb am Di., 17. Dez. 2024,
19:57:

paul-buerkner:
Great, thank you! Merging now :-)

PR #1724, opened 2025-01-14
initial work on lprior tags

### LINKED ISSUE #1585: Interface for defining which prior log densities are stored for priorsense
Currently, `brms` stores all prior log densities in `lprior` variable. This is fine as the default behavior and makes it easy to use `priorsense` for quick tests without increasing the memory usage too much. To be able to focus on specific priors, we could allow priors to be tagged with labels.

For example, using a modified version of the example from `set_prior` doc

```
make_stancode(rating ~ treat + period + carry + (1|subject),
              data = inhaler, family = cumulative(),
              prior = c(prior(normal(0,10), class = "b", lprior = "b"),
                        prior(normal(1,2), class = "b", coef = "treat", lprior = "treat"),
                        prior(cauchy(0,2), class = "sd", group = "subject", coef = "Intercept"))
```

This would store the corresponding priors in variables `lprior_b` and `lprior_treat`. The same tag could be used for several priors, in which case the log density contributions from those priors would be summed together.

ping @n-kall

n-kall:
This would allow for tagging priors for inclusion in different lprior variables, enabling priorsense to selectively power-scale priors. It would fix #1585

### DIFF
--- a/R/priors.R
+++ b/R/priors.R
@@ -360,8 +360,8 @@
 #' @export
 set_prior <- function(prior, class = "b", coef = "", group = "",
                       resp = "", dpar = "", nlpar = "",
-                      lb = NA, ub = NA, check = TRUE) {
-  input <- nlist(prior, class, coef, group, resp, dpar, nlpar, lb, ub, check)
+                      lb = NA, ub = NA, lprior = "", check = TRUE) {
+  input <- nlist(prior, class, coef, group, resp, dpar, nlpar, lb, ub, lprior, check)
   input <- try(as.data.frame(input), silent = TRUE)
   if (is_try_error(input)) {
     stop2("Processing arguments of 'set_prior' has failed:\n", input)
@@ -375,7 +375,7 @@ set_prior <- function(prior, class = "b", coef = "", group = "",
 
 # validate arguments passed to 'set_prior'
 .set_prior <- function(prior, class, coef, group, resp,
-                       dpar, nlpar, lb, ub, check) {
+                       dpar, nlpar, lb, ub, lprior, check) {
   prior <- as_one_character(prior)
   class <- as_one_character(class)
   group <- as_one_character(group)
@@ -386,16 +386,17 @@ set_prior <- function(prior, class = "b", coef = "", group = "",
   check <- as_one_logical(check)
   lb <- as_one_character(lb, allow_na = TRUE)
   ub <- as_one_character(ub, allow_na = TRUE)
+  lprior <- as_one_character(lprior)
   if (dpar == "mu") {
     # distributional parameter 'mu' is currently implicit #1368
     dpar <- ""
   }
   if (!check) {
     # prior will be added to the log-posterior as is
-    class <- coef <- group <- resp <- dpar <- nlpar <- lb <- ub <- ""
+    class <- coef <- group <- resp <- dpar <- nlpar <- lb <- ub <- lprior <- ""
   }
   source <- "user"
-  out <- nlist(prior, source, class, coef, group, resp, dpar, nlpar, lb, ub)
+  out <- nlist(prior, source, class, coef, group, resp, dpar, nlpar, lb, ub, lprior)
   do_call(brmsprior, out)
 }
 
@@ -558,7 +559,7 @@ default_prior.default <- function(object, data, family = gaussian(), autocor = N
   # explicitly label default priors as such
   prior$source <- "default"
   # apply 'unique' as the same prior may have been included multiple times
-  to_order <- with(prior, order(resp, dpar, nlpar, class, group, coef))
+  to_order <- with(prior, order(resp, dpar, nlpar, class, group, coef, lprior))
   prior <- unique(prior[to_order, , drop = FALSE])
   rownames(prior) <- NULL
   class(prior) <- c("brmsprior", "data.frame")
@@ -1565,7 +1566,7 @@ get_sample_prior <- function(prior) {
 # create data.frames containing prior information
 brmsprior <- function(prior = "", class = "", coef = "", group = "",
                       resp = "", dpar = "", nlpar = "", lb = "", ub = "",
-                      source = "", ls = list()) {
+                      lprior = "", source = "", ls = list()) {
   if (length(ls)) {
     if (is.null(names(ls))) {
       stop("Argument 'ls' must be named.")
@@ -1580,7 +1581,7 @@ brmsprior <- function(prior = "", class = "", coef = "", group = "",
   }
   out <- data.frame(
     prior, class, coef, group,
-    resp, dpar, nlpar, lb, ub, source,
+    resp, dpar, nlpar, lb, ub, lprior, source,
     stringsAsFactors = FALSE
   )
   class(out) <- c("brmsprior", "data.frame")
@@ -1594,7 +1595,7 @@ empty_prior <- function() {
   brmsprior(
     prior = char0, source = char0, class = char0,
     coef = char0, group = char0, resp = char0,
-    dpar = char0, nlpar = char0, lb = char0, ub = char0
+    dpar = char0, nlpar = char0, lb = char0, ub = char0, lprior = char0
   )
 }
 
@@ -1623,7 +1624,7 @@ prior_bounds <- function(prior) {
 # all columns of brmsprior objects
 all_cols_prior <- function() {
   c("prior", "class", "coef", "group", "resp",
-    "dpar", "nlpar", "lb", "ub", "source")
+    "dpar", "nlpar", "lb", "ub", "lprior", "source")
 }
 
 # relevant columns for duplication checks in brmsprior objects
@@ -1915,7 +1916,7 @@ as.brmsprior <- function(x) {
 
   defaults <- c(
     class = "b", coef = "", group = "", resp = "",
-    dpar = "", nlpar = "", lb = NA, ub = NA
+    dpar = "", nlpar = "", lb = NA, ub = NA, lprior = ""
   )
   for (v in names(defaults)) {
     if (!v %in% names(x)) {
--- a/R/stan-prior.R
+++ b/R/stan-prior.R
@@ -95,6 +95,7 @@ stan_prior <- function(prior, class, coef = NULL, group = NULL,
           c(index) <- j
         }
         prior_ij <- subset2(prior, coef = coef[i, j])
+        lprior_tag <- prior_ij$lprior
         if (NROW(px) > 1L) {
           # disambiguate priors of coefficients with the same name
           # coming from different model components
@@ -131,7 +132,13 @@ stan_prior <- function(prior, class, coef = NULL, group = NULL,
               coef_prior, par_ij, broadcast = broadcast,
               bound = bound, resp = px$resp[1], normalize = normalize
             )
+            # add to the lprior
             str_add(out$tpar_prior) <- paste0(lpp(), coef_prior, ";\n")
+            # add to the lprior of the tag if specified
+            if (!is.null(lprior_tag) && lprior_tag != "") {
+              str_add(out$tpar_prior) <- paste0(lpp(tag = lprior_tag), coef_prior, ";\n")
+            }
+
           }
         }
       }
@@ -241,7 +248,7 @@ stan_base_prior <- function(prior, col = "prior", sel_prior = NULL, ...) {
       return(brmsprior()[, col])
     }
   }
-  vars <- c("group", "nlpar", "dpar", "resp", "class")
+  vars <- c("group", "nlpar", "dpar", "resp", "class", "lprior")
   for (v in vars) {
     take <- nzchar(prior[[v]])
     if (any(take)) {
@@ -698,7 +705,11 @@ stopif_prior_bound <- function(prior, class, ...) {
 }
 
 # lprior plus equal
-lpp <- function(wsp = 2) {
+lpp <- function(wsp = 2, tag = NULL) {
   wsp <- collapse(rep(" ", wsp))
-  paste0(wsp, "lprior += ")
+  if (is.null(tag)) {
+    paste0(wsp, "lprior", " += ")
+  } else {
+    paste0(wsp, "lprior_", tag, " += ")
+  }
 }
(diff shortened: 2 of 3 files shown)


paul-buerkner:
I think this looks pretty good already. You mentioned that so far it only works with the b coefficents but I couldn't readily see this limitation in your code. Can you elaborate?

Also, could you show an example model here to illustrate the features use?

n-kall:
Certainly, here is an example:

Basically, the current code works for priors that are set on the coefficient level, i.e. `has_coef_prior == TRUE` [see here](https://github.com/n-kall/brms/blob/df007fbe1707dbdf6fb128154d13921e4951363c/R/stan-prior.R#L83)

I wasn't sure how to properly handle the lprior tagging for a class when there is no individual coef prior. Now the lprior variables are created, but not added to if there is no individual coefficient prior.


```r
prior <- c(
  prior(normal(0, 5), coef = "K1", lprior = "covariates"), # works
  prior(normal(0, 1), coef = "N1", lprior = "treatment"), # works
  prior(normal(0, 100), class = "Intercept", lprior = "Intercept"), # does not work
  prior(normal(0, 10), class = "sigma", lprior = "sigma") # does not work
)


x <- brm(yield ~ N + K + P, data = npk,
         prior = prior, backend = "cmdstanr",
         empty = TRUE
         )

prior_summary(x)

s <- stancode(x)

s


##  > prior_summary(x)
##           prior     class coef group resp dpar nlpar lb ub     lprior       source
##          (flat)         b                                                  default
##    normal(0, 5)         b   K1                             covariates         user
##    normal(0, 1)         b   N1                              treatment         user
##          (flat)         b   P1                                        (vectorized)
##  normal(0, 100) Intercept                                   Intercept         user
##   normal(0, 10)     sigma                             0         sigma         user
## > s <- stancode(x)
## > s
## // generated with brms 2.22.8
## functions {
## }
## data {
##   int<lower=1> N;  // total number of observations
##   vector[N] Y;  // response variable
##   int<lower=1> K;  // number of population-level effects
##   matrix[N, K] X;  // population-level design matrix
##   int<lower=1> Kc;  // number of population-level effects after centering
##   int prior_only;  // should the likelihood be ignored?
## }
## transformed data {
##   matrix[N, Kc] Xc;  // centered version of X without an intercept
##   vector[Kc] means_X;  // column means of X before centering
##   for (i in 2:K) {
##     means_X[i - 1] = mean(X[, i]);
##     Xc[, i - 1] = X[, i] - means_X[i - 1];
##   }
## }
## parameters {
##   vector[Kc] b;  // regression coefficients
##   real Intercept;  // temporary intercept for centered predictors
##   real<lower=0> sigma;  // dispersion parameter
## }
## transformed parameters {
##   real lprior = 0;  // prior contributions to the log posterior
##   real lprior_covariates = 0;
##   real lprior_treatment = 0;
##   real lprior_Intercept = 0;
##   real lprior_sigma = 0;
##   lprior += normal_lpdf(b[1] | 0, 1);
##   lprior_treatment += normal_lpdf(b[1] | 0, 1);
##   lprior += normal_lpdf(b[2] | 0, 5);
##   lprior_covariates += normal_lpdf(b[2] | 0, 5);
##   lprior += normal_lpdf(Intercept | 0, 100);
##   lprior += normal_lpdf(sigma | 0, 10)
##     - 1 * normal_lccdf(0 | 0, 10);
## }
## model {
##   // likelihood including constants
##   if (!prior_only) {
##     target += normal_id_glm_lpdf(Y | Xc, Intercept, b, sigma);
##   }
##   // priors including constants
##   target += lprior;
## }
## generated quantities {
##   // actual population-level intercept
##   real b_Intercept = Intercept - dot_product(means_X, b);
## }
```

paul-buerkner:
I see. okay, I will take a look at make it work for all priors even without the `coef` specified

paul-buerkner:
I have now updated your PR so it should now work for all priors. I have also improved the logic of adding priors to the lprior variables to avoid double computation of the prior lpdf statements. Can you double check that the logic is correct?

Some further comments:
- Upon now seeing it in action, I am not sure we should call the argument to `set_prior` and friends `lprior`. It reads as if people should provide a log prior there, which is not what we want to indicate. Do you think just having the argument named `tag` or so could be better? I am oben for other suggestions
- However we call the argument, it would still need to be documented in `set_prior` at least including an example. 
- In the end, could you add some tests to check that the generated stan code is correct?

n-kall:
Great, thanks! I will check it out.

Regarding the naming, originally I had used `tag` also, but then I wasn't sure if it is distinct enough from `group`. But yes, `lprior` is unclear and `tag` could work with correct explanation.

I'll test it out and add examples / tests

n-kall:
Your changes look great! One thing I noticed though is that the lprior tag is not vectorized as I would expect. 

For example, if I specify the tag on `class = "b"` and no individual coefficients, it works as intended. But if I then specify one specific coefficient to have a different tag, the tag for the class is not used anymore. See below:
Do you think it is reasonable for the tag to `(vectorize)` like the `prior` itself?

```r
prior1 <- c(
  prior(normal(0, 5), class = "b", lprior = "covariates") # vectorizes
)
prior2 <- c(
  prior(normal(0, 5), class = "b", lprior = "covariates"), # is removed
  prior(normal(0, 1), coef = "N1", lprior = "treatment") # works  
)
x <- brm(yield ~ N + K + P, data = npk,
         prior = prior1, backend = "cmdstanr",
         empty = TRUE
         )
#> prior_summary(x)
                 prior     class coef group resp dpar nlpar lb ub     lprior       source
            normal(0, 5)         b                                  covariates         user
            normal(0, 5)         b   K1                                        (vectorized)
            normal(0, 5)         b   N1                                        (vectorized)
            normal(0, 5)         b   P1                                        (vectorized)
 student_t(3, 55.6, 6.2) Intercept                                                  default
    student_t(3, 0, 6.2)     sigma                             0                    default
```


```r
#> stancode(x)
```

```stan
// generated with brms 2.22.8
functions {
}
data {
  int<lower=1> N;  // total number of observations
  vector[N] Y;  // response variable
  int<lower=1> K;  // number of population-level effects
  matrix[N, K] X;  // population-level design matrix
  int<lower=1> Kc;  // number of population-level effects after centering
  int prior_only;  // should the likelihood be ignored?
}
transformed data {
  matrix[N, Kc] Xc;  // centered version of X without an intercept
  vector[Kc] means_X;  // column means of X before centering
  for (i in 2:K) {
    means_X[i - 1] = mean(X[, i]);
    Xc[, i - 1] = X[, i] - means_X[i - 1];
  }
}
parameters {
  vector[Kc] b;  // regression coefficients
  real Intercept;  // temporary intercept for centered predictors
  real<lower=0> sigma;  // dispersion parameter
}
transformed parameters {
  // prior contributions to the log posterior
  real lprior_covariates = 0;
  real lprior = 0;
  lprior_covariates += normal_lpdf(b | 0, 5); // covariates tag works as intended
  lprior += student_t_lpdf(Intercept | 3, 55.6, 6.2);
  lprior += student_t_lpdf(sigma | 3, 0, 6.2)
    - 1 * student_t_lccdf(0 | 3, 0, 6.2);
  lprior += lprior_covariates;
}
model {
  // likelihood including constants
  if (!prior_only) {
    target += normal_id_glm_lpdf(Y | Xc, Intercept, b, sigma);
  }
  // priors including constants
  target += lprior;
}
generated quantities {
  // actual population-level intercept
  real b_Intercept = Intercept - dot_product(means_X, b);
}
```

```r
x2 <- brm(yield ~ N + K + P, data = npk,
         prior = prior2, backend = "cmdstanr",
         empty = TRUE
         )
#> prior_summary(x2)
                   prior     class coef group resp dpar nlpar lb ub     lprior       source
            normal(0, 5)         b                                  covariates         user
            normal(0, 5)         b   K1                                        (vectorized)
            normal(0, 1)         b   N1                              treatment         user
            normal(0, 5)         b   P1                                        (vectorized)
 student_t(3, 55.6, 6.2) Intercept                                                  default
    student_t(3, 0, 6.2)     sigma                             0                    default
```

```r
stancode(x2)
```

```stan
// generated with brms 2.22.8
functions {
}
data {
  int<lower=1> N;  // total number of observations
  vector[N] Y;  // response variable
  int<lower=1> K;  // number of population-level effects
  matrix[N, K] X;  // population-level design matrix
  int<lower=1> Kc;  // number of population-level effects after centering
  int prior_only;  // should the likelihood be ignored?
}
transformed data {
  matrix[N, Kc] Xc;  // centered version of X without an intercept
  vector[Kc] means_X;  // column means of X before centering
  for (i in 2:K) {
    means_X[i - 1] = mean(X[, i]);
    Xc[, i - 1] = X[, i] - means_X[i - 1];
  }
}
parameters {
  vector[Kc] b;  // regression coefficients
  real Intercept;  // temporary intercept for centered predictors
  real<lower=0> sigma;  // dispersion parameter
}
transformed parameters {
  // prior contributions to the log posterior
  real lprior_covariates = 0;
  real lprior = 0;
  real lprior_treatment = 0;
  lprior_treatment += normal_lpdf(b[1] | 0, 1); // treatment tag works
  lprior += normal_lpdf(b[2] | 0, 5); // covariate tag no longer holds the other coefficients
  lprior += normal_lpdf(b[3] | 0, 5);
  lprior += student_t_lpdf(Intercept | 3, 55.6, 6.2);
  lprior += student_t_lpdf(sigma | 3, 0, 6.2)
    - 1 * student_t_lccdf(0 | 3, 0, 6.2);
  lprior += lprior_covariates;
  lprior += lprior_treatment;
}
model {
  // likelihood including constants
  if (!prior_only) {
    target += normal_id_glm_lpdf(Y | Xc, Intercept, b, sigma);
  }
  // priors including constants
  target += lprior;
}
generated quantities {
  // actual population-level intercept
  real b_Intercept = Intercept - dot_product(means_X, b);
}
```

paul-buerkner:
Good catch. Fixed. Will you make the other adjustments then?

n-kall:
Yes, can do! Thank you again!

n-kall:
I now changed the name to "tag" and made it such that `prior_tag` can be specified in priorsense functions. There is still a bit of cleanup required, and documentation, but it seems to be working as intended now

```r
prior_cov <- prior(normal(0, 10), class = "b", tag = "covariates")
prior_trt <- prior(normal(0, 1), class = "b", coef = "Trt1", tag = "treatment")
prior_sigma <- prior(exponential(10), class = "sigma", tag = "sigma")

f <- brm(count ~ Trt + zAge + zBase + (1 | patient), data = epilepsy, prior = c(prior_cov, prior_trt, prior_sigma), backend = "cmdstanr")

priorsense::powerscale_sensitivity(f) # all priors

priorsense::powerscale_sensitivity(f, prior_tag = "treatment") # only treatment tag

priorsense::powerscale_sensitivity(f, prior_tag = "covariates") # only covariates tag

priorsense::powerscale_sensitivity(f, prior_tag = "sigma") # only sigma tag
```

paul-buerkner:
Very cool! with "cleanup required" do you mean in the new brms code or in priorsense?

n-kall:
I didn't actually change any code in priorsense, only in the brmsfit method for `create_priorsense_data`. Cleanup would include deciding if this way makes sense or if I should adjust priorsense to handle tags (currently there is `prior_selection` which is numeric)

paul-buerkner:
I think doing it in the brms method sounds sensible given that this is brms-specific stuff. Or what would be the benefit of doing it in priorsense?

Am I understanding it correctly, that you will also add some small tests and doc to the PR?

n-kall:
I was thinking that a user could define e.g. `lprior_treatment` and `lprior_covariates` as variables directly in the Stan model code, and then use the same `prior_tag` (or maybe the current `prior_selection`) argument to do this selection. It might make sense to have the same way to do things in priorsense regardless of whether the fit is brmsfit or stanfit.

And yes, I can add the tests and docs to finish the PR

paul-buerkner:
I see, makes sense. Thank you!

paul-buerkner:
I would love to merge this PR soon. Would you have some time for tests and docs perhaps in the next couple of days?

n-kall:
I'll work on this during the rest of this week. As mentioned I still need to finalise the interface on the priorsense side, but almost done there

paul-buerkner:
Should now be fixed.

n-kall:
I added tests for checking the prior tagging. It also took a bit more work to get all existing tests to pass (because the new column `tag` has to be added to the prior in order to do updating and restructuring, if it doesn't exist). 

I think this is ready for another review / merge if you don't see issues. I still need to do a release of priorsense with updates to the selection to be compatible, but that can be done after merging this PR to brms.

There is still a vignette coming, but that could either go in brms or priorsense (or both).

paul-buerkner:
thanks! Will add the review soon.

paul-buerkner:
Thank you. I edited the addition of tag to old priors. 

brms as a package is already quite large so I don't want to store any further example models. For this reason, please remove brmsfit_example7 and move the related tests to the local/ subfolder. You could add the tests in tests-models-5.R for example. Please also remember to revert the changes to sysdata.Rdata

n-kall:
Sounds good, I can make those changes

n-kall:
I think this should be ready now. I still had to regenerate the brmsfit examples to get some tests unrelated to this PR to pass

paul-buerkner:
Thanks! Can you say which test was setting version <- NULL?

n-kall:
[restructure tests, here](https://github.com/paul-buerkner/brms/blob/adf6a3bf7b733b3cf395d1f7980d1d55c2fc4832/tests/testthat/tests.restructure.R#L7)

paul-buerkner:
I see. Well it should not fail so rerunning the internal test models is not a longer lasting fix. I will investigate.

paul-buerkner:
Okay, all working and good also from my side now. Can I safely merge this or will the new brms code lead to a failure in the current cran version of priorsense? If so, when will the new priorsense version be on CRAN?

n-kall:
I don't think this change would break anything, but maybe be safe and wait until the end of the week as I should have then made a new CRAN release for priorsense

paul-buerkner:
Thanks! If it is now or end of this week, doesn't really matter. I think it should be safe. Merging now.

And thank you for working on this issue!

Issue #1728, opened 2025-01-21
creating a custom_family with a categorical-like response (e.g., Dirichlet-multinomial)


tom-peatman:
I've been attempting to create a `custom_family` for the Dirichlet-multinomial in `brms v2.22.0`, following the approach set out in the [custom family vignette](https://cran.r-project.org/web/packages/brms/vignettes/brms_customfamilies.html).

However, I'm running into difficulties. The stan code suggests that my attempted Dirichlet-multinomial family does not have a categorical-like response:

```{r}
library(brms)

# make custom family
dirichlet_multinomial <- custom_family(
  "dirichlet_multinomial",
  dpars = c("mu", "phi"),
  links = c("logit", "log"),
  lb = c(NA, 0),
  type = "int",
  loop = TRUE
)

# stan function for Dirichlet-multinomial lpmf
#  - mu is the logit probabilities
#  - phi is the sum of alphas of dirichlet distribution
stan_funs <- "
real dirichlet_multinomial_lpmf(array[] int y, vector mu, real phi) {
  // get Dirichlet alphas from logit probabilities
  int N = num_elements(mu);
  vector[N] alpha = phi * softmax(mu);

  // get trials from y
  real T = sum(y);

  real output = lgamma(phi) + lgamma(T + 1.0) - lgamma(T + phi) +
    sum(lgamma(to_vector(y) + alpha)) - sum(lgamma(alpha)) - sum(lgamma(to_vector(y) + 1));
  return output;
}
"

stan_vars <- stanvar(scode = stan_funs, block = "functions")

# simulate some data
N <- 15
dat <- data.frame(
    y1 = rbinom(N, 10, 0.3), y2 = rbinom(N, 10, 0.5), 
    y3 = rbinom(N, 10, 0.7), x = rnorm(N)
)
dat$size <- with(dat, y1 + y2 + y3)
dat$y <- with(dat, cbind(y1, y2, y3))

# inspect multinomial stan code
m_code <- make_stancode(bf(y ~ x), data = dat, family = multinomial)
m_code

# inspect Dirchlet-multinomial stan code
dm_code <- make_stancode(bf(y ~ x), data = dat, family = dirichlet_multinomial, stanvars = stan_vars)
dm_code

```

[...]

PR #1729, opened 2025-01-24
Create native Dirichlet multinomial family

tom-peatman:
Following discussion in #1728, I've had a go at creating a native Dirichlet multinomial family.

I've attempted to update the documentation of `brmsfamily`, and added some tests too.

I'm not confident that the [`posterior_epred.R`](https://github.com/paul-buerkner/brms/blob/master/R/posterior_epred.R) function is correct, and probably should not have included it.

Hopefully this is helpful - please let me know if you need anything more from me. And thanks again @paul-buerkner for all your help with this!


paul-buerkner:
Thank you! This PR looks good overall, I think.

Why are not confident about posterior_epred? You should be able to test it by comparing its output against the known(?) mean of the dirichlet_mulinomial distribution. I currently don't have time to check this detail myself.

R/distributions.R
```diff
@@ -2185,6 +2185,18 @@ dmultinomial <- function(x, eta, log = FALSE) {
   out
 }
 
+# density of the multinomial distribution with the softmax transform
+# @param x positive integers not greater than ncat
+# @param eta the linear predictor (of length or ncol ncat)
+# @param phi the dispersion parameter
+# @param log return values on the log scale?
+ddirichletmultinomial <- function(x, eta, phi, log = FALSE) {
```
paul-buerkner, line 2193: does this accept the same kinds of inputs as **brms::dmultinomial** (just with an additional phi?). E.g. is the right dimensionality for x and eta taken care of?

R/log_lik.R
```diff
@@ -837,6 +837,17 @@ log_lik_multinomial <- function(i, prep) {
   log_lik_weight(out, i = i, prep = prep)
 }
 
+log_lik_dirichlet_multinomial <- function(i, prep) {
+  stopifnot(prep$family$link == "logit")
+  eta <- get_Mu(prep, i = i)
+  eta <- insert_refcat(eta, refcat = prep$refcat)
+  phi <- get_dpar(prep, "phi", i = i)
+  cats <- seq_len(prep$data$ncat)
```
paul-buerkner, line 845: the variables cats and alpha seem redundant here. Leftovers from previous versions?

R/posterior_epred.R
```diff
@@ -587,6 +587,21 @@ posterior_epred_multinomial <- function(prep) {
   out
 }
 
+posterior_epred_dirichlet_multinomial <- function(prep) {
+  get_counts <- function(i) {
+    eta <- insert_refcat(slice_col(eta, i), refcat = prep$refcat)
+    dcategorical(cats, eta = eta) * trials[i]
+  }
+  # dirichlet part included in mu
```
paul-buerkner, line 595: I don't understand this comment

R/posterior_epred.R
```diff
@@ -587,6 +587,21 @@ posterior_epred_multinomial <- function(prep) {
   out
 }
 
+posterior_epred_dirichlet_multinomial <- function(prep) {
```
paul-buerkner, line 590: Since this is now the same as the one for multinomial, perhaps it makes sense to just call it directly with a comment saying why this is valid?

paul-buerkner (changes requested):
I have now checked the PR once again in a bit more detail and found a few more minor things to adjust before the PR is ready to be merged.

R/distributions.R
tom-peatman: That is correct (or at least, that's my intention). Perhaps there should be checks on the structure of eta (i.e., those implemented in dmultinomial)?

R/log_lik.R
tom-peatman: Correct - I initially constructed `log_lik_dirichlet_multinomial` using `log_lik_dirichlet` as a template, and forgot to remove `cats` and `alpha`. Thanks for picking this up!

R/posterior_epred.R
tom-peatman: Agree that it's confusing and should be removed (and see response to your next comment).

As an aside, there's a similar comment in `posterior_epred_beta_binomial` which I took inspiration from (`# beta part included in mu`), though it wasn't particularly clear to me what it meant.

R/posterior_epred.R
tom-peatman: Agreed - and have implemented this approach. Hopefully the comment is sufficiently clear.

R/distributions.R
paul-buerkner: I think that would be preferable to make sure that the two families work with the same input.

tom-peatman:
Thanks @paul-buerkner for reviewing my pull request, and for your helpful comments and suggestions. I've made some minor updates in the latest commit, which hopefully address your questions.

R/posterior_epred.R
paul-buerkner: I see. thanks.

paul-buerkner:
Thanks! Can you add the initial transformations and check also do the density function as discussed above? Then the PR should be ready for merging.

tom-peatman:
I'll implement the checks in the density function as requested above.

Though I'm not sure what you mean by initial transformations?

Do you also want me to add a specific help page for the dirichlet multinomial describing the parameterisation of the distribution (similar to `?BetaBinomial`)? I just realised that I didn't do this first time around

R/distributions.R
tom-peatman: Thanks for the quick response! Checks on eta now added

paul-buerkner:
I meant 

```R
if (is.null(dim(eta))) {
    eta <- matrix(eta, nrow = 1)
  }
```

but you already added that. I think it is fine if we keep the density internal for now. Just as we have it with dmultinomial. I will check the PR once again and then merge.

tom-peatman:
OK great - thanks again for all your help with this!

paul-buerkner:
Will merge now. Thank you for your contribution to brms! I highly appreciate it!

tom-peatman:
Thanks @paul-buerkner, it was a pleasure to contribute and the least I can do as a very grateful user of brms!

PR #1735, opened 2025-02-01
Align step() with Stan and add int_step()

### LINKED ISSUE #1734: Mismatch of `step(0)` value between Stan and `brms`
Hi @paul-buerkner ,

thanks for the great `brms` package. I was working on a model where I needed a changepoint in the linear predictor at time 0, and it took me a few days to figure out why things did not work as expected. Besides me using the Stan function `step()` wrongly, there is a mismatch of `step(0)` giving 1 in Stan (surprisingly, but documented) and 0 in `brms` (as I would have expected Stan to work too) which contributed to my confusion. In a nutshell: 

- Problem:
  - `step(0) = 1` in Stan (see the warning in the [doc](https://mc-stan.org/docs/functions-reference/integer-valued_basic_functions.html#absolute-functions))
  - `step(0) = 0` in `brms` as reimplemented in the R function [here](https://github.com/paul-buerkner/brms/blob/master/R/numeric-helpers.R#L67)
- Impact:
  - If you have a model that is of the form `y ~ step(time > 0) * foo(time) + step(time <= 0) * bar(time)`, then, while you should write a model `y ~ int_step(time > 0) * foo(time) + int_step(time <= 0) * bar(time)`...
  - Stan will interpret your model essentially as `y ~ foo(time) + bar(time)` because the boolean in the `step()` will be true or false, coerced to 1 or 0, but both will evaluate with the `step()` call to 1
  - `brms` will interpret your model as `y ~ int_step(time > 0) * foo(time) + int_step(time <= 0) * bar(time)`
  - Therefore the model fit, e.g. generated by `posterior_predict()` etc. will be wrong, i.e. not match what Stan is actually evaluating as linear predictor
- Proposed solution:
  - change the `step()` definition in `brms` to match the Stan evaluation, i.e. use `ifelse(x >= 0, 1, 0)` instead of current `ifelse(x > 0, 1, 0)`
  - add the `int_step()` R function to `brms` 
  - Ideally, we can warn the user somehow, e.g. in `step()` we could consider a warning in case that `x` is logical, to point the user to the fact that they probably want to use `int_step()` in their model instead.
    - I just tested this via `fixInNamespace()` hack and it can work nicely:
      ````
      Warning messages:
      1: In step(year > 0) :
        you probably want to use int_step() instead of step(), because step(FALSE) = 1 in Stan!
      2: In step(year <= 0) :
        you probably want to use int_step() instead of step(), because step(FALSE) = 1 in Stan!
      ```` 

If that sounds good, I am happy to prepare a PR- what do you think?

Cheers,
Daniel

danielinteractive:
closes #1734

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,8 @@
 * Fit extended-support Beta models via family `xbeta` 
 thanks to Ioannis Kosmidis. (#1698)
 * Add a `seed` argument to `loo_R2` thanks to Marco Colombo. (#1713)
+* Add the `int_step` R function to match the corresponding Stan 
+function. (#1734)
 
 ### Bug Fixes
 
@@ -15,6 +17,8 @@ in parallel even on Unix systems. To avoid potential speed loss for small
 models, `log_lik` will not use `option(mc.cores)` anymore.
 These changes may be reverted once the underlying causes of this
 issue have been fixed. (#1658)
+* Align the definition of the R function `step()` with the definition in Stan,
+such that `step(0) == 1` (#1734)
 
 ### Other Changes
 
--- a/R/numeric-helpers.R
+++ b/R/numeric-helpers.R
@@ -66,9 +66,9 @@ log1m <- function(x) {
 
 step <- function(x) {
   if (is.logical(x)) {
-    warning(
+    warning2(
       "Consider using int_step() instead of step(), ",
-      "because step(FALSE) = step(TRUE) = 1 in Stan"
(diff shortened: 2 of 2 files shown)


danielinteractive (commented):
Hi @paul-buerkner, here is the PR, please let me know if and where I should add tests, I was not sure about that. Thanks!

paul-buerkner:
Thank you! That's perfect :-)

Issue #1739, opened 2025-02-05
[bug] Incorrect predictions from approximate GP terms


nicholasjclark:
Apologies if this is not the correct forum for this issue. I am working with approximate GPs for time series modelling, and I'm interested in extrapolating into the future for some of these models. However I find that the predictions tend to be quite inconsistent and dependent on the overlap of the prediction data with the original training data. To illustrate, here is a small reprex:

```r
library(brms) # version 2.22.9
library(ggplot2); theme_set(theme_classic())

# Simulate some Poisson observations of a latent GP
N <- 70
set.seed(0)

mean_func = function(x){
  0.5 + 0.02 * x + sin(0.25 * x) - cos(-0.05 * x) + 
    rep(sin(1:4), 20)[1:length(x)]
}

dat <- data.frame(time = 1:N,
                  y = rpois(N, exp(mean_func(1:N))))

#### Approximate GP model ####
gp_approx <- brm(y ~ gp(time, 
                        c = 5/4, 
                        k = 30,
                        scale = FALSE),
                 data = dat,
                 family = poisson(),
                 backend = 'cmdstanr',
                 chains = 4,
                 cores = 4,
                 silent = 2)

# Predictions: why does the prediction of the second peak keep 
# shifting as we extrapolate further?
patchwork::wrap_plots(
  
# predict for the training period; happy days
  plot(conditional_effects(gp_approx,
                           int_conditions = list(time = 1:75),
                           method = 'posterior_linpred'),
       plot = FALSE)[[1]] +
    geom_point(data = data.frame(truth = mean_func(1:N),
                                 time = 1:N,
                                 lower__ = 0,
                                 upper__ = 0), aes(time, truth)) +
    scale_x_continuous(limits = c(0, 110)) +
    geom_vline(xintercept = 70),
  
  # predict for the training period plus another 20 time pionts;
  # looks a bit odd
  plot(conditional_effects(gp_approx,
                           int_conditions = list(time = 1:95),
                           method = 'posterior_linpred'),
       plot = FALSE)[[1]] +
    geom_point(data = data.frame(truth = mean_func(1:N),
                                 time = 1:N,
                                 lower__ = 0,
                                 upper__ = 0), aes(time, truth)) +
    scale_x_continuous(limits = c(0, 110)) +
    geom_vline(xintercept = 70),
  
  # repeat as above but extrapolate even further; 
  # now looks very odd (seems to be "stretching" the original function)
  plot(conditional_effects(gp_approx,
[...]

PR #1740, opened 2025-02-06
store original bounday factor L for approximate GPs

nicholasjclark:
Storing and re-using the originally computed boundary factor `L` will ensure that predictions for new covariate values work appropriately when predicting from approximate GPs

### DIFF
--- a/R/data-predictor.R
+++ b/R/data-predictor.R
@@ -642,19 +642,24 @@ data_gp <- function(bframe, data, internal = FALSE, ...) {
     out[[paste0("cmeans", sfx)]] <- cmeans
     # required to compute inverse-gamma priors for length-scales
     out[[paste0("Xgp_prior", sfx)]] <- Xgp
-    # required to compute eigenfunctions of approximate GPs with new data
-    out[[paste0("Lgp", sfx)]] <- choose_L(Xgp, c = c)
   }
   if (!isNA(k)) {
     # basis function approach requires centered variables
     Xgp <- sweep(Xgp, 2, cmeans)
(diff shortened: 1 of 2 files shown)


R/data-predictor.R
```diff
@@ -642,12 +642,19 @@ data_gp <- function(bframe, data, internal = FALSE, ...) {
     out[[paste0("cmeans", sfx)]] <- cmeans
     # required to compute inverse-gamma priors for length-scales
     out[[paste0("Xgp_prior", sfx)]] <- Xgp
+    # required to compute eigenfunctions of approximate GPs with new data
```
paul-buerkner, line 645: this is a bit in an awkward spot since it may now also run for non-exact GPs. would it make sense to store L after it has been regularily computed a few lines later (those you also changed). And wrapped in an if (internal) statement there?

paul-buerkner (changes requested):
Thank you for finding this issue and even fixing it youtself! I have just one minor thing that I think needs to be changed before merging. Would you mind checking?

paul-buerkner:
fixes #1739

R/data-predictor.R
nicholasjclark: Ah yes, that makes sense now. I've rearranged that section as you recommended. Thanks for the quick review

paul-buerkner: perfect!

paul-buerkner:
This PR looks good now. Merging :-)

nicholasjclark:
> This PR looks good now. Merging :-)

Excellent thanks very much for the amazing work @paul-buerkner

Issue #1719, opened 2024-12-15
Feature request: Allow weights to also be specified at group level


bschneidr:
# Summary

This request is to allow the user to specify not only unit-level weights in a brms formula, but group-level weights as well. This would extend the current syntax from this:

```r
# Specify a formula with unit-level weights
y | weights(w) ~ x + (1 | g)
```

to maybe something like this

```r
# Specify a formula with group-level weights as well
y | weights(w) ~ x + ( 1 | g | group_weights(w_g) )
```

where $w$ is a unit-level weight variable with potentially $n$ distinct values, and $w_g$ is a group-level weight variable with one value per group. 

I don't know what the best syntax for this functionality would be, but I hope this illustrates the idea well enough.

The motivation for this is to make `brms` able to fit survey-weighted multilevel models: [Savitsky and Williams (2022)](https://journals.sagepub.com/doi/abs/10.2478/jos-2022-0039) showed that only weighting the unit-level likelihood doesn't produce reasonable estimates for multilevel models, but adding group-level weights to the random effects' priors works well.

I think would be a valuable addition to `brms` for many folks who want to analyze data from health and social surveys. I would be happy to work on implementing this if you are open to it and have a preferred way for how this would look.

# Example Data

Below is an example dataset of size $n=6$ with $M=3$ groups. The variable `w` is a unit level weight, and the variable `w_g` is a group-level weight.

```r
example_data <- data.frame(
  y   = rnorm(n = 6),
  x   = rnorm(n = 6),
  w   = runif(n = 6, min = 0.9, max = 1),
  g   = c('a', 'a',  'b',  'b',  'c',  'c'),
  w_g = c(0.5, 0.5, 0.75, 0.75, 0.25, 0.25)
)
```

We want to estimate a simple varying-intercepts model of the form `y ~ x + (1 | g)`. 

Below is Stan code that would be generated for this model:

```r
brms::make_stancode(
    data    = example_data,
    formula = y | weights(w) ~ x + (1 | g),
    family  = gaussian()
)
```

<details>
<summary>Click to show generated Stan code</summary>

```stan
// generated with brms 2.22.0
functions {
}
data {
  int<lower=1> N;  // total number of observations
  vector[N] Y;  // response variable
  vector<lower=0>[N] weights;  // model weights
  int<lower=1> K;  // number of population-level effects
  matrix[N, K] X;  // population-level design matrix
  int<lower=1> Kc;  // number of population-level effects after centering
[...]

PR #1748, opened 2025-02-25
Fix PW stan code.

bschneidr:
# Summary
Hi Paul,

I'm sorry to bother you with more on this prior-weights functionality from #1719. In applying this functionality to a different dataset from the 'survey' package, I realized there was a problem in the Stan code I wrote for this, linked here:

https://github.com/paul-buerkner/brms/blob/c741df7b008a9fccfff1b4d04a8102e0e8e6c1d7/R/stan-predictor.R#L641
https://github.com/paul-buerkner/brms/blob/c741df7b008a9fccfff1b4d04a8102e0e8e6c1d7/R/stan-predictor.R#L739

The problem is that `std_normal_lpdf()` function in Stan evaluates to a scalar, not a vector. So multiplying its output by the weights vector is incorrect. Instead, the code should have specified a loop over each group, adding the weights and the likelihood contribution from each group separately. 

This small PR fixes that issue by changing the Stan code to do this loop, and it also updates the unit tests accordingly.

# Reproducible Example to Illustrate the Problem

To illustrate the problem, I fit a model where the group-level weights are all equal to 1. The resulting estimates should be equivalent to fitting an unweighted model. But from the reprex below, we can see that we get very different estimates for the variance component `sd(Intercept)` depending on whether we specify group-level weights.

```r
remotes::install_github("paul-buerkner/brms@c741df7", upgrade=FALSE, force=TRUE)
library(brms)
library(survey)
data(api, package = 'survey')

wtd_fit <- brm(
  formula = api00 ~ ell + (1 | gr(dnum, pw = dwgt)),
  data    = apiclus1 |> transform(dwgt = 1), 
  family  = gaussian(),
  seed    = 1999,
  iter    = 5000,
  backend = "cmdstanr"
)

unwtd_fit <- brm(
  formula = api00 ~ ell + (1 | gr(dnum)),
  data    = apiclus1, 
  family  = gaussian(),
  seed    = 1999,
  iter    = 5000,
  backend = "cmdstanr"
)
summary(wtd_fit)
summary(unwtd_fit)
```

From the weighted fit, we have:
```r
#> Multilevel Hyperparameters:
#> ~dnum (Number of levels: 15) 
#>               Estimate Est.Error l-95% CI u-95% CI Rhat Bulk_ESS Tail_ESS
#> sd(Intercept)   259.66     55.61   172.90   387.21 1.00     2450     3840
```

From the unweighted fit, we have:
```r
#> Multilevel Hyperparameters:
#> ~dnum (Number of levels: 15) 
#>               Estimate Est.Error l-95% CI u-95% CI Rhat Bulk_ESS Tail_ESS
#> sd(Intercept)    73.23     16.40    47.62   112.38 1.00     2605     3988
```

After this pull request, we have output for the weighted fit that is equivalent to the unweighted fit:

```r
#> Multilevel Hyperparameters:
#> ~dnum (Number of levels: 15) 
#>               Estimate Est.Error l-95% CI u-95% CI Rhat Bulk_ESS Tail_ESS
#> sd(Intercept)    73.64     16.69    48.29   111.67 1.00     2168     4060
```

### DIFF
--- a/R/stan-predictor.R
+++ b/R/stan-predictor.R
@@ -638,8 +638,8 @@ stan_re <- function(bframe, prior, normalize, ...) {
     )
     if (has_pw) {
       str_add(out$model_prior) <- glue(
-        "  for (n in 1:N_{id}) {{\n",
-        "    target += PW_{id}[n] * std_normal_{lpdf}(to_vector(z_{id}[,n]));\n",
+        "  for (j in 1:N_{id}) {{\n",
+        "    target += PW_{id}[j] * std_normal_{lpdf}(z_{id}[, j]);\n",
         "  }\n"
       )
     } else {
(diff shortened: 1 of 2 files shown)


bschneidr:
I tested it out with a couple datasets and various model options, and it's working as expected. I think it's good to merge. Thanks again Paul!

paul-buerkner:
Great! Merging now.

PR #1757, opened 2025-03-20
allow newdata to pass to log_lik in create_priorsense_data

### LINKED ISSUE #1752: create_priorsense_data.brmsfit should handle `newdata`
Issue n-kall/priorsense/issues/84 shows that `newdata`  (which is needed when the model is using mi for missing data) doesn't work currently with priorsense. I think this is because it is not being passed to the `log_lik` function in `create_priorsense_data`.

I suppose the fix would be to pass `...` through to `log_lik()`, which seems to be the way it works for `loo.brmsfit`. @paul-buerkner do you think this is the right approach?

n-kall:
Fixes #1752 and does not seem to cause any issues.

### DIFF
--- a/R/priorsense.R
+++ b/R/priorsense.R
@@ -8,7 +8,7 @@
 #' functions, so you will rarely need to call it manually yourself.
 #'
 #' @param x A \code{brmsfit} object.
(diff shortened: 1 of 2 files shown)


paul-buerkner:
Thank you!

PR #1759, opened 2025-03-28
Keep track of chains in `log_lik_draws`

n-kall:
The `log_lik_draws.brmsfit` function which is used by priorsense was merging chains, which made it problematic to use if one wanted to bind the `log_lik` draws to the posterior `draws` object. This PR adds chain information to the returned array. I also cleaned up some uses of `posterior::` that were not needed as the functions were already imported in brms.


R/priorsense.R
```diff
@@ -51,13 +51,19 @@ create_priorsense_data.brmsfit <- function(x, ...) {
 
 #' @exportS3Method priorsense::log_lik_draws
 log_lik_draws.brmsfit <- function(x, ...) {
+  nchains <- nchains(x)
+  niters <- niterations(x)
   log_lik <- log_lik(x, ...)
-  log_lik <- posterior::as_draws_array(log_lik)
+  nobs <- length(log_lik) / (nchains * niters)
```
paul-buerkner, line 57: This is dangerous if `...` already subsets draws for some reason. So you need to keep track of `ndraws` and `draw_ids` arguments as passed to `log_lik` and only store chains information if these arguments were not specified.

paul-buerkner, line 57: `nobs` is always the number of columns of `log_lik` so no need to infer that from length, nchains, and niters.

paul-buerkner (changes requested):

R/priorsense.R
n-kall: Of course! That's much simpler

n-kall: Very good point. I'll revise this, and update the PR

n-kall: I now changed it to check if `nrow(log_lik(x, ...)) < niterations(x) * nchains(x)` and if so, then just set `nchains <- 1` and `niters <- nrow(log_lik(x, ...))`. I think this is simpler than checking for given arguments in `...`, which I tried in the previous commit.

n-kall:
@paul-buerkner I think this is now ready for another look

paul-buerkner (approved):
Looks good now. approved :-)

paul-buerkner:
Checks are passing locally. Merging this PR now.

PR #1760, opened 2025-04-01
compress plots in vignettes

wds15:
I noticed that the brms package size is quite large, which is due to the rendered vignettes to a large extent. The proposed changes implement best practices I have come to learn on the net. The trick implemented uses the pngquant engine to compress the plots substantially. For example, the non-linear vignette has a size of 580 kb with the released version of brms while the version rendered with the changes on this PR reduces this to just 164 kb.

As I lack some dependencies to build the entire package, I can't report the total reduction of the final package (currently about 4MB), but I'd expect that the size will go down quite a bit. So please, do make a check to build the package and compare the gains before merging.

Given on how many servers the R package is replicated to, I'd guess that the size savings are worthwhile. However note that this makes `pngquant` a dependency. This is available from homebrew for macOS - and I guess there are various other sources for it for the different OSes, etc..

### DIFF
--- a/vignettes/brms_missings.Rmd
+++ b/vignettes/brms_missings.Rmd
@@ -17,17 +17,20 @@ params:
 ```{r, SETTINGS-knitr, include=FALSE}
 stopifnot(require(knitr))
 options(width = 90)
+knitr::knit_hooks$set(pngquant = knitr::hook_pngquant)
 opts_chunk$set(
   comment = NA,
   message = FALSE,
   warning = FALSE,
   eval = if (isTRUE(exists("params"))) params$EVAL else FALSE,
-  dev = "jpeg",
-  dpi = 100,
+  dev = "ragg_png",
+  dpi = 72,
+  fig.retina = 1.5,
   fig.asp = 0.8,
   fig.width = 5,
   out.width = "60%",
-  fig.align = "center"
+  fig.align = "center",
+  pngquant = "--speed=1 --quality=50"
 )
 library(brms)
 ggplot2::theme_set(theme_default())
--- a/vignettes/brms_missings.Rmd
+++ b/vignettes/brms_missings.Rmd
@@ -17,7 +17,7 @@ params:
 ```{r, SETTINGS-knitr, include=FALSE}
 stopifnot(require(knitr))
 options(width = 90)
-knitr::knit_hooks$set(pngquant = knitr::hook_pngquant)
+knit_hooks$set(pngquant = knitr::hook_pngquant)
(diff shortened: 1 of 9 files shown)


paul-buerkner:
Thank you! Can you forsee any issues that CRAN might have with pngquant?

vignettes/brms_missings.Rmd
```diff
@@ -17,17 +17,20 @@ params:
 ```{r, SETTINGS-knitr, include=FALSE}
 stopifnot(require(knitr))
 options(width = 90)
+knitr::knit_hooks$set(pngquant = knitr::hook_pngquant)
```
paul-buerkner, line 20: Since we have loaded knitr at this point already, shall we drop the ´knitr::` here? At least I don't call `knitr::` for `opts_chunk` so its a bit inconsistent.

wds15:
I am using the exact same approach for `RBesT` and `OncoBayes2`, which I am maintaining on CRAN. I picked this approach up from here: https://nanx.me/blog/post/rpkgs-pngquant-ragg/

So no, CRAN is happy with it.

paul-buerkner (changes requested):
Looks good. I only have a minor suggestions for changing something.

Building the package right now to see how much the size is reduced overall

vignettes/brms_missings.Rmd
wds15: It's a style question of what to do here. I don't mind prefixing knitr to the call here, but I am happy to drop them, sure.

paul-buerkner:
Package building looks good, thank you! Merging now.

wds15:
Great. What is the package size now?

paul-buerkner:
Still around 4MB for some reason but the vignettes are clearly smaller now.

wds15 ***@***.***> schrieb am Di., 1. Apr. 2025, 15:32:

wds15:
Weird - at least the installed size of the package will go down somewhat.

If you like to substantially reduce the package size then another way is to not package most vignettes with brms, but instead keep these as articles. For `RBesT` I have a few articles on the pkgdown generated homepage: https://opensource.nibr.com/RBesT/

The package on CRAN only contains the introductory vignette. The implication is then that the articles are only available through the package homepage and are not distributed to the user. If that is of interest, I can file another PR.

paul-buerkner:
Perhaps in the future, thank you! Right now, I think I am sufficiently happy with the current size.

wds15:
ok... maybe I take another look and attempt to compress what is currently there better.

Issue #607, opened 2019-02-22
Conway-Maxwell-Poisson distribution


paul-buerkner:
The Conway-Maxwell-Poisson distribution (https://en.wikipedia.org/wiki/Conway%E2%80%93Maxwell%E2%80%93Poisson_distribution) is a generalization of the Poisson distribution that allows for both over- and underdispersion. There have been some attempts to get this distribution running in Stan, but I haven't seen one that works fast enough for most practical applications. Once we got such an implementation though, I think the distribution would make a very valuable addition to brms.

paul-buerkner:
An *experimental* version of the COM Poisson distribution is now available on GitHub. It is not yet documented or offically supported but can already be accessed via `brmsfamily("com_poisson")`. 

 I currently use the mode parameterization as proposed by Guikema and Goffelt (https://onlinelibrary.wiley.com/doi/abs/10.1111/j.1539-6924.2008.01014.x) with a (hopefully) accurate approxiation of the normalizing constant via truncation. The implementation is not terribly efficient (read: an order of magnitute or more slower than the poisson distribution), but it seems reasonable stable at least.

Ideally, I would like parameterize the mean, but I haven't found a way, yet, to get this parameterization working in a Bayesian framework.

In any case, I would love if users could try out the distribution and provide feedback.

bgoodri:
In some situations, it would make sense to utilize the results in https://arxiv.org/abs/1612.06618 to approximate the normalizing constant well-enough.

StaffanBetner:
An other option could be to implement Consul's generalized Poisson distribution, which could be parameterized by mean and variance.

paul-buerkner:
Thanks @bgoodri, just implemented the approximation.

@StaffanBetner Do you have a reference for me about the 
Consul's generalized Poisson distribution?

StaffanBetner:
@paul-buerkner The two main references are Consul & Jain (http://dx.doi.org/10.1080/00401706.1973.10489112) and Consul & Shoukri (https://doi.org/10.1080/03610918508812463) (discussing some truncation problems).

StaffanBetner:
Huang (https://doi.org/10.1177/1471082X17697749) would probably be a complement to the pre-print suggested by @bgoodri.

paul-buerkner:
[...]

PR #1765, opened 2025-04-29
Improved implementation of Conway-Maxwell-Poisson distribution

wellington36:
For various parameter intervals, the calculation of the normalization constant of the Conway-Maxwell-Poisson is shown to be imprecise (related to #607), both in relation to the _true_ value (in this case, I consider an evaluation of the series with at least 1 million terms) and compared to other libraries:

image
_Table generated by https://github.com/wellington36/test_comp_brms_
 
However, this issue can be corrected, and the series (of the normalization constant) can be evaluated with a guaranteed error as small as desired.
This particular series passes the ratio test (with limit $L = 0$), and since the ratio of consecutive terms is monotonic, the following holds:

image

where \(S\) is the true value, \(S_n\) is the sum of the terms up to \(n\), and \(a_n\) is the n-th term of the series.
Based on this, a new version of the function `log_Z_com_poisson` for evaluating the normalization constant was developed using this result as a stopping criterion.
Additionally, some bugs were corrected, particularly in the Gaunt approximation function~[[1]](#2019gaunt), with contributions from @maxbiostat (my advisor).

The proof of the validity of this stopping criterion (and its algorithmization) is presented **[here (read time expected: ~12 minutes)](https://wellington36.github.io/Report_to_COMPoisson_brms.pdf)**, along with examples using the new version. These include an application with MCMC, based on our work~[[2]](#2025carvalho).

Here is the updated version of the table above:

image
_Table generated by https://github.com/wellington36/test_comp_brms_

And the corresponding table with errors instead of values:

image
_Table generated by https://github.com/wellington36/test_comp_brms_

Obs: Notice that the last two rows, where the error is not zero, fall under the Gaunt approximation~[[1]](#2019gaunt), which raises some discussion on whether the speed gain justifies the error.

Obs2: The repository version had a bug in the list index that was fixed so that the more demanding tests would be fair. The above test in particular did not fall into this bug.

### Referencies
  [1]<a id="2019gaunt"></a>: Gaunt, R. E., Iyengar, S., Daalhuis, A. B. O., and Simsek, B. (2019). An asymptotic expansion
for the normalizing constant of the conway–maxwell–poisson distribution. Annals of the Institute of Statistical Mathematics, 71(1):163–180.

  [2]<a id="2025carvalho"></a>: Carvalho, Luiz Max and Silva, Wellington J and Moreira, Guido A. (2022). Adaptive truncation of infinite sums: applications to Statistics. __arXiv preprint [arXiv:2202.06121](https://arxiv.org/abs/2202.06121)__.


wellington36:
A simple test can be make first installing from my fork:

```R
devtools::install_github("wellington36/brms")
```

then run:

```R
library(brms)

zinb <- read.csv("https://paul-buerkner.github.io/data/fish.csv")

fit <- brm(count ~ 1,
                  cores=4,
                  iter=1000,
                 data = zinb,
                 family = "com_poisson",
                 backend = "cmdstanr",
                 control = list(adapt_delta = 0.99))

summary(fit)
plot(fit)
```

paul-buerkner:
Great! Thank you, for fixing the brms implementation!

Out of interest, how much faster is the evaluation now?

wellington36:
> Out of interest, how much faster is the evaluation now?

In the code, each normalisation constant is calculated with 64-bit machine precision (2^-52 ~ 10^-16 of precision in each constant). This ensures accurate constants, but makes the process slower. This precision value can be relaxed to increase speed, at the cost of losing some precision. There are other methods, such as evaluating in bucket, that can help in terms of speed.

I am implementing some test versions to see how some strategies affect speed and precision as branches in my fork.

inst/chunks/fun_com_poisson.stan
```diff
@@ -1,118 +1,139 @@
...
+// based on equations (4) and (31) of doi:10.1007/s10463-017-0629-6
+// Args: see log_Z_com_poisson()
+real log_Z_com_poisson_approx(real log_mu, real nu) {
+  real nu2 = nu^2;
+  real log_common = log(nu) + log_mu/nu;
+  array[4] real resids;
+  real ans;
+  real lcte = (nu * exp(log_mu/nu)) -
+    ( (nu-1)/(2*nu)* log_mu + (nu-1)/2*log(2*pi()) + 0.5 *log(nu));
+  real c_1 = (nu2-1)/24;
```
paul-buerkner, line 11: can you please add proper spacing everywhere?

inst/chunks/fun_com_poisson.stan
```diff
@@ -1,118 +1,139 @@
...
+  real c_2 = (nu2-1)/1152*(nu2 + 23);
+  real c_3 = (nu2-1)/414720* (5*square(nu2) - 298*nu2 + 11237);
+  resids[1] = 1;
```
paul-buerkner, line 14: The resids terms look like they could be made more numerically stable, via some log sum exp style code. Or is this done intentionally like that? See my old implementation for inspiration.

inst/chunks/fun_com_poisson.stan
```diff
@@ -1,118 +1,139 @@
...
-    real nu2 = nu^2;
-    // first 4 terms of the residual series
-    real log_sum_resid = log1p(
-      nu_mu^(-1) * (nu2 - 1) / 24 +
-      nu_mu^(-2) * (nu2 - 1) / 1152 * (nu2 + 23) +
-      nu_mu^(-3) * (nu2 - 1) / 414720 * (5 * nu2^2 - 298 * nu2 + 11237)
-    );
-    return nu_mu + log_sum_resid  -
-      ((log(2 * pi()) + log_mu) * (nu - 1) / 2 + log(nu) / 2);
+// log approximate normalizing constant of the COM poisson distribuion
```
paul-buerkner, line 1: could you add 2 space in front of everything such that, when auto added to the brms stan code, it has the proper indentation?

paul-buerkner (changes requested):
PR looks good overall! I just have some minor comments. Mostly formatting.

inst/chunks/fun_com_poisson.stan
wellington36: Done c560adcaa76a2d846dd0703decfcdfa2042f7a99

inst/chunks/fun_com_poisson.stan
wellington36: Done c560adcaa76a2d846dd0703decfcdfa2042f7a99

wellington36:
> Also, since this is a non-trival contribution to brms, I think you both should be added as contributors. You can add yourself to the description file at the bottom of the list of contributors as part of this PR.

Done 8f9b7f2fdae93be8339d8316eb797fc9f1b9b34c.

Additionally, in your version the stopping criterion is evaluated every 50 terms, I added this to gain a bit of speed. (995707a440643f1cdc1309cdf69ee3dd2f0d4991)

Reducing the precision is more effective, but I think it has to be done carefully.

inst/chunks/fun_com_poisson.stan
paul-buerkner: Let's keep it in for now. But I do see why log scale evaluation is more complicated. Let's leave it as is for now.

paul-buerkner:
From my side the PR is ready. Is there anything to you want to change before I merge?

Did you run a couple of brms models with the new COM poisson already to see how it works out it practice?

wellington36:
> From my side the PR is ready. Is there anything to you want to change before I merge?

On my side too. I can think of some improvements, but it's not urgent.

> Did you run a couple of brms models with the new COM poisson already to see how it works out it practice?

It works well. I did some tests with this dataset: https://paul-buerkner.github.io/data/fish.csv and in Section 5.1 of this paper some more robust tests were documented: https://arxiv.org/html/2202.06121v2#S5. One problem is that if you want a normalisation constant with a guarantee of error below the machine epsilon, the method ends up being slower than methods without the same guarantee.

paul-buerkner:
Do we want/need this quarantee? Can we relax it or let the user control it via an argument in the com_poisson family?

wellington36:
It is a very strong and simple guarantee. Basically a stopping criterion for the terms of the normalisation constant series, when a certain value, calculated by the bound_remainder function that receives the last two terms, is less than a given epsilon. Then the sum up to this guy has an error less than or equal to epsilon.

This epsilon can be passed by the user, by default I left epsilon as 2^(-52) ~ 10^(-16). If you think it is a very low value I can increase it, which would make the code faster.

paul-buerkner:
we can also defer the corresponding interface to another PR once we know
how much accuracy we usually need in practice

Paul Buerkner ***@***.***> schrieb am Fr., 2. Mai 2025, 14:55:

wellington36:
> how can the user pass this in brms?

At first, I don't know. It seems more reasonable to me to leave the minimum amount of work to the user, especially with internal choices.

> we can also defer the corresponding interface to another PR once we know
> how much accuracy we usually need in practice

I like the idea, maybe talk in the issue #607.  In my tests, a precision of $10^{-7}$ usually gives the same result as a precision of $10^{-16}$. I find it very difficult to know how this error will affect the final result.

wellington36:
In an ideal world. I would like it if during the MCMC we could identify points that need more precision and adjust the epsilon automatically.

paul-buerkner:
Can you check how faster is 10-7 vs. 10-16 in a simple example, say:

```R
fit1 <- brm(
  count ~ zBase * Trt,
  data = epilepsy, family = brmsfamily("com_poisson"),
  prior = prior(normal(0, 10), class = b)
)
```

paul-buerkner:
Can you check how faster is 10-7 vs. 10-16 in a simple example, say:

```R
fit1 <- brm(
  count ~ zBase * Trt,
  data = epilepsy, family = brmsfamily("com_poisson"),
  prior = prior(normal(0, 10), class = b)
)
```

And how much the posterior results will differ?

wellington36:
In my machine, the average of the chains was 225 seconds to $10^{-16}$ and 169 seconds to $10^{-7}$. The summary for $10^{-16}$:

image

And $10^{-7}$:

image

wellington36:
To test the gaunt's approximation, I run $10^{-7}$ without the approximation and the time was 168 seconds, with

image

I had left the approach following the original version, but I don't think it's worth it.

paul-buerkner:
Shall we set it to 10-7 then for now? Should suffice. If some experiments you may perform in the future suggest we need smaller episilon we can still set it smaller than.

wellington36:
> Shall we set it to 10-7 then for now?
Done in b1299991fc3d7f0f3cff7897c08c35707500676e

wellington36:
For me, it's ready.

PR #1777, opened 2025-05-14
add loo_pit_ecdf to aps_types

avehtari:
`ppc_loo_pit_ecdf` was recently added to `bayesplot`. This PR will add "loo_pit_ecdf" to aps_types, so that by default all posterior draws are used to make the plot. Otherwise, using github version of `bayesplot` `pp_check(fit, type="loo_pit_ecdf")` works already


paul-buerkner:
Thanks!

PR #1784, opened 2025-06-06
adds few new functions to let user define a brms_cache_folder Option …

SermetPekin:
Introduced a new global option, brms_cache_folder, allowing users to specify a default folder for saving and loading cached brmsfit objects.

Updated get_cache_folder() to read from getOption("brms_cache_folder") whenever the supplied file path has no directory component.

Modified split_folder_and_file(), check_brmsfit_file(), read_brmsfit(), and write_brmsfit() so that:

    If file is passed without a path, it is automatically placed into (and looked for in) the folder returned by get_cache_folder().

    The target cache directory is created (or validated) at write time, and checked at read time.

Ensured backward compatibility: existing workflows that pass a full path to read_brmsfit()/write_brmsfit() continue to function exactly as before.


paul-buerkner:
Thank you! would you mind reverting all the auto-formating changes so I could review the actual changes better?

SermetPekin:
> Thank you! would you mind reverting all the auto-formating changes so I could review the actual changes better?

Sure. Probably it would be easier to delete this one and start a new one.

PR #1790, opened 2025-06-23
adds few new functions to let user define a brms_cache_folder Option-…

SermetPekin:
(Revised version)
Summary of Changes:

- Introduced a new global option, brms.cache_folder, which allows users to define a default directory for saving and loading cached brmsfit objects.

- created get_cache_folder() function to read from getOption("brms.cache_folder") when the supplied file path does not include a directory component.

- Added  split_folder_and_file(), modified check_brmsfit_file(), read_brmsfit(), and write_brmsfit() to support the use of brms.cache_folder as the default folder when no explicit directory is provided.
- Changed return type to NULL instead of TRUE in check_folder function

### DIFF
--- a/tests/testthat/tests.brmsfit-helpers.R
+++ b/tests/testthat/tests.brmsfit-helpers.R
@@ -107,7 +107,8 @@ test_that("brmsfit_needs_refit works correctly", {
   fake_fit <- brm(y ~ x, data = data_model1, empty = TRUE)
 
   fake_fit_file <- fake_fit
-  fake_fit_file$file <- cache_tmp
+  # align windows with unix encoding of file paths
+  fake_fit_file$file <- gsub("\\", "/", cache_tmp, fixed = TRUE)
 
   scode_model1 <- make_stancode(y ~ x, data = data_model1)
   sdata_model1 <- make_standata(y ~ x, data = data_model1)
@@ -116,7 +117,6 @@ test_that("brmsfit_needs_refit works correctly", {
   data_model2$x[1] <- data_model2$x[1] + 1
   scode_model2 <- make_stancode(y ~ 0 + x, data = data_model2)
   sdata_model2 <- make_standata(y ~ 0 + x, data = data_model2)
-
 
   write_brmsfit(fake_fit, file = cache_tmp)
   cache_res <- read_brmsfit(file = cache_tmp)


wds15:
This looks nice, but does not seem to cover the case of cmdstanr backend fits. For this to work as well one would need to control the directory where stan models are store via the option `cmdstanr_write_stan_file_dir` (see here: https://mc-stan.org/cmdstanr/reference/write_stan_file.html).

Maybe you did not intend to include cmdstanr... but it would be really nice.

In case you like to include the cmdstanr backend binaries, I am happy to give more input as to how to make it safe for use in high-performance computing cluster environments.

SermetPekin:
> This looks nice, but does not seem to cover the case of cmdstanr backend fits. For this to work as well one would need to control the directory where stan models are store via the option `cmdstanr_write_stan_file_dir` (see here: https://mc-stan.org/cmdstanr/reference/write_stan_file.html).
> 
> Maybe you did not intend to include cmdstanr... but it would be really nice.
> 
> In case you like to include the cmdstanr backend binaries, I am happy to give more input as to how to make it safe for use in high-performance computing cluster environments.

Thanks for your comment! Just to clarify, my PR is only about controlling where brms saves its cached fit/results (via brms.cache_folder). It’s not meant to change where Stan model files or binaries are written (that’s what cmdstanr_write_stan_file_dir is for, and users can set that separately if needed).
I agree it’s an important distinction, since these folders often have different purposes—especially on clusters. My change doesn’t affect the backend or the Stan file location, just the brms cache; so users can still set both locations independently.

paul-buerkner:
The PR is now ready from my side. Is there anything you want to change before I merge?

SermetPekin:
Thank you very much. I really appreciate the acknowledgment and the kind note. Everything looks perfect from my side.

paul-buerkner:
One windows test fails (https://github.com/paul-buerkner/brms/actions/runs/15922510886/job/44923494317?pr=1790) for a reason that could be related to this PR. Do you have access to a windows machine to check what is happening?

paul-buerkner:
So there is an issue in file name encodings (windows vs. unix style):

```R
> cache_res$file
[1] "C:/Users/Jedhoff/AppData/Local/Temp/RtmpSkKiJI/file1a587d5250d3.rds"

> fake_fit_file$file
[1] "C:\\Users\\Jedhoff\\AppData\\Local\\Temp\\RtmpSkKiJI\\file1a587d5250d3.rds"
```

paul-buerkner:
I think we got a fix. Will push it soon

paul-buerkner:
Looks good now. merging.

PR #1793, opened 2025-06-29
file_auto parameter for brm function

SermetPekin:
### 🚀 Summary of Changes

This PR introduces file_auto parameter for `brm()`: 

#### ✅ New Features

- **`file_auto = TRUE`**  
  Automatically computes a hash from key arguments to generate a file name, enabling:
  - Auto-reuse of cached fits
  - overrides file argument creating a hash for the call
  - overrides file_refit argument with `on_change` 

#### 📦 Internal Improvements

 - creates tests/testthat/setup.R file for setting up temporary options for tests using `withr` package  (suggests withr )


DESCRIPTION
```diff
@@ -52,7 +52,8 @@ Imports:
     utils,
     parallel,
     grDevices,
-    backports
+    backports,
+    digest
```
paul-buerkner, line 56: which aspects make digest imports? could it also be suggests with the usual brms check if the package is installed when needed?

R/brm.R
```diff
@@ -231,6 +231,11 @@
 #'   (e.g., initial values, number of iterations, control arguments, ...). A
 #'   known limitation is that a refit will be triggered if within-chain
 #'   parallelization is switched on/off.
+#' @param file_auto Logical If \code{TRUE} file parameter will be calculated
```
paul-buerkner, line 234: I would call it "argument" not "parameter". Also put `file` and other arguments mentioned here in \code{}

R/brm.R
```diff
@@ -231,6 +231,11 @@
...
+#'   automatically with the parameters given omitting the ones that do not change
```
paul-buerkner, line 235: the doc here is unclear. can you try to improve it? If you are unsure how, I can edit it directly

R/brm.R
```diff
@@ -452,21 +457,15 @@ brm <- function(formula, data, family = gaussian(), prior = NULL,
                 algorithm = getOption("brms.algorithm", "sampling"),
                 backend = getOption("brms.backend", "rstan"),
                 future = getOption("future", FALSE), silent = 1,
-                seed = NA, save_model = NULL, stan_model_args = list(),
+                seed = NA,
+                save_model = NULL,
```
paul-buerkner, line 461: continue to keep them in the same line

R/brm.R
```diff
@@ -452,21 +457,15 @@ brm <- function(formula, data, family = gaussian(), prior = NULL,
...
+                stan_model_args = list(),
                 file = NULL, file_compress = TRUE,
                 file_refit = getOption("brms.file_refit", "never"),
+                file_auto = getOption("brms.file_auto", FALSE ),
```
paul-buerkner, line 465: remove space before brackets

R/brm.R
```diff
@@ -486,6 +485,83 @@ brm <- function(formula, data, family = gaussian(), prior = NULL,
   empty <- as_one_logical(empty)
   rename <- as_one_logical(rename)
 
+  # this check is to allow other tests to continue doing what they were expected to do
+  #   when function was called with data parameter missing. We avoid an early
+  #   fail by adding this conversion to d for in our hash function
+  if(missing(data)){
```
paul-buerkner, line 491: ensure proper spacing. see the rest of the code base for style

R/brm.R
```diff
@@ -486,6 +485,83 @@ brm <- function(formula, data, family = gaussian(), prior = NULL,
...
+    d<- NULL
+  }else{
+    d <- data
+  }
+
+  # This list must include only/all the parameters that may change the result
+  .params_list <- list(
```
paul-buerkner, line 498: again proper spacing

paul-buerkner, line 498: I suggest you to use nlist for auto-naming. will make this here shorter

R/brm.R
```diff
@@ -486,6 +485,83 @@ brm <- function(formula, data, family = gaussian(), prior = NULL,
   empty <- as_one_logical(empty)
   rename <- as_one_logical(rename)
 
+  # this check is to allow other tests to continue doing what they were expected to do
+  #   when function was called with data parameter missing. We avoid an early
+  #   fail by adding this conversion to d for in our hash function
+  if(missing(data)){
+    d<- NULL
```
paul-buerkner, line 492: is this if-statement necessary here? Can we make it more clear? Can we perhaps just default data = NULL? (you would need to check if the errors are still informative)

R/brm.R
```diff
@@ -486,6 +485,83 @@ brm <- function(formula, data, family = gaussian(), prior = NULL,
...
+    stan_model_args = stan_model_args,
+    # file = file,
+    # file_compress = file_compress,
+    # file_refit = file_refit,
+    empty = empty
+    # rename = rename,
+    # ...
+  )
+  # one way hash from parameters
+  hash  <-  digest::digest(
```
paul-buerkner, line 528: so hashing is always done but we only need it in certain cases. Shall we make this conditional?

R/brm.R
```diff
@@ -486,6 +485,83 @@ brm <- function(formula, data, family = gaussian(), prior = NULL,
...
+  # this check is to allow other tests to continue doing what they were expected to do
+  #   when function was called with data parameter missing. We avoid an early
+  #   fail by adding this conversion to d for in our hash function
+  if(missing(data)){
+    d<- NULL
+  }else{
+    d <- data
+  }
+
+  # This list must include only/all the parameters that may change the result
```
paul-buerkner, line 497: there is now so much code overhead here that I would prefer not having all of it in the brm() function. can we do it in a separate function that is just being called here?

paul-buerkner:
Did you close this in purpose? You can just edit your existing PR

R/brmsfit-helpers.R
```diff
@@ -1091,6 +1091,19 @@ check_brmsfit_file <- function(file, .check_folder = TRUE) {
   full_name
 }
 
+
+# formula and family parameters may come with their own environment which requires
+#  special care when we need to create a hash from call
+clean_for_hash <- function(x) {
```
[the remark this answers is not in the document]
SermetPekin: we could basically use match.call() and distinguish many of the cases such as gaussian(), student() in the family and formulas etc however since we use partial matching in our arguments we preferred to have a same hash when partial matches evaluated to same values and for data we would like to have a different hash even when we have a different row than the previous one in order to make sure it evaluates to another result. 
for family case I fixed it to deparse1 method  and just for a quick note we are just making a copy of all the arguments passed without touching their original values inside our `brm()` method.  
```
clean_for_hash <- function(x) {
  if (inherits(x, "formula")) {
    environment(x) <- NULL
    x <- as.character(x)
  }
  if (inherits(x, "family")) {
   x <- deparse1(substitute(x))
  }
  x
}
```

R/brm.R
SermetPekin: Now we have a better smaller condition inside brm() function just for file_auto just to override file and file_refit parameters when argument was given  as `file_auto=TRUE` which is also quite better for our unit tests.

R/brm.R
SermetPekin: Now we have a better smaller condition inside brm() function just for file_auto just to override file and file_refit parameters when argument was given as file_auto=TRUE which is also quite better for our unit tests.

R/brm.R
```diff
@@ -486,6 +485,83 @@ brm <- function(formula, data, family = gaussian(), prior = NULL,
...
+    thin = thin, cores = cores,
+    threads = threads, opencl = opencl,
+    normalize = normalize, control = control,
+    algorithm = algorithm, backend = backend,
+    future = future,
+    silent = silent,
+    orig_seed = orig_seed,
+    # save_model = save_model,
+    stan_model_args = stan_model_args,
+    # file = file,
```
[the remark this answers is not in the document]
SermetPekin: It was on purpose wanted to show which arguments I am not including in the hash function but now I removed them and it looks better.

R/brm.R
```diff
@@ -486,6 +485,83 @@ brm <- function(formula, data, family = gaussian(), prior = NULL,
...
+  hash  <-  digest::digest(
+    lapply(.params_list, clean_for_hash),
+    algo = "xxhash64"
+  )
+
+
+  # Handle file_auto is TRUE case
+  #   will define a value for file argument automatically to return previous result
+  #   with same parameters
+  if( isTRUE( file_auto  ) ) {
```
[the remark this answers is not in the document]
SermetPekin: Thanks. I fixed this one too.

R/brm.R
```diff
@@ -605,4 +693,4 @@ brm <- function(formula, data, family = gaussian(), prior = NULL,
     x <- write_brmsfit(x, file, compress = file_compress)
   }
   x
-}
+}
```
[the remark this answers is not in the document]
SermetPekin: fixed this one.

SermetPekin:
Hi Paul,
Thank you for your review. I have addressed all your comments:

    Refactored the refit_auto condition to minimize its effect on the overall volume of the brm() function.
    Fixed all spacing and formatting issues.

Please let me know if you have any further feedback.

tests/testthat/tests.brm_file_auto.R
```diff
@@ -0,0 +1,86 @@
+# Mock brm function to check file_auto parameter
+brm_mock_for_file_auto <- function(formula, data = NULL , family = gaussian(), prior = NULL,
```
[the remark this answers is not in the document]
SermetPekin: I will check mock backend. Let me check if I can get file or hash information with params of (... empty = TRUE , backend='mock' , file_auto=TRUE ). What I need is a quick check for the different hash values created for each different call or same hash for similar call.

SermetPekin:
I am still considering the best approach to robustly hash our brm call. One method I would like to propose is incorporating the generated stancode—along with the data, the brms version, backend, and a few other relevant elements that do not influence the stancode. This should help ensure that the hash accurately reflects the essential components of the model call.

However, this approach may introduce some overhead, as generating the stancode would need to occur earlier or potentially twice. Alternatively, we might need to adjust our file and refit cache check to happen slightly later within the brm function than is currently implemented. I wanted to highlight these trade-offs for consideration.

paul-buerkner:
One option is that you go recursively through the list structure of the formula and family arguments and remove all environment attributes if they exist

SermetPekin:
> One option is that you go recursively through the list structure of the formula and family arguments and remove all environment attributes if they exist

Thank you for the suggestion. I will implement this approach.

SermetPekin:
Hi Paul, @paul-buerkner,

I’d like to propose adding S3 `hash()` methods for the key parameter types accepted by brms, such as `formula`, `family`, `data.frame`, `prior`, and `stanvars`.

The main goal is to ensure that hashing captures only the model-defining content—by removing environments, sorting named lists, and hashing data frames based on their actual content rather than object reference. 

Do you think this approach would provide a robust foundation for caching and reproducibility in brms, or do you see potential drawbacks or complexity that might outweigh the benefits?

paul-buerkner:
that could work and would be a clean approach if correctly implemented.

another question: will this then also replace the on_change checking? I.e.
that the hashing will be used to determine if a relevant change has
happened? or does that not make sense?

Sermet Pekin ***@***.***> schrieb am Do., 3. Juli 2025, 08:41:

SermetPekin:
I was planning to override file_refit as `on_change` to make it possible an extra check by the current design which checks `brmsfit_needs_refit` currently our addition is only creating a file argument for the user automatically and setting  `file_refit` as `on_change` to make it possible to get a fast result back. I hope it would be extra secure to have a few redundant refit than having wrong cache result although something is different that our cache function did not capture.

SermetPekin:
To make things even more robust, we could add another layer of checking by hashing the generated stancode after the initial check passes. That way, we can be extra sure that nothing important has changed under the hood, and we avoid any chance of returning a stale or incorrect cached result.

SermetPekin:
> I have quickly checked parts of the hashing but not yet all. I assume you still want to work on it before I properly review?

This PR is actually kind of draft now. I would like to work on this PR until we are sure it captures all parameters and function properly. I am happy to get feedback on the way, however most of these functions and files are actually draft and will be replaced and refactor to the code base with minimum change. Some other small PRs that I hope I can propose can help this PR to better test and function better. 
Such as a tailor made `brm` call class proposal if accepted which would I think it could help us refactor existing `brm()` function body to become a bit more concise before adding one more condition. 
I would be happy to get advice from you but still some more work to do here.

paul-buerkner:
Okay, let me know once you would like some feedback :-)

R/brm-hash.R
```diff
@@ -0,0 +1,252 @@
+# Helper function that will capture formal elements of brm function and return
+# as a list
+brm_to_call <- function(formula, data = NULL, family = gaussian(), prior = NULL,
```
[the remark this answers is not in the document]
SermetPekin: this is actually helping me while testing. I will remove it soon. I will be using empty=TRUE case which I tried previously but I get error  when I try to debug with with summary method. I will try again since we fixed that one.

SermetPekin:
> Okay, let me know once you would like some feedback :-)

Hi Paul,
I’ve opened another PR that refactors the brm() method by introducing a few internal helper functions to delegate specific tasks. I believe this structure will make it easier to design and test new features moving forward.

One key improvement is the introduction of the brm_call class, which encapsulates all arguments passed to brm(). This made it simpler to pass call parameters between functions and streamlined the hashing process by removing conditional logic from the main brm() body.

Because brm_call retains all the necessary components for generating Stan code, we were also able to compute the hash value only when needed—specifically, when file_auto = TRUE is set—without complicating the main function.

The brm_call class is created and used entirely internally, so it won't affect existing usage or require any changes from users.

Happy to hear your thoughts or suggestions on the approach.

paul-buerkner:
Thanks! Is that email saying you would like the PR to be reviewed now?

Am So., 6. Juli 2025 um 22:22 Uhr schrieb Sermet Pekin <
***@***.***>:

SermetPekin:
- I should probably close this PR cause a lot will be already changed with other PR - brm_call class -  I will be revisiting this to read your valuable feedback and to copy some functions I created here to move to other PR. though.

PR #1797, opened 2025-07-02
require testthat (>= 3.1.9) in Suggests

### LINKED ISSUE #1796: Suggestion: Update testthat version in DESCRIPTION to support expect_in
It appears that the `expect_in` function we are using is available only in a later version of the `testthat` package. I would like to suggest updating the `testthat` version requirement in the `DESCRIPTION` file to ensure compatibility.
```plaintext
testthat 3.1.9

New expect_contains() and expect_in() that works similarly to expect_true(all(expected %in% object)) or expect_true(all(object %in% expected)) but give more informative failure messages (#1346).
```

SermetPekin:
Update testthat version in Suggests to >= 3.1.9, since recent tests depend on it.

Closes #1796.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -54,7 +54,7 @@ Imports:
     grDevices,
     backports
 Suggests:
-    testthat (>= 0.9.1),
+    testthat (>= 3.1.9),
     emmeans (>= 1.4.2),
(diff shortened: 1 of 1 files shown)


paul-buerkner:
Thanks!

PR #1801, opened 2025-07-03
add colorspace to Suggests

SermetPekin:
add colorspace to Suggests


paul-buerkner:
This seems to fix it. Thank you!

Issue #1789, opened 2025-06-20
get_inits()?


ASKurz:
Would you be open to supporting a `get_inits()` function, such as used by {rstan}? See https://mc-stan.org/rstan/reference/stanfit-class.html

paul-buerkner:
Can you elaborate what exactly you have in mind?

ASKurz:
Sure. I'm guessing the {rstan} implementation is more rigorous than this, but at a basic level I'm looking for something that makes it easier to extract the initial values from a model. E.g., 

```
get_inits <- function(object) {
  # Check
  if(!is.brmsfit(object)) stop("`object` must be of class `brmsfit`.")
  object$fit@inits
}

```

paul-buerkner:
Sure, why not (via an S3 method). Could you make a PR for that perhaps?

ASKurz:
That's above my skill set. I don't really know what an S3 method is.

paul-buerkner:
It is not difficult and it could be a good practice if you want to invest half an hour or so. Essentially you define an S3 generic like so:

```R
#' documentation
#' @export
brmsterms <- function(x, ...) {
  UseMethod("brmsterms")
}
```

and then define individual methods like so:

```R
#' @rdname brmsterms
#' @export
brmsterms.formula <- function(x, data, ...) {
  ...
}
```

paul-buerkner:
Now implemented, thanks to @behramulukir .

PR #1813, opened 2025-08-05
adding get_inits()

behramulukir:
In #1789, the implementation of `get_inits()` function, which would work similarly to [rstan's version](https://mc-stan.org/rstan/reference/stanfit-class.html) is requested. As the task didn't seem too complicated, I decided to work on this small request.

As far as I understood, brms currently has two backend options, `rstan` and `cmdstanr`. However, as stated in #1807, results are stored as `stanfit` objects regardless of the backend used in the model. That being said, it seems like there is work to store results directly as `cmdstanfit` objects, if the backend is chosen as `cmdstanr`. Based on my searches, `get_inits()` function only exists in `rstan`, and I couldn't find any way to fetch the inits for `cmdstanfit` objects. Another point my tests revealed was that when models with `cmdstanr` backends are transformed into `stanfit` objects, the information regarding their initial values is erased.

Based on this information, my implementation simply relies on checking the backend used to fit the model, and if it is an `rstan` model, calling `rstan::get_inits()`. If the backend is `cmdstanr`, the function throws an error as it is capable of retrieving initial values for `rstan` backends.

I also wrote a couple of lines of tests to make sure it keeps working correctly, as well as creating a new `brmsfit_example7` which is the same as `brmsfit_example1` except that this new example uses `cmdstanr` as its backend.

According to unit tests as well as my manual testing, the function should be working without an issue. However, if there is any issue, please let me know. I will take a deeper look into it.


paul-buerkner:
Thank you! @ASKurz is this what you had in mind?

@behramulukir can you please remove the `brmsfit_example7` again. It further increases the size of the R package, which CRAN already complains about for years. So I cannot effort more example fits at the moment.

ASKurz:
I haven't reviewed the code, but this sounds good, to me.

behramulukir:
Alright, I removed `brmsfit_example7`. I hope everything is good now!

paul-buerkner:
Thank you! I renamed get_inits to inits (to avoid name clashing mit rstan). I will run checks now and merge as soon as they pass.

PR #1814, opened 2025-08-06
validate_newdata(): ignore levels missing in new/old data

alyst:
I've run into an issue when applying the fit model to a new data, because one of the factors had the levels there were not present in the data used to fit the model.
However, the new data contained no actual observations with the factor set to that level, so it was a false alarm.

OTOH, when looking at the code of `validate_newdata()`, it seems the false no-alarm may also happen -- the training data may lack some specific levels of the factor (and the corresponding effects too), but the check will consider that it is there.

This PR should fix both errors by calling `droplevels(factorvar)` for new and old data to ignore the levels that are not actually present in the dataset.

### DIFF
--- a/R/data-helpers.R
+++ b/R/data-helpers.R
@@ -494,12 +494,12 @@ validate_newdata <- function(
         if (!is.factor(new_factor)) {
           new_factor <- factor(new_factor)
         }
-        old_levels <- levels(factors[[i]])
+        old_levels <- levels(droplevels(factors[[i]]))
         if (length(old_levels) <= 1L) {
           # contrasts are not defined for factors with 1 or fewer levels
           next
         }
-        new_levels <- levels(new_factor)
+        new_levels <- levels(droplevels(new_factor))
         old_contrasts <- contrasts(factors[[i]])
         old_ordered <- is.ordered(factors[[i]])
         to_zero <- is.na(new_factor) | new_factor %in% "zero__"


paul-buerkner:
Thank you! Can you provide a minimal reprex when this issue occurs? I want to make sure it is not intentional behavior

paul-buerkner:
Quick follow up. Is this PR still relevant?

alyst:
It should be relevant, unless it was fixed independently. Sorry, I was busy/away. I'll try to provide MRE soon.

paul-buerkner:
No worries, thank you!

jsocolar:
Here's a reprex.  A `brmsfit` stores in `fit$data` something like `droplevels(df)` where `df` is the input dataframe.

`validate_newdata` does something like compare `levels(fit$data$var)` to `levels(newdata$var)`.  

The question is whether it should instead compare `levels(fit$data$var)` to `levels(droplevels(newdata)$var)`.

```
library(brms)

df0 <- data.frame(
  y = rnorm(30), 
  x = factor(
    c("a", "b", "c"), 
    levels = c("a", "b", "c")
    )
  )

df1 <- data.frame(
  y = rnorm(30),
  x = factor(
    c("a", "b", "c"), 
    levels = c("a", "b", "c", "d")
  )
)

df2 <- data.frame(
  y = rnorm(40),
  x = factor(
    c("a", "b", "c", "d"), 
    levels = c("a", "b", "c", "d")
  )
)

m0 <- brm(
  y ~ x, 
  data = df0,
  backend = "cmdstanr"
  )

m1 <- brm(
  y ~ x, 
  data = df1,
  backend = "cmdstanr"
)

# The below errors, because it finds new factor levels in df1,
# even though those factor levels are not in `droplevels(df1)`.
# This error may or may not be intended behavior.
predict(m0, newdata = df1)

# Note that the below also errors, for the same reason, which 
# might be especially unexpected.
predict(m1, newdata = df1)

# this works
predict(m0, newdata = droplevels(df1))

# this is expected to error, and it does error informatively. @alyst was
# worried that this might not error informatively, as every level in `df2`
# is also present in `df1`, but brms gets this one right by only storing
# `droplevels(df1)` in `m1$data`.
predict(m1, newdata = df2)
```

paul-buerkner:
Got it. So the important new line is `new_levels <- levels(droplevels(new_factor))`. 

The other new line `old_levels <- levels(droplevels(factors[[i]]))` is potentially dangerous since it may remove unused levels that are intentionally present.

paul-buerkner:
I made the change and added also a small test directly to master. Thank you for contributing to brms!

alyst:
> The other new line old_levels <- levels(droplevels(factors[[i]])) is potentially dangerous since it may remove unused levels that are intentionally present.

Does it mean that the BRMS model will contain effects/parameters for the missing levels -- and their values will be inferred based on priors only?
This is also potentially dangerous as `predict(newdata)` will then use these parameters to predict the outcomes in data, where these levels are actually present --- resulting in unreliable predictions.
Should the users be given a warning in these situations?

paul-buerkner:
Only if users set `drop_unused_levels = FALSE` in `brm` and then they know the implications.

alyst:
> Only if users set drop_unused_levels = FALSE in brm and then they know the implications.

Ah, ok, I missed that flag. :) It might be still worth to give a warning, especially in some verbose execution modes, since `predict()` is another call (and the user that calls `predict()` might be different from the user that called `brm()`).

PR #1818, opened 2025-08-25
Improved posterior density approximation

### LINKED ISSUE #1816: Negative `Evid.Ratio` from `hypothesis` function if BayesFactors are very small
Every now and then I get a negative `Evid.Ratio` when using the `hypothesis` function to evaluate hypothesis on a `brmsfit` object. I assume this is related to the approximation of the posterior likelihood at zero. From what I can see, currently `brms` uses a spline approximation of the density relying on `baseR` functions. This seems not to be numerically stable, if there are very few samples around zero for a posterior that is far away from zero.

Other packages have used the `dlogspline` function from the `logspline` package that seem to provide more stable likelihood approximations based on posterior or prior samples. I understand if you do not want to add another dependency to the package, but it would still be nice to have more stable BF estimation.

I do understand that in these cases the exact numerical value of the Bayes Factor might not really be relevant as the BF provides clear evidence against the null hypothesis. But for reporting it is still somewhat annoying to deal with these negative BayesFactors, as you have to recompute them by yourself using the prior and posterior samples saved in the hypothesis object.

Another alternative would be to provide an option for how the BayesFactors should be estimated and include the option for a normal approximation, that is calculating the mean and standard deviation of the prior and posterior and using these to evaluate the density of the normal distribution at a given value: `dnorm(0, mean = mean(posterior), sd = sd(posterior)`. In many cases this also provides an adequate approximation of the Bayes Factor, assuming that both priors and posteriors are approximately normal, which might not always be the case.

Here is a short example illustrating the problem with the current density approximation and how the log spline or normal approximation avoids the problem of negative density estimations:

```
n_iter <- 200
n_postSamples <- 50000

densities <- numeric(n_iter)
densities_logspline <- numeric(n_iter)
densities_normal <- numeric(n_iter)
densities_true <- numeric(n_iter)

eval_density <- function(x, point) {
  # evaluate density of x at point
  from <- min(x)
  to <- max(x)
  if (from > point) {
    from <- point - sd(x) / 4
  } else if (to < point) {
    to <- point + sd(x) / 4
  }
  dens <- density(x, from = from, to = to)
  return(spline(dens$x, dens$y, xout = point)$y)
}

eval_density_logspline <- function(x, point) {
  logspline::dlogspline(point, logspline::logspline(x))
}

eval_density_normal <- function(x, point) {
  dnorm(point, mean = mean(x), sd = sd(x))
}

for (i in 1:n_iter) {
  rand_mean = runif(1, 8,12)
  rand_sd = runif(1, 0.5,1)
  x = rnorm(n_postSamples, mean = rand_mean, sd = rand_sd)
  
  densities[i] <- eval_density(x, 0)
  densities_logspline[i] <- eval_density_logspline(x, 0)
  densities_normal[i] <- eval_density_normal(x, 0)
  densities_true[i] <- dnorm(0, mean = rand_mean, sd = rand_sd)
}

min(densities)
min(densities_logspline)
min(densities_normal)

plot(densities, densities_true)
plot(densities_logspline, densities_true)
plot(densities_normal, densities_true)
```

GidonFrischkorn:
Resolve #1816 by improving the density approximation in `density_ratio`. Following [Deng & Wickham (2011)](https://vita.had.co.nz/papers/density-estimation.pdf) I suggest to implement the density approximation implemented in the `KernSmoot.h` package.

Specifically, I changed the density approximation from `stats::density` to `KernSmooth::bkde` and then approximated the density at the required `point` using `stats::approx`.

Initially, I implemented the density approximation with the `logspline` package, but the examples of the `hypothesis` function threw some warnings, and I thought this would not be ideal.

Let me know, what you think of these changes and if I should run some additional tests.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -46,8 +46,7 @@ Imports:
     utils,
     parallel,
     grDevices,
-    backports,
-    logspline
+    backports
 Suggests:
     testthat (>= 0.9.1),
     emmeans (>= 1.4.2),
@@ -73,7 +72,8 @@ Suggests:
     gtable,
     shiny,
     knitr,
-    rmarkdown
+    rmarkdown,
+    KernSmooth
 Description: Fit Bayesian generalized (non-)linear multivariate multilevel models
     using 'Stan' for full Bayesian inference. A wide range of distributions
     and link functions are supported, allowing users to fit -- among others --
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -46,7 +46,8 @@ Imports:
     utils,
     parallel,
     grDevices,
-    backports
+    backports,
+    logspline
 Suggests:
     testthat (>= 0.9.1),
     emmeans (>= 1.4.2),
--- a/R/hypothesis.R
+++ b/R/hypothesis.R
@@ -447,8 +447,7 @@ density_ratio <- function(x, y = NULL, point = 0, n = 4096, ...) {
     } else if (to < point) {
       to <- point + sd(x) / 4
     }
-    dens <- do_call(density, c(nlist(x, from, to), dots))
-    return(spline(dens$x, dens$y, xout = point)$y)
+    logspline::dlogspline(point, logspline::logspline(x, lbound = from, ubound = to))
   }
 
   out <- ulapply(point, eval_density, x = x)
--- a/R/hypothesis.R
+++ b/R/hypothesis.R
@@ -332,8 +332,8 @@ eval_hypothesis <- function(h, x, class, alpha, robust, name = NULL) {
   # summarize hypothesis
   wsign <- switch(sign, "=" = "equal", "<" = "less", ">" = "greater")
   probs <- switch(sign,
-    "=" = c(alpha / 2, 1 - alpha / 2),
-    "<" = c(alpha, 1 - alpha), ">" = c(alpha, 1 - alpha)
+                  "=" = c(alpha / 2, 1 - alpha / 2),
+                  "<" = c(alpha, 1 - alpha), ">" = c(alpha, 1 - alpha)
   )
   if (robust) {
     measures <- c("median", "mad")
@@ -447,7 +447,8 @@ density_ratio <- function(x, y = NULL, point = 0, n = 4096, ...) {
     } else if (to < point) {
       to <- point + sd(x) / 4
     }
-    logspline::dlogspline(point, logspline::logspline(x, lbound = from, ubound = to))
+    density <- KernSmooth::bkde(x, range.x = c(from,to), gridsize = n)
+    stats::approx(density$x, density$y, xout = point)$y
   }
 
   out <- ulapply(point, eval_density, x = x)


GidonFrischkorn:
I noticed that `KernSmooth::bkde` can still return negative densities. So maybe we should go with `logspline` although it might throw an occasional warning.

What do you think @paul-buerkner ?

paul-buerkner:
Yes I think so too. Will you make that change?

Gidon Frischkorn ***@***.***> schrieb am Mo. 25. Aug. 2025 um
16:22:

GidonFrischkorn:
Yes, I take care of that.

paul-buerkner:
Any could you make the PR against the latest version of master? Right now I
think the PR has merge conflicts

Paul Buerkner ***@***.***> schrieb am Mo. 25. Aug. 2025 um
16:22:

GidonFrischkorn:
I have just pushed the required changes, and also looked into the tests for the `evidence_ratio` function. I have slightly adapted them, as the previous tests did not work with the `logspline` approximation. I hope this was ok.

paul-buerkner:
Thank you! For me, the old tests don't return a warning, but the following test just fails:

```R
expect_true(evidence_ratio(ps, prior_samples = prs) > 1)
```

I get around `0.811` as result but `> 1` is expected. Do you have an intuition on what the problem here is?

paul-buerkner:
Perhaps my old test was garbage :-D I cannot remember how I came up with it, but I don't just want to ignore the failing test without understanding what happens.

GidonFrischkorn:
> Thank you! For me, the old tests don't return a warning, but the following test just fails:
> 
> ```r
> expect_true(evidence_ratio(ps, prior_samples = prs) > 1)
> ```
> 
> I get around `0.811` as result but `> 1` is expected. Do you have an intuition on what the problem here is?

hmm... Maybe with the number of samples I choose the approximation is not very stable. I assume increasing the number of samples to `1000` or even more should reliably result in the test not failing.

GidonFrischkorn:
now it should not fail anymore.

paul-buerkner:
Sorry, I meant my old tests + old data (the data you erased). This one was failing and I was wondering why.

paul-buerkner:
It seems you are approximating the log density with dlogspline not the density. Can you please double check that the approach is correct the way you implemented it? Of course, it is easy to go from log density to density, should that be the case.

 Could it be that this then also explains the old test failing? Please use my original tests as basis and try to get them to pass. And can you check for an example model if the old and new version yields the same results? Sorry to bother you with this. I just need to make sure this works out correctly.

GidonFrischkorn:
The `dlogspline` function approximates the density but with a `logspline` fit object. The naming is somewhat confusing, but the values returned are actual densities.

But I noticed that the specification of the bounds severely changes the density estimates as the bounds impose a strong constraint that the density is zero for any value below `lbound` and any value above `ubound`. So I deleted the inclusion of bounds for the density approximation and then the new `density_ratio` function passes also the old tests.

GidonFrischkorn:
Here is an example how the bounds influence the density approximation. The blue and red lines are `logspline` density approximations without bounds. And the green and grey linse are the density approximations with bounds. 
 
```
ps <- -4:10
prs <- -2:12

density_ps_new <- logspline::logspline(ps)
density_prs_new <-logspline::logspline(prs)
density_ps_new_bounds <- logspline::logspline(ps, lbound = min(ps), ubound = max(ps))
density_prs_new_bounds <- logspline::logspline(prs, lbound = min(prs), ubound = max(prs))

plot(density_ps_new, col = "red")
plot(density_prs_new, col = "blue", add = TRUE)
plot(density_ps_new_bounds, col = "darkgreen", add = TRUE)
plot(density_prs_new_bounds, col = "darkgray", add = TRUE)
```

image

GidonFrischkorn:
With respect to comparing the old and new results, would it be ok to take the examples provided in the hypothesis function and compare results from the current results using the old `density_ratio` function and compare these with the results from the updated `density_ratio` approximation?

This is something I should be able to provide within the next few days.

paul-buerkner:
Yes, that would be okay. Of course, only the "=" tests are relevant because only these use the density_ratio.

I have just run the first example and get for one of the tests the warnings: 

```R
1: In logspline::logspline(x) : too much data close together
2: In logspline::logspline(x) : re-ran with oldlogspline
```

It would be great if this warning message would not come up because I assume it is a quite frequent one.

GidonFrischkorn:
Yes, I had similar thoughts about the warning message. One option would be to wrap the fitting of the `logspline` in a `suppressWarnings` call. But I wanted to check with you, if that's ok within the package. 

And I will prepare a comparison of the `hypothesis` results of the example with both methods.

paul-buerkner:
suppressWarnings is not ideal there because we don't know what other important warnings we may be suppressing. Is there perhaps some other option? If not, surprising warnings is okay for me.

GidonFrischkorn:
After all the issues with the `logspline` package, I looked into the current function again, and it seems the problem is the interpolation of the density value at the required `point` using the `spline` function. The `density` function itself does not return any densities smaller than zero, however the `spline` approximation can yield negative values.

So, my suggestion is to keep the original density approximation using `stats::density` but switch to `stats::approx` as interpolation method. This will be a less sophisticated linear interpolation of the density at the required `point`, but has the benefit of avoiding negative values.

The alternative would be to use `pmax` around the current `spline` interpolation and set the density interpolation to zero, whenever the density is interpolated to values below zero.

I am sorry for the back and forth over this small issue.

In any case. I would consider including a warning in any case where the density is approximated as zero that either more posterior samples are required or there are too few samples around the `point` of interest to adequately approximate the posterior density.

paul-buerkner:
Thanks. I have now added the pmax approach and pushed to master. :-)

PR #1822, opened 2025-09-01
Updated default prior for phi to inv_gamma(0.4, 0.3)

jan-joh:
Updated default prior for negative binomial's phi was previously overseen for zero-inflated and hurdle models. I have added the new prior to the family-lists.R(experimental) and also to priors.R. The documentation should already be accurate, mentioning inv_gamma(0.4, 0.3) for negative binomial families.

### DIFF
--- a/R/family-lists.R
+++ b/R/family-lists.R
@@ -513,7 +513,14 @@
     ad = c("weights", "subset", "cens", "trunc", "index"),
     include = "fun_hurdle_negbinomial.stan",
     specials = c("sbi_log", "sbi_hu_logit"),
-    normalized = ""
+    normalized = "",
+    # experimental use of default priors stored in families #1614
+    prior = function(dpar, link = "identity", ...) {
+      if (dpar == "shape" && link == "identity") {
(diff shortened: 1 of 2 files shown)


paul-buerkner:
Great! You also seem to have changed the default prior for shape more generally? This will also affect a few other families. Is this intentional?

jan-joh:
Oh, sorry, I thought this just applied to negative binomial families. I personally can't tell if the new `inv_gamma(0.4, 0.3)` is a better default for the others as well. It surely wasn't intentional. If you think we should apply the `inv_gamma(0.4, 0.3)` only to the negbin family, I can quickly revert the change in `Priors.R` (I guess that's the one affecting the other families?)

paul-buerkner:
Yes reverting this would be great!

JohJan ***@***.***> schrieb am So. 7. Sept. 2025 um 14:35:

paul-buerkner:
Quickly changed it myself and merging now. Thank you for opening this PR and for contributing to brms!

PR #1824, opened 2025-09-04
extend documentation on parallelisation and suggest use of mirai via future.mirai

wds15:
Adds instructions for users how to consider parallelisation and suggests the use of mirai via future.mirai as it enables to push companions to a computer cluster.

Optionally we could also add a link to https://opensource.nibr.com/bamdd/src/03b_parallel.html where these things are detailed, but that is not yet in the text.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -82,7 +82,9 @@ Suggests:
     knitr,
     rmarkdown,
     ragg,
-    colorspace
+    colorspace,
+    mirai,
+    future.mirai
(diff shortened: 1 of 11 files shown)


paul-buerkner:
Thanks! You have added an @template call but I don't quite see where you have defined the template. Am I just overlooking it?

wds15:
I forgot to add that new file... now there.

paul-buerkner:
Thank you! Merging now!

PR #1835, opened 2025-11-13
FIX: `update.brmsfit` returns error when updating an empty fit

trinhdhk:
An empty fit is useful when users make post-hoc edit on Stancode but still want to leverage brmsmethods like `standata`.
Currently `update.brmsfit` throws S4 error when trying to access `object$fit@sim`.  I think this is R's design inconsistency when `$` is a safe call but `@` isn't.

This pull request fixes this issue by implementing a check on `object$fit`.

```r
empty_fit <- !isS4(object$fit)
```

Test:

```r
m0 <- brm(mpg~cyl, data=mtcars, empty=TRUE)
mtcars2 <- dplyr::mutate(mtcars, cyl=1)
m1 <- update(m0, newdata=mtcars2, empty=TRUE)
m1; head(m1$data)
```

Expected output:
- m1 should be printed properly as an empty fit
- m1$data$cyl should be 1

```
 Family: gaussian 
  Links: mu = identity 
Formula: mpg ~ cyl 
   Data: mtcars2 (Number of observations: 32) 

The model does not contain posterior draws.
                   mpg cyl
Mazda RX4         21.0   1
Mazda RX4 Wag     21.0   1
Datsun 710        22.8   1
Hornet 4 Drive    21.4   1
Hornet Sportabout 18.7   1
Valiant           18.1   1
```

Test for recomplication:
```r
m1 <- update(m0, newdata=mtcars2)
```

Expected output: 
- The message "The desired updates require recompiling the model" should not appear.

```
Compiling Stan program...
```

---

**FIXME**:

I also put some safeguard in `update.brmfit_mutiple` but I did not use this function extensively enough to assure that is sufficient. From the surface I think it would err out when calling `brm_multiple(empty=TRUE)`:

```
Compiling the C++ model
Fitting imputed model 1
Warning: Caught simpleError. Canceling all iterations ...
Error:
! unable to find an inherited method for function ‘get_stanmodel’ for signature ‘object = "NULL"’
```

I think it is not even allowed. Given the ancient cat "Compiling the **C++** model" I believe it has been long since this function is updated.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -113,4 +113,4 @@ SystemRequirements: pngquant
 VignetteBuilder:
     knitr,
     R.rsp
-RoxygenNote: 7.3.2
+RoxygenNote: 7.3.3
--- a/R/brmsformula.R
+++ b/R/brmsformula.R
@@ -992,7 +992,7 @@ split_bf <- function(x) {
   resp <- terms_resp(x$formula, check_names = FALSE)
   str_adform <- formula2str(x$formula)
   str_adform <- get_matches("\\|[^~]*(?=~)", str_adform, perl = TRUE)
-  if (length(resp) > 1L) {
+  if (isTRUE(attr(resp, "mvbind"))) {
     # mvbind syntax used to specify MV model
     flist <- named_list(resp)
     for (i in seq_along(resp)) {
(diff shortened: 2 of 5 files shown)


paul-buerkner:
Thanks! I will review this PR soon.

Issue #1842, opened 2025-12-12
Issue using specials "ordinal" and "extra_cat" when specifying custom family


yuee289:
When `extra_cat` is supplied, it will return `family == "hurdle_cumulative" is not TRUE`. 

When `ordinal` is supplied, a block of code with unclosed bracket will be added on top of the stan code. 

Here I will use the hurdle_sratio custom family specified by @sjwild in #1429 as an example, I assume it worked in the past:

```
stan_funs <- "

   real hurdle_sratio_lpmf(int y, real mu, real hu, real disc, vector thres) {
     int nthres = num_elements(thres);
     real mu_logit = logit(mu);
     vector[nthres + 1] p;
     vector[nthres] q;
     int k = 1;
     if (y == 0){
      return bernoulli_lpmf(1 | hu);
     } else {
       while (k <= min(y, nthres)) {
         q[k] = log1m_inv_logit(disc * (thres[k] - mu_logit));
         p[k] = log1m_exp(q[k]);
         for (kk in 1:(k - 1)) p[k] = p[k] + q[kk];
         k += 1;
       }
       if (y == nthres + 1) {
         p[nthres + 1] = sum(q);
       }
     }
     return p[y] + bernoulli_lpmf(0 | hu);
   }

  "

stan_vars <- stanvar(scode = stan_funs, block = "functions")

hurdle_sratio <-
    # Create a custom family that is logit if y = DK, cumulative if not
    custom_family("hurdle_sratio",
        dpars = c("mu", "hu", "disc"),
        links = c("logit", "logit", "log"),
        specials = c("ordinal", "extra_cat"),
        type = "int",
        threshold = "flexible"
    )

marriage <- read.csv("https://osf.io/download/n6dzh/")

# set "divorced" to be hurdle
marriage$yearsh <- ifelse(marriage$divorced == FALSE, 0, marriage$years)

fit_ma2 <- brm(
    bf(
        yearsh ~ 1 + age + together,
        hu ~ 1 + age
    ),
    data = marriage,
    family = hurdle_sratio,
    stan_funs = stan_funs,
    prior = prior_ma2,
    cores = 4,
    chains = 4,
    iter = 3500,
    warmup = 1000
)
```
It will return:
```
Error in stan_hurdle_ordinal_lpmf("custom", "logit") : 
  family == "hurdle_cumulative" is not TRUE
```
If we remove `extra_cat` and extract the stan code we can see another issue
```
stan_code <- make_stancode(
[...]

PR #1843, opened 2025-12-13
Custom family fix when specials = "ordinal" or "extra_cat" is supplied 

yuee289:
see #1842


paul-buerkner:
Thank you! Made some minor changes and now ready to merge.

Issue #1836, opened 2025-11-13
Consider replacing `asym_laplace_*` with native version from Stan?


trinhdhk:
Stan since 2.28 has natively adopted Asym Laplace and named skew double exponential distribution. 

Perhaps it would speed up the sampler by using the native version for asym_laplace family?

paul-buerkner:
That is a great idea! Let's do that.

trinhdhk:
For backward compatibility I think we can create a new family and soft deprecate the old one. The only difference is that quantile in asym_laplace is formulated as tau in Stan.

PR #1845, opened 2025-12-19
Adding skew_double_exponential family

trinhdhk:
Adding experimental `skew_double_exponential` family, targeting #1836
Speed gains about 25%.
Caveat: currently I have noticed that `sigma` in Stan's implementation is twice larger than its counterpart in `asym_laplace`. This is a breaking change in terms of interpretation (and prior specs), so I'm still keeping `asym_laplace`.
I added one test but I think we need to add further test to ensure equivalence. 
Please check if I added the test in the correct directory (currently in local test). And kindly let me know.
And sorry for CLRF (if there's still left-over).

### DIFF
--- a/R/distributions.R
+++ b/R/distributions.R
@@ -1090,7 +1090,7 @@ dskew_double_exponential <- function(x, mu = 0, sigma = 1, tau = 0.5,
   dasym_laplace(
     x = x,
     mu = mu,
-    sigma = sigma,
+    sigma = sigma/2,
     quantile = tau,
     log = log
   )
@@ -1120,7 +1120,7 @@ pskew_double_exponential <- function(x, mu = 0, sigma = 1, tau = 0.5,
   pasym_laplace(
     x = x,
     mu = mu,
-    sigma = sigma,
+    sigma = sigma/2,
     quantile = tau,
     lower.tail = lower.tail,
     log.p = log.p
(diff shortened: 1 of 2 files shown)


paul-buerkner:
I am fine with replacing `asym_laplace` Stan code with the native implementation in Stan. I would prefer not to add a new family for this. That causes too much confusion. 

Do you know why one has sigma while the other has sigma / 2? Which one is the more "natural" parameterization?

trinhdhk:
Good question. I believe it has to do with the fact that they have to guarantee its equivalence to double exponential distribution at tau = 0.5.  See https://github.com/stan-dev/stan/issues/2312

Trinh Dong
________________________________

paul-buerkner:
Let's keep the current brms `asym_laplace` parameterization then but use Stan's built-in functions for fitting. Would you mind editing your PR accordingly? I am sorry that this means that we won't need most of the code you currently included in the PR.

trinhdhk:
> Let's keep the current brms `asym_laplace` parameterization then but use Stan's built-in functions for fitting. Would you mind editing your PR accordingly? I am sorry that this means that we won't need most of the code you currently included in the PR.

I'm happy to but it does mean a breaking change for past studies that have used asym laplace. For a mature software, this is not a good practice.

Most of my code now is just a wrapper for asym_laplace so an inplace replacement would reduce the time for testing equivalence etc. But like I said, I was thinking of the researchers (me).

paul-buerkner:
What would it break exactly?

trinhdhk:
> What would it break exactly?

Ah I misread. Did you mean doubling sigma before feeding into Stan so that we keep the current formulation?

paul-buerkner:
Correct.

trinhdhk:
Clear!

________________________________

PR #1850, opened 2026-01-17
fix convergence language

avehtari:
Some of the help texts talked about models converging, while they should have talked about the inference converging. This PR fixes the language.

(when taling about iterative model building it would be possible to talk about models converging)


paul-buerkner:
Thank you!

PR #1852, opened 2026-01-23
Adding ordered beta regression support to brms

saudiwin:
Hi Paul - 

As you might be aware, for some time I've had the ordbetareg package that defines the ordered beta model as a custom distribution for `brms`. I decided to finally try to implement it directly in `brms` as the `ordbeta()` family and I believe I have succeeded. The trick is that the ordered beta distribution is a mixed discrete-continuous distribution (like the ZOIB), but it uses ordered cutpoints like an ordinal model. Defining the ordered cutpoints directly as a distributional parameter proved to be very difficult due to `brms` having special functions for the `cumulative` family--most of which are not necessary as the ordered beta model only uses two cutpoints & defining effects over the cutpoints aren't the main point.

For these reasons, in the implementation in this PR I define the cutpoints implicitly as two distributional parameters--`xi` and `kappa`--that are then passed to the likelihood function. I chose these two because `xi` is unbounded by default and `kappa` is strictly positive--I can then define an implicit cutpoint vector by `[xi, xi + kappa]`.

This approach seems to work well & does not involve any additional changes to the code base. I have included two test scripts, `ordbeta_example.R` and `ordbeta_example_zoi_kappa.R` that show parameter recovery & convergence. Auxiliary functions also seem to work fine.

The only thing that is missing is that in my original model I use an induced dirichlet prior for the cutpoints (a la Betancourt), which is cool although it's not necessary & would require some kind of explicit definition of the cutpoints (though perhaps that could be calculated within the likelihood). In any case, I think this parameterization would meet most user needs.

I understand this parameterization is a bit unusual, and I'm happy to make any changes or additions. Thanks for considering this PR! A specific summary of changes is below.

### New Files
- **`inst/chunks/fun_ordbeta.stan`**: Stan functions implementing the ordered beta log-PDF with three overloaded versions (scalar, vector/scalar phi, fully vectorized)

### Modified Files

| File | Changes |
|------|---------|
| `R/distributions.R` | Added `dordbeta()`, `pordbeta()`, `qordbeta()`, `rordbeta()` distribution functions |
| `R/families.R` | Added `ordbeta()` family constructor with documentation |
| `R/family-lists.R` | Defined family specification (dpars, links, bounds, priors) |
| `R/log_lik.R` | Added `log_lik_ordbeta()` for LOO/WAIC computation |
| `R/posterior_epred.R` | Added `posterior_epred_ordbeta()` for expected predictions |
| `R/posterior_predict.R` | Added `posterior_predict_ordbeta()` for posterior predictive samples |
| `R/stan-likelihood.R` | Added ordbeta to likelihood generation |
| `R/stan-response.R` | Minor adjustment for response handling |
| `NAMESPACE` | Exported new functions |

### DIFF
--- a/R/families.R
+++ b/R/families.R
@@ -1643,6 +1643,10 @@ is_cox <- function(family) {
   "cox" %in% family_info(family, "specials")
 }
 
+is_ordbeta <- function(family) {
+  "ordbeta" %in% family_info(family, "specials")
+}
+
 # has joint link function over multiple inputs
 has_joint_link <- function(family) {
   "joint_link" %in% family_info(family, "specials")
--- a/R/family-lists.R
+++ b/R/family-lists.R
@@ -114,7 +114,7 @@
     }  
       NULL 
     },
-    specials = character(0),
+    specials = c("ordbeta"),
     normalized = ""
   )
 }


paul-buerkner:
Thank you for this PR! I will need a bit of time to review the PR but I have it on my list of the next release so it won't forgotten.

saudiwin:
Great, thanks!
All best - 

Bob

PR #1856, opened 2026-02-19
A few minor updates to the function docs

jdblischak:
This PR includes a few minor fixes to the docs that I noticed while reading through them recently.

First, I noticed the typo `dpi` instead of `doi`. While fixing this, I updated the formatting of the citation to match the `@references` section in `?brm` (ie use `\emph{}` and `\code{}`).

https://github.com/paul-buerkner/brms/blob/2ba68d4056b2198d0884c14c643522151f225610/R/brm.R#L311-L314

Second, the documentation for the parameters `ub` and `lb` for `set_prior()` said that the default value was `NULL`, but this was changed to `NA` back in 2018 in https://github.com/paul-buerkner/brms/commit/07128709c93f2a99d235c76129af238b1575aa17.

### DIFF
--- a/R/priors.R
+++ b/R/priors.R
@@ -17,9 +17,9 @@
 #' @param nlpar Name of a non-linear parameter.
 #'   Only used in non-linear models.
 #' @param lb Lower bound for parameter restriction. Currently only allowed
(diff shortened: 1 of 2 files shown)


paul-buerkner:
Thanks!

Issue #1860, opened 2026-03-02
set 'silent' argument through options()


LucianoDeBenedictis:
I am aware that the `silent` argument for `brm()` has evolved through time, and as of now it's a great tool to control the verbosity of the model fitting. I have checked previous issues, and I couldn't find one with this same suggestion:

- make it possible to set `silent` globally through `options()`

I have also checked the source for `brm()` and it doesn't seem to be getting an option for the argument. Hoping this is simple to achieve, it would be handy to set it once for all the following fits. Personally, I would use it to set it at 2, because I find the sampling updates to be a bit cluttered (using `cmdstanr`, setting silent to 2 seems enough), especially for notebooks.

Thanks for considering this!

paul-buerkner:
That's a great idea! Would you mind preparing a PR?

LucianoDeBenedictis:
> That's a great idea! Would you mind preparing a PR?

I will try my best!

PR #1861, opened 2026-03-02
Add "brms.silent" option

LucianoDeBenedictis:
Following #1860 , brm() and brm_multiple() now support the global option "brms.silent".
Argument default for "silent" in  brm() and brm_multiple() changed to silent = getOption("brms.silent", 1) and updated documentation.
`devtools::test()` mostly passes, except for a few errors of this kind, which I think are unrelated:

> Error (tests.brmsfit-methods.R:456:3): loo has reasonable outputs
> Error in `r_eff_helper(lik_fun, chain_id = chain_id, draws = draws, allow_na = allow_na, ...)`: anyNA() applied to non-(list or vector) of type 'closure'

I've given it a try with

```
library(brms)

options("brms.silent" = 2)

data <- data.frame(x= rnorm(100),
                   y = rnorm(100))

fit <- brm(y ~ x, 
           data = data,
           prior = prior(normal(0,1), class = "b"),
           backend = "cmdstanr")
```
and also with rstan, although `refresh` is more useful there.

### DIFF
--- a/R/brm.R
+++ b/R/brm.R
@@ -196,7 +196,8 @@
 #'   If \code{2}, even more messages are suppressed. The actual
 #'   sampling progress is still printed. Set \code{refresh = 0} to turn this off
 #'   as well. If using \code{backend = "rstan"} you can also set
-#'   \code{open_progress = FALSE} to prevent opening additional progress bars. Can be set globally for the current \R session via the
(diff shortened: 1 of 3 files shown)


paul-buerkner:
Perfect, thank you!

PR #1863, opened 2026-03-09
Failure of `r_eff_helper` check in tests.brmsfit-methods.R

florence-bockting:
## Description
I was noticinv that the tests for `tests.brmsfit-methods.R` currently fail with the error message:
```
Error in `r_eff_helper(lik_fun, chain_id = chain_id, draws = draws, allow_na = allow_na, ...)`: anyNA() applied to non-(list or vector) of type 'closure'
```
After debugging I assume that the origin of the failure is that the function `r_eff_helper(x, chain_id, allow_na = TRUE, ...)` checks `(anyNA(x) || any(is.infinite(x))`. However, if `pointwise = TRUE` the argument x is a function in which case the check yields the above error message. 

## Proposed solution

Including a !is.function(x) in check to differentiate between different cases (x is function / x is matrix with posterior draws).

## TODOs
+ [x] modify check
+ [x] update function documentation
+ [x] ensure that checks pass now

### DIFF
--- a/R/loo.R
+++ b/R/loo.R
@@ -747,10 +747,7 @@ subset.psis <- function(x, subset, ...) {
 # @param allow_na allow NA values in the output?
 # @return numeric vector of relative efficiencies, length NCOL(x) or nobs for function
 r_eff_helper <- function(x, chain_id, allow_na = TRUE, ...) {
-  # if pointwise=TRUE, x is a function else it is a matrix of draws
-  x <- if (is.function(x)) list(...)$draws else x
-
-  if (anyNA(x) || any(is.infinite(x))) {
+  if (!is.function(x) && (anyNA(x) || any(is.infinite(x)))) {
     warning2("Ignoring relative efficiencies due to NA or infinte inputs.")
     return(rep(1, NCOL(x)))
   }


paul-buerkner:
Thanks for the PR! This warning message is misleading now. Shall there perhaps be a separate check and warning message that relative efficiencies will be ignored if pointwise = TRUE?

florence-bockting:
> Thanks for the PR! This warning message is misleading now. Shall there perhaps be a separate check and warning message that relative efficiencies will be ignored if pointwise = TRUE?

The second check does check for NA in `out`. Do you want a further check? Do you think it would be good to specify in the error message whether the argument "pointwise" was True or False to differentiate these cases?

```
r_eff_helper <- function(x, chain_id, allow_na = TRUE, ...) {
  if (!is.function(x) && (anyNA(x) || any(is.infinite(x)))) {
    warning2("Ignoring relative efficiencies due to NA or infinte inputs.")
    return(rep(1, NCOL(x)))
  }
  out <- loo::relative_eff(x, chain_id = chain_id, ...)
  if (!allow_na && anyNA(out)) {
    # avoid error in loo if some but not all r_effs are NA
    out <- rep(1, length(out))
    warning2(
      "Ignoring relative efficiencies as some were NA. ",
      "See argument 'r_eff' in ?loo::loo for more details."
    )
  }
  out
}
```

paul-buerkner:
Yes, sorry that is what I meant. I meant indeed a further check specific for the `is.function` (pointwise) case.

florence-bockting:
> Yes, sorry that is what I meant. I meant indeed a further check specific for the `is.function` (pointwise) case.

Would the following modification go into the right direction?

```
r_eff_helper <- function(x, chain_id, allow_na = TRUE, ...) {

  is_fun <- is.function(x)
  draws <- if (is_fun) list(...)$draws else x

  if (anyNA(draws) || any(is.infinite(draws))) {
    warning2(
      "Ignoring relative efficiencies due to NA or infinite inputs",
      if (is_fun) " for pointwise = TRUE." else " for pointwise = FALSE."
    )
    return(rep(1, NCOL(draws)))
  }

  out <- loo::relative_eff(x, chain_id = chain_id, ...)

  if (!allow_na && anyNA(out)) {
    # avoid error in loo if some but not all r_effs are NA
    out <- rep(1, length(out))
    warning2(
      "Ignoring relative efficiencies as some were NA. ",
      "See argument 'r_eff' in ?loo::loo for more details."
    )
  }
  out
}
```

paul-buerkner:
Ah, you even fixed the handling of the pointwise case entirely. Then, why not just go for:

```R
r_eff_helper <- function(x, chain_id, allow_na = TRUE, ...) {
  x <- if (is.function(x)) list(...)$draws else x

  if (anyNA(x) || any(is.infinite(x))) {
    warning2("Ignoring relative efficiencies due to NA or infinite inputs.")
    return(rep(1, NCOL(x)))
  }

  out <- loo::relative_eff(x, chain_id = chain_id, ...)

  if (!allow_na && anyNA(out)) {
    # avoid error in loo if some but not all r_effs are NA
    out <- rep(1, length(out))
    warning2(
      "Ignoring relative efficiencies as some were NA. ",
      "See argument 'r_eff' in ?loo::loo for more details."
    )
  }
  out
}
```

Essentially changing nothing but to correctly handle the is.function case via 

```R
# x is a function for pointwise = TRUE
x <- if (is.function(x)) list(...)$draws else x
```

florence-bockting:
Unfortunately, it was not as easy as I thought. Even though we pass a "draws" object, I realised that it is not the same as the expected posterior draws matrix when x is not a function. So, I returned to the case where we exclude x from the check when it is a function and it is currently for me also not clear how we can test the function (thus, include an additional check in case x is a function).

paul-buerkner:
Okay. Fine by me. I will merge it tomorrow.

PR #1866, opened 2026-03-11
replace model_params with variables

### LINKED ISSUE #1864: `rename_pars()` silently drops population-level coefficients with cmdstanr 0.9.0.9000 when `metadata$model_params` is NULL
**Bug description**

When fitting a model with 3 or more population-level predictors using `backend = "cmdstanr"` and cmdstanr 0.9.0.9000, brms silently drops coefficients from `fixef()` output. 

No warnings or errors are raised. The same model fitted with `backend = "rstan"` or directly via `cmdstanr::cmdstan_model()$sample()` returns all parameters correctly.

**Cause**

In `read_csv_as_stanfit()`, parameter dimensions are determined by grepping `csfit$metadata$model_params`:

```r
pdims_num <- ulapply(model_pars, function(x)
  sum(grepl(paste0("^", x, "\\[.*\\]$"), csfit$metadata$model_params))
)
par_dims[pdims_num != 0] <-
  csfit$metadata$stan_variable_sizes[model_pars][pdims_num != 0]
```

In cmdstanr 0.9.0.9000, `metadata$model_params` has been deprecated (in favor of `stan_variables`) and is `NULL`. Grepping against `NULL` returns `logical(0)`, so `pdims_num` evaluates to `0` for every parameter. The assignment on the next line never executes, and `par_dims[["b"]]` remains at its initialized value of `integer(0)` (scalar).

This creates an internal inconsistency in the stanfit object: `fnames_oi` correctly lists `b[1]` through `b[K]` (from the CSV column names), but `dims_oi[["b"]]` says `b` is a scalar. When `rename_pars()` subsequently reorders and renames parameters using `dims_oi`, parameters are silently dropped.

**Minimal reproducible example**

```r
library(brms)

set.seed(42)
n <- 200
dat <- data.frame(
  y  = rnorm(n),
  x1 = rnorm(n), x2 = rnorm(n), x3 = rnorm(n)
)

# cmdstanr backend — silently drops Intercept
fit_cmd <- brm(
  y ~ x1 + x2 + x3,
  data = dat, family = gaussian(),
  chains = 1, iter = 500,
  backend = "cmdstanr"
)
fixef(fit_cmd)
#>     Estimate  Est.Error  Q2.5  Q97.5
#> x1  ...       ...        ...   ...
#> x3  ...       ...        ...   ...
#> x2  ...       ...        ...   ...
## Only 3 rows — Intercept is missing (expected 4)

# Verify dims_oi is wrong:
fit_cmd$fit@sim$dims_oi$b
#> integer(0)
## Should be 3 (number of non-intercept coefficients)

# rstan backend — all parameters present
fit_rst <- brm(
  y ~ x1 + x2 + x3,
  data = dat, family = gaussian(),
  chains = 1, iter = 500,
  backend = "rstan"
)
fixef(fit_rst)
#>            Estimate  Est.Error  Q2.5  Q97.5
#> Intercept  ...       ...        ...   ...
#> x1         ...       ...        ...   ...
#> x2         ...       ...        ...   ...
#> x3         ...       ...        ...   ...
## All 4 rows present
```

The number of dropped coefficients increases with model size:

| Non-intercept predictors | Expected `fixef()` rows | Actual | Lost |
|--------------------------|------------------------|--------|------|
| 1 | 2 | 2 | 0 (scalar `b` coincidentally correct) |
| 2 | 3 | 3 | 0 (latent bug: `dims_oi$b` is wrong but no visible loss) |
| 3 | 4 | 3 | 1 (Intercept) |
| 5 | 6 | 5 | 1 (Intercept) |
| 8 | 9 | 6 | 3 (Intercept, x7, x8) |

The set of dropped parameters is deterministic for a given model formula (verified across different seeds and data).

**Diagnostic trace**

| Layer | b parameters | Status |
|-------|-------------|--------|
| CmdStan CSV output | b[1]..b[K] | All present |
| `cmdstanr::read_cmdstan_csv()` draws | b[1]..b[K] | All present |
| `brms::read_csv_as_stanfit()` `fnames_oi` | b[1]..b[K] | All present |
| `brms::read_csv_as_stanfit()` `dims_oi["b"]` | `integer(0)` | **Wrong — should be K** |
| `csfit$metadata$model_params` | `NULL` | **Root cause — deprecated in cmdstanr 0.9.0.9000** |
| After `rename_pars()` → `fixef()` | Subset of K | **Parameters silently lost** |

**Suggested fix**

In `read_csv_as_stanfit()`, fall back to `stan_variable_sizes` when `model_params` is `NULL`:

```r
if (is.null(csfit$metadata$model_params)) {
  sizes <- csfit$metadata$stan_variable_sizes[model_pars]
  for (p in model_pars) {
    s <- sizes[[p]]
    if (!is.null(s) && prod(s) > 1) {
      par_dims[[p]] <- s
    }
  }
} else {
  pdims_num <- ulapply(model_pars, function(x)
    sum(grepl(paste0("^", x, "\\[.*\\]$"), csfit$metadata$model_params))
  )
  par_dims[pdims_num != 0] <-
    csfit$metadata$stan_variable_sizes[model_pars][pdims_num != 0]
}
```

Alternatively, use `stan_variable_sizes` directly — it is the authoritative source and is present in both the stable and dev versions of cmdstanr:

```r
sizes <- csfit$metadata$stan_variable_sizes[model_pars]
for (p in model_pars) {
  s <- sizes[[p]]
  if (!is.null(s) && prod(s) > 1) {
    par_dims[[p]] <- s
  }
}
```

A companion issue will be filed on https://github.com/stan-dev/cmdstanr/issues noting that removing `model_params` from the metadata can break downstream consumers.

**Session info**

```
R:         4.5.3
brms:      2.23.1
cmdstanr:  0.9.0.9000 (dev, SHA a8adbb1)
CmdStan:   2.38.0
posterior:  1.6.1.9000
rstan:     2.36.0.9000
OS:        macOS (Apple Silicon, Darwin 25.3.0)
```

**Impact**

This bug is **silent** — brms produces a valid-looking model object with no warnings or errors, but population-level coefficients are absent from the posterior. Any user running brms with `backend = "cmdstanr"` on the dev version (0.9.0.9000) and fitting a model with 3+ population-level predictors is affected. Models with 1–2 non-intercept predictors have the latent bug (incorrect `dims_oi`) but do not show visible parameter loss.

jgabry:
Closes #1864 

cmdstanr now recommends using `variables` instead of `model_params`. I didn't add any new tests, assuming (hoping) that at least one existing test would fail if this breaks something. But it shouldn't break anything since they are identical in cmdstanr: 

https://github.com/stan-dev/cmdstanr/blob/0c31c1b628ad9106a1349a53a9114b7dfe7f44e2/R/csv.R#L331

In the future we will remove `model_params` from cmdstanr so this PR will ensure that nothing is broken when that happens.

### DIFF
--- a/R/backends.R
+++ b/R/backends.R
@@ -767,7 +767,7 @@ read_csv_as_stanfit <- function(files, variables = NULL, sampler_diagnostics = N
   names(par_dims) <- model_pars
   par_dims <- lapply(par_dims, function(x) integer(0))
   pdims_num <- ulapply(model_pars, function(x)
-    sum(grepl(paste0("^", x, "\\[.*\\]$"), csfit$metadata$model_params))
+    sum(grepl(paste0("^", x, "\\[.*\\]$"), csfit$metadata$variables))
   )
   par_dims[pdims_num != 0] <-
     csfit$metadata$stan_variable_sizes[model_pars][pdims_num != 0]


paul-buerkner:
Thank you!

PR #1878, opened 2026-03-31
kfold(): validate non-empty Ksub (fixes #1877)

### LINKED ISSUE #1877: kfold(): Ksub = integer(0) leads to invalid behavior
Hi, 

Thanks for the great package!

I noticed that `Ksub = integer(0)` passes validation in `kfold()` but results in no fold being evaluated, which leads to downstream errors:

```
Error in apply(lppds, 2, posterior::pareto_khat, are_log_weights = TRUE) : 
  dim(X) must have a positive length
```

It may be worth adding a check to ensure `Ksub` is non-empty.

edelweiss611428:
This PR addresses #1877  by adding a validation check for `Ksub` in `kfold()`.

Currently, `Ksub = integer(0)` passes validation but results in no folds being evaluated, leading to downstream errors or inconsistent output. This PR ensures that `Ksub` must be non-empty.

This is a minimal change and does not affect existing valid usage.


paul-buerkner:
Thanks!

Issue #1815, opened 2025-08-11
bayes_R2 computation does not match Bayesian R-squared paper eq (3)


avehtari:
"R-squared for Bayesian Regression Models" paper at https://doi.org/10.1080/00031305.2018.1549100 titles Equation (3) as "Bayesian R^2". 

brms uses the following implementation
```
.bayes_R2 <- function(y, ypred, ...) {
  e <- -1 * sweep(ypred, 2, y)
  var_ypred <- matrixStats::rowVars(ypred)
  var_e <- matrixStats::rowVars(e)
  as.matrix(var_ypred / (var_ypred + var_e))
}
```
which matches Equation (2) titled as "alternative R^2". This could be also called as "residual R^2".

The code for the paper is at https://avehtari.github.io/bayes_R2/bayes_R2.html. Modifying the bayes_R2 code for brms would look like
```
bayes_R2 <- function(fit) {
  mupred <- brms::posterior_epred(fit)
  var_mupred <- apply(mupred, 1, var)
  sigma2 <- as.matrix(fit, variable = c("sigma"))^2
  var_mupred / (var_mupred + sigma2)
}
```

The important part of Bayes-R2 is that it doesn't depend on y. 

The biggest difference with the current implementation and actual Bayes-R^2 can be seen when using sampling from the prior with `sample_prior = "only"`. The above fixed bayes_R2 code correctly returns draws from the implied prior on R^2. 

When sampling from the posterior there are differences for small n, and with big enough n, the differences get smaller (see examples at  https://avehtari.github.io/bayes_R2/bayes_R2.html)

I suggest
- change bayes_r2.brmsfit to match the paper on normal and Bernoulli, for Bernoulli use Tjur's pseudo-variance
- for other families which do not yet have pseudo-variance, give a warning and fall back to residual-R^2
- for other families, add pseudo-variance when adding pseudo-variance support for R2D2 type priors

The benefit of using Bayes-R^2 as in the paper is that we can get the implied prior on R^2 for any priors on coefficients, and in case of R2D2 type priors the implied priors should match the beta prior used.

paul-buerkner:
Thank you, Aki! I will think about how to best change the implementation, but your suggestions make sense to me.

PR #1880, opened 2026-04-18
Fix Bayesian R2 computation for normal and Bernoulli

florence-bockting:
## Description
This PR refers to [Issue#1815](https://github.com/paul-buerkner/brms/issues/1815).
Update `bayes_R2()` to use model-based residual variance for supported families (Gaussian and Bernoulli), while preserving fallback behavior (i.e., residual-based R2) for unsupported families.

Note, changes in this PR might be **breaking** as the output of the plot might change. However, the change in computation is rather considered as a "bug" and this PR as a **fix**.  For safety reasons I run a reverse dependency check for this PR which indicated no issues with dependent packages.

## Changes
- Refactored `bayes_R2.brmsfit()` to compute R2 via model-based residual variance for:
  - Gaussian models (`E[sigma^2]` per draw)
  - Bernoulli models (Tjur pseudo-variance: `E[pi * (1 - pi)]` per draw)
- Added explicit fallback to residual-based computation for other families.
- Added a warning when fallback is used, emitted once per family.
- Split internals into helper functions for clarity:
  - `.bayes_R2_res()`
  - `.bayes_R2_model()`
  - `.bayes_R2_var_res_gaussian()`
  - `.bayes_R2_var_res_bernoulli()`

## TODO
- [x] update function documentation
- [x] update tests
- [x] add release note in `NEWS.md` 
- [x] run reverse dependency check

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -19,6 +19,9 @@ Thanks to Sermet Pekin. (#1790)
 
 ### Bug Fixes
 
+* `bayes_R2` now uses model-based residual variances for Gaussian and Bernoulli 
+models and falls back to residual-based computation for other families. This 
+change may lead to changes in plotting results. (#1815)
 * Avoid the creation of zombie workers when executing `log_lik`
 in parallel thanks to Aki Vehtari and Noa Kallioinen. 
 For now, `log_lik` will use PSOCK clusters if run
@@ -28,7 +31,6 @@ These changes may be reverted once the underlying causes of this
 issue have been fixed. (#1658)
 * Align the definition of the R function `step()` with the definition in Stan,
 such that `step(0) == 1` thanks to Daniel Sabanes Bov. (#1734)
-* `bayes_R2` now uses model-based residual variances for Gaussian and Bernoulli models (using Tjur's pseudo-variance for Bernoulli), and warns before falling back to residual-based computation for other families, aligning `bayes_R2` with the Bayes-R2 definition in the reference paper for supported families. (#1815)
 * Make `read_csv_as_stanfit()` store `adapt_delta` and `max_treedepth` values in
 `$control` so rstan can find these values. Thanks to Tristan Mahr (#1767).
 * Enable updating argument `data2` for `brmsfit_multiple` objects. (#1776)
--- a/NEWS.md
+++ b/NEWS.md
@@ -28,6 +28,7 @@ These changes may be reverted once the underlying causes of this
 issue have been fixed. (#1658)
 * Align the definition of the R function `step()` with the definition in Stan,
 such that `step(0) == 1` thanks to Daniel Sabanes Bov. (#1734)
+* `bayes_R2` now uses model-based residual variances for Gaussian and Bernoulli models (using Tjur's pseudo-variance for Bernoulli), and warns before falling back to residual-based computation for other families, aligning `bayes_R2` with the Bayes-R2 definition in the reference paper for supported families. (#1815)
 * Make `read_csv_as_stanfit()` store `adapt_delta` and `max_treedepth` values in
 `$control` so rstan can find these values. Thanks to Tristan Mahr (#1767).
 * Enable updating argument `data2` for `brmsfit_multiple` objects. (#1776)
--- a/R/bayes_R2.R
+++ b/R/bayes_R2.R
@@ -15,10 +15,10 @@
 #'
 #' @details For an introduction to the approach, see Gelman et al. (2019)
 #'  and \url{https://github.com/jgabry/bayes_R2/}.
-#'  For Gaussian and Bernoulli models, \code{bayes_R2} uses model-based
-#'  residual variances as proposed in Gelman et al. (2019), with Bernoulli
-#'  models using Tjur's pseudo-variance. For other families, \code{bayes_R2}
-#'  warns and falls back to residual-based variances.
+#'  For Gaussian (homoscedastic sigma) and Bernoulli models, \code{bayes_R2} 
+#'  uses model-based residual variances as proposed in Gelman et al. (2019), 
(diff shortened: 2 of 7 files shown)


florence-bockting:
@avehtari can you check the current implementation?
Also: You mention in the Issue:
> for other families, add pseudo-variance when adding pseudo-variance support for R2D2 type priors

This part is not yet implemented as I read it as a task for future implementations but I am not sure whether I misread this point - therefore I want to point this explicitly out here.

florence-bockting:
@avehtari Can you please check the correctness of the computation. Thank you.

R/bayes_R2.R
```diff
@@ -93,9 +117,23 @@ bayes_R2.brmsfit <- function(object, resp = NULL, summary = TRUE,
...
 }
+
+.bayes_R2_model <- function(ypred, var_res, ...) {
+  var_ypred <- matrixStats::rowVars(ypred)
+  as.matrix(var_ypred / (var_ypred + var_res))
+}
+
+.bayes_R2_var_res_gaussian <- function(sigma, ...) {
+  sigma <- as.matrix(sigma)
+  matrixStats::rowMeans2(sigma^2)
```
avehtari: I don't understand why there is rowMeans. The paper and the online appendix provided equations only for scalar sigma, and I think just taking mean for vector sigma in heteroscedastic case is not correct. Even if more correct approach would be too complicated to implement, at least this approximation should be documented.

florence-bockting: Thank you for pointing this out. I decided now for the following approach: check whether sigma is heteroscedastic. If true,  a warning is provided and R2 computation falls back to residual-based R2.

florence-bockting: After discussing this point again, we (Aki, Florence) decided to use also for the heteroscedastic case  the mean of sigma as an approximation as this is consistent with the discussion provided in Tjur (2009). A corresponding note is added in the function documentation for the user as well as in the code as a comment for the developer.

To sum up, we use model-based variances for (homo- and heteroscedastic) Gaussian and Bernoulli models. Otherwise residual-based variances are used and the user warned.

R/bayes_R2.R
```diff
@@ -70,18 +80,41 @@ bayes_R2.brmsfit <- function(object, resp = NULL, summary = TRUE,
       "'bayes_R2' which is likely invalid for ordinal families."
     )
   }
-  args_y <- list(object, warn = TRUE, ...)
```
paul-buerkner, line 73: can we please add an argument allowing to switch between model and residual based version? Otherwise, users will always be annoyed by the warnings which are constantly thrown at them.

NEWS.md
```diff
@@ -19,6 +19,9 @@ Thanks to Sermet Pekin. (#1790)
 
 ### Bug Fixes
 
+* `bayes_R2` now uses model-based residual variances for Gaussian and Bernoulli
```
paul-buerkner, line 22: why is this a bug-fix? looks like "other changes" to me. Also why does it lead to changes only in "plotting results"?

R/bayes_R2.R
```diff
@@ -70,18 +80,41 @@ bayes_R2.brmsfit <- function(object, resp = NULL, summary = TRUE,
...
-    args_ypred$resp <- args_y$resp <- resp[i]
-    y <- do_call(get_y, args_y)
+    args_ypred$resp <- resp[i]
     ypred <- do_call(posterior_epred, args_ypred)
     if (is_ordinal(family(object, resp = resp[i]))) {
       ypred <- ordinal_probs_continuous(ypred)
     }
-    R2[[i]] <- .bayes_R2(y, ypred)
+    family_name <- family(object, resp = resp[i])$family
+    if (family_name %in% "gaussian") {
```
paul-buerkner: this pattern works for now. If we later on support the model-based approach for more than two families, we should provide separate functions per family, similar to how we do it with a lot of other post-processing funcftions.

paul-buerkner (changes requested):
Looks good overall. just a few minor comments

NEWS.md
avehtari: I think Bayes-R^2 is model based R^2, and the code was not implementing model based R^2 and thus it was not implementing Bayes-R^2, but it was named as bayes_R2, so if this is not a bug fix, it's a fix anyway.

paul-buerkner:
Okay, we can leave it there in NEWS :-)

Am Mi., 29. Apr. 2026 um 17:08 Uhr schrieb Aki Vehtari <
***@***.***>:

NEWS.md
florence-bockting: I mark this conversation now as resolved as I think we agreed all on the aspect of considering this as a "conceptual bug" and reporting it like this in the NEWS.md

R/bayes_R2.R
florence-bockting: I added an argument "method" that is by default NULL and takes otherwise the values "model" or "residual". If the value "residual" is set explicitly by the user no warning is provided.
I updated also the function documentation.

R/bayes_R2.R
florence-bockting: I refactored the code and outsourced the family-specific computations for better maintainability and easier future updating. Thank you for pointing this out.

paul-buerkner:
I refactored the code a little. Tests seem to pass but can you, @florence-bockting, please check if all is still well after my refactor? Once you give your okay, I will merge.

paul-buerkner (approved):

florence-bockting:
> I refactored the code a little. Tests seem to pass but can you, @florence-bockting, please check if all is still well after my refactor? Once you give your okay, I will merge.

Thank you @paul-buerkner. I went over your refactored code and to me it looks good. I also run it on an internal vignette that I created and everything works fine.

paul-buerkner:
Perfect, thank you! Merging now.

PR #1881, opened 2026-04-23
Update restructure.R to recompute basis if <2.23.0 to include boundary factor (L)

AngusMcLure:
Since (#1740) v2.23.0, brmsfit objects have started storing the boundary factor (`L`) in the `basis` slot.
This resolved issues for out-of-sample predictions for newer models (#1739), but post-processing functions like `fitted` failed when applied to <v2.23.0 models that do not have the `L` stored.

This PR updates `restructure` so that it recomputes the `basis` slot. There was already code in `restructure` to do this for models <v2.19.3 so the fix is just a matter of updating the version number so that this is done for anything <v2.23.0.

### DIFF
--- a/R/restructure.R
+++ b/R/restructure.R
@@ -262,13 +262,6 @@ restructure_v2 <- function(x) {
     # that are directly passed to the Stan backends for later reuse (#1373)
     x$stan_args <- list()
   }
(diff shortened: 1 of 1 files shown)


paul-buerkner (approved):
Looks good. Thank you!

PR #1883, opened 2026-04-27
Fix grouped-thresholds support for custom ordinal families

rpsychologist:
This PR fixes an error when a custom ordinal family is used with `thres(gr = ...)`, where the generated Stan code doesn't pass the `merged_Intercept` to the log PMF. This extends the fix in #1843                            
                                                                
Reprex
```r
library(brms)
 
dat <- data.frame(
  response = sample(1:5, 100, replace = TRUE),
  thres = factor(sample(1:4, 100, replace = TRUE))
)

custom_thres_family <- custom_family(
  "custom_thres_family",
  dpar = c("mu", "disc"),
  links = c("identity", "log"),
  type = "int",
  specials = "ordinal",
  threshold = "flexible"
)
                                                                                                                                           
fun_block <- "                                                                                                                             
  real acat_logit_lpmf(int y, real mu, real disc, vector thres) {                                                                        
    int nthres = num_elements(thres);                                                                                                      
    vector[nthres + 1] p = append_row(0, cumulative_sum(disc * (mu - thres)));                                                             
    return p[y] - log_sum_exp(p);           
  }                                                                                                                                        
  real custom_thres_family_merged_lpmf(                                                                        
      int y, real mu, real disc, vector thres, array[] int j) {                                                                            
    return acat_logit_lpmf(y | mu, disc, thres[j[1]:j[2]]);                                                                                
  }                                                                                                                                        
"   

make_stancode(
  bf(
    response | thres(gr = thres) ~ 1
  ),
  family = custom_thres_family,
  stanvar = stanvar(scode = fun_block, block = "functions"),
  data = dat
)

```
Which leads to this incorrect line
```bash
   -------------------------------------------------
    59:      vector[N] mu = rep_vector(0.0, N);
    60:      for (n in 1:N) {
    61:        target += custom_thres_family_lpmf(Y[n] | mu[n], disc, Intercept);
                                                                      ^
    62:      }
    63:    }
   -------------------------------------------------

Identifier 'Intercept' not in scope. Did you mean 'Intercept_1'?
```

After the patch we get the correct Stan code
```stan
target += custom_thres_family_merged_lpmf(Y[n] | mu[n], disc, merged_Intercept, Jthres[n]);
```

PR also includes a small fix so that `threshold = c("equidistant", "sum_to_zero")` is picked up in custom families.

Changes shouldn't cause any regressions and all tests pass.

PR #1884, opened 2026-05-13
Fix threading with custom ordinal families and grouped thresholds

rpsychologist:
My previous PR #1883 didn't pass `threads` to `stan_log_lik_advars`, which means the wrong index is used for `Jthres` inside `partial_log_lik_lpmf` when threading is enabled for custom ordinal families with grouped thresholds. 

Before this PR:
```stan
 for (n in 1:N) {
      int nn = n + start - 1;
      ptarget += custom_thres_family_merged_lpmf(Y[nn] | mu[n], disc, merged_Intercept, Jthres[n]);
 }
```

After the fix `Jthres` gets the correct `nn` indexing
```stan
for (n in 1:N) {
      int nn = n + start - 1;
      ptarget += custom_thres_family_merged_lpmf(Y[nn] | mu[n], disc, merged_Intercept, Jthres[nn]);
 }
```

### DIFF
--- a/R/stan-likelihood.R
+++ b/R/stan-likelihood.R
@@ -980,7 +980,9 @@ stan_log_lik_custom <- function(bterms, threads = NULL, ...) {
       }
       str_add(lpdf) <- "_merged"
       p$thres <- paste0("merged_Intercept", prefix)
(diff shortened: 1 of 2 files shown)


paul-buerkner:
Thanks!

PR #1890, opened 2026-06-03
Fix `kfold_predict()` for posterior predictions with more than 2 dimensions

### LINKED ISSUE #1889: `kfold_predict()` fails when posterior predictions have more than two dimensions (e.g. categorical)
## Description
`kfold_predict()` always allocates `yrep` as a matrix with dimensions draws x observations and assigns each fold with `yrep[, obs] <- ....` 
That works for families where `posterior_predict()` / `fitted()` return a 2D object, but fails when the prediction method returns an array with additional dimensions (e.g. category probabilities for categorical() models: draws x observations x categories).

## Steps to reproduce
```
library(brms)
data("penguins", package = "palmerpenguins")
penguins <- subset(penguins, stats::complete.cases(penguins))

fit <- brm(
  species ~ bill_length_mm + bill_depth_mm,
  data = penguins,
  family = categorical(),
  chains = 2,
  cores = 2,
  iter = 200,
  seed = 42
)

kfold1 <- kfold(fit, save_fits = TRUE)
kfold_predict(kfold1, method = "fitted")  # errors
```
## Problem
In `kfold_predict()`, `yrep` is created as:
```
yrep <- matrix(NA, nrow = ndraws, ncol = npredicted)
and filled with yrep[, obs_names] <- method(...)
```
which assumes 2D predictions.

## Proposed direction
Allocate `yrep` as an array whose dimensions are taken from the first fold’s prediction object, with the observation dimension set to the total number of predicted rows across folds.

florence-bockting:
Fixes #1889 

## Summary
+ `kfold_predict()` no longer assumes `yrep` is a draws x observations matrix; it allocates an array whose dimensions follow the chosen prediction method, with the observation axis spanning all out-of-fold rows.
+ Adds helper `.assign_yrep()` for multidimensional assignment into the observation dimension.
+ Updates `@return` documentation and extends local model tests "tests/local/tests.models-1.R".

## Problem
For families where `fitted()` / `posterior_epred()` return arrays with extra dimensions (e.g. categorical(): draws x observations x categories), `kfold_predict()` failed because it pre-allocated a matrix and used `yrep[, obs] <- ....`

## Suggested solution
+ Infer `dim(yrep)` from the first fold’s predictions and set the second dimension to the total number of predicted observations.

## Note
+ The suggested solution yields a change in the type of the output (i.e., from matrix to array)
+ might lead to breaking reverse dependencies. For example if tests like `is.matrix()` are used.

### DIFF
--- a/R/kfold.R
+++ b/R/kfold.R
@@ -476,35 +476,26 @@ kfold_predict <- function(x, method = "posterior_predict", resp = NULL, ...) {
   all_predicted <- sort(unlist(x$fits[, "predicted"]))
   npredicted <- length(all_predicted)
   
-  y <- rep(NA, npredicted)
-  names(y) <- as.character(all_predicted)
+  y <- setNames(rep(NA, npredicted), as.character(all_predicted))
   yrep <- NULL
 
   for (k in seq_rows(x$fits)) {
     fit_k <- x$fits[[k, "fit"]]
     predicted_k <- x$fits[[k, "predicted"]]
-
-    obs <- match(predicted_k, all_predicted)
+    # indexes the obs of the current fold within the total obs
+    obs_k <- match(predicted_k, all_predicted)
     newdata <- x$data[predicted_k, , drop = FALSE]
     
-    y[obs] <- get_y(fit_k, resp, newdata = newdata, ...)
+    y[obs_k] <- get_y(fit_k, resp, newdata = newdata, ...)
     
     yrep_k <- method(
       fit_k, newdata = newdata, resp = resp,
       allow_new_levels = TRUE, summary = FALSE, ...
     )
     
-    if (is.null(yrep)) {
-      yrep_dim <- dim(yrep_k)
-      yrep_dim[2] <- npredicted
-      yrep_dimnames <- dimnames(yrep_k)
-      if (is.null(yrep_dimnames)) {
-        yrep_dimnames <- vector("list", length(yrep_dim))
-      }
-      yrep_dimnames[[2]] <- as.character(all_predicted) 
-      yrep <- array(NA, dim = yrep_dim, dimnames = yrep_dimnames)
-    }
-    yrep <- .assign_yrep(yrep, obs, yrep_k)
+    yrep <- .update_yrep(
+      yrep, obs_k, yrep_k, npredicted, all_predicted
+    )
   }
   nlist(y, yrep)
 }
@@ -524,9 +515,23 @@ validate_joint <- function(joint) {
   match.arg(joint, options)
 }
 
-# internal function
-.assign_yrep <- function(yrep, obs, value) {
-    idx <- rep(list(TRUE), length(dim(yrep)))
-    idx[[2]] <- obs
-    do.call("[<-", c(list(yrep), idx, list(value = value)))
+# internal function for initializing the yrep array and filling it with
+# predictions of each fold.
+.update_yrep <- function(yrep, obs, value, nrep, all_pred) {
+  # On first fold: 
+  if (is.null(yrep)) {
+    # initialise as NA array with same dimensions as yrep_k (returned from method)
+    d <- dim(value)
+    # expand 2nd dimension (number of observations) to cover all obs across folds
+    d[2] <- nrep
+    dn <- dimnames(value) %||% vector("list", length(d))
+    dn[[2]] <- as.character(all_pred)
+    # create array with correct dimensions and provide names for dimensions
+    yrep <- array(NA, dim = d, dimnames = dn)
   }
+  idx <- rep(list(TRUE), length(dim(yrep)))
+  idx[[2]] <- obs
+  # equivalent to: list(yrep)[idx] <- value
+  # overwrites the NA entries of yrep with values from this fold (via idx)
+  do.call("[<-", c(list(yrep), idx, list(value = value)))
+}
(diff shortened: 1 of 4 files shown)


R/kfold.R
```diff
@@ -470,22 +473,38 @@ kfold_predict <- function(x, method = "posterior_predict", resp = NULL, ...) {
...
-    yrep[, obs_names] <- method(
+    
+    y[obs] <- get_y(fit_k, resp, newdata = newdata, ...)
+    
+    yrep_k <- method(
       fit_k, newdata = newdata, resp = resp,
       allow_new_levels = TRUE, summary = FALSE, ...
     )
+    
+    if (is.null(yrep)) {
```
paul-buerkner: could this part (within `if (is.null(yrep)) { }`) perhaps be integrated into the .assign_yrep function?

R/kfold.R
```diff
@@ -505,3 +524,9 @@ validate_joint <- function(joint) {
   match.arg(joint, options)
 }
 
+# internal function
+.assign_yrep <- function(yrep, obs, value) {
+    idx <- rep(list(TRUE), length(dim(yrep)))
```
paul-buerkner: I struggle to parse this code. could you add some comments on what happens here?

R/kfold.R
```diff
@@ -505,3 +524,9 @@ validate_joint <- function(joint) {
...
+    idx[[2]] <- obs
+    do.call("[<-", c(list(yrep), idx, list(value = value)))
+  }
```
paul-buerkner: add new line at the end of the file

R/kfold.R
```diff
@@ -470,22 +473,38 @@ kfold_predict <- function(x, method = "posterior_predict", resp = NULL, ...) {
...
   y <- rep(NA, npredicted)
-  yrep <- matrix(NA, nrow = ndraws, ncol = npredicted)
-  names(y) <- colnames(yrep) <- all_predicted
+  names(y) <- as.character(all_predicted)
+  yrep <- NULL
+
   for (k in seq_rows(x$fits)) {
     fit_k <- x$fits[[k, "fit"]]
     predicted_k <- x$fits[[k, "predicted"]]
-    obs_names <- as.character(predicted_k)
```
paul-buerkner, line 482: do you happen to have an idea why I used obs_names instead of obs previously? I don't quite remember. just trying to make sure that we don't overlook something that made using obs_names necessary/preferred somehow.

paul-buerkner (changes requested):
Thanks! Looks good. Just minor comments.

R/kfold.R
florence-bockting: I updated the internal function and included the assignment and initialisation into the helper function which makes the logic of  `kfold_predict` cleaner.

R/kfold.R
florence-bockting: I added comments and renamed also the helper function to make it more readable and the code logic clearer.

R/kfold.R
florence-bockting: Well, I added a new line at the end of the file, but I don't know whether this got through... hm... perhaps I have some automatic linter activated somewhere?

R/kfold.R
florence-bockting: I think in this case named or numeric indexing should be equivalent, but I don't know whether there might certain other differences between these two methods. I changed the code to `obs_names` just to be on the save side.

florence-bockting:
> Thanks! Looks good. Just minor comments.

Thank you for the review. I made changes according to your comments, let me know if you want to have any further adjustments.

R/kfold.R
```diff
@@ -470,22 +473,28 @@ kfold_predict <- function(x, method = "posterior_predict", resp = NULL, ...) {
   }
   method <- get(validate_pp_method(method), mode = "function")
   resp <- validate_resp(resp, x$fits[[1, "fit"]], multiple = FALSE)
-  all_predicted <- as.character(sort(unlist(x$fits[, "predicted"])))
+  all_predicted <- sort(unlist(x$fits[, "predicted"]))
   npredicted <- length(all_predicted)
-  ndraws <- ndraws(x$fits[[1, "fit"]])
-  y <- rep(NA, npredicted)
-  yrep <- matrix(NA, nrow = ndraws, ncol = npredicted)
-  names(y) <- colnames(yrep) <- all_predicted
```
paul-buerkner, line 478: I can see that y now has the names again. How about yrep? Does it's colnames have the names still? I am not sure I am seeing this in the new code

paul-buerkner (commented):
Looks good. Just one question about naming columns of yrep to match previous behavior. Once this is resolved, I will merge.

R/kfold.R
florence-bockting: I am not sure whether I understand your question. The assignment of names for `yrep` happens in `.update_yrep`. I paste here the output behaviour for transparency:

New output:
```
> str(y)
 Named int [1:333] 1 1 1 1 1 1 1 1 1 1 ...
 - attr(*, "names")= chr [1:333] "1" "2" "3" "4" ...

> str(yrep)
 num [1:200, 1:333, 1:3] 1 1 1 1 1 ...
 - attr(*, "dimnames")=List of 3
  ..$ : NULL
  ..$ : chr [1:333] "1" "2" "3" "4" ...
  ..$ : chr [1:3] "Adelie" "Chinstrap" "Gentoo"
```

vs. old output (different example):

```
> str(kfp1$y)
 Named num [1:236] 5 3 2 4 7 5 6 40 5 14 ...
 - attr(*, "names")= chr [1:236] "1" "2" "3" "4" ...
 
> str(kfp1$yrep)
 int [1:4000, 1:236] 2 2 2 5 1 1 4 2 0 3 ...
 - attr(*, "dimnames")=List of 2
  ..$ : NULL
  ..$ : chr [1:236] "1" "2" "3" "4" ...
```

paul-buerkner:
Thanks! Merging now!

PR #1892, opened 2026-07-05
implement `refcat = NA` for mixture families

### LINKED ISSUE #1450: Predicting all mixture proportions in mixture families, similar to refract = NA for ordinal or multinomial models.
Hi Paul,
I am currently working on implementing cognitive measurement models for memory tasks in brms. Specifically, a colleague and I have looked at mixture models for visual working memory tasks. To have more flexibility in implementing some models, it would be great if all mixture promotions could be freely estimated. In my mind, this should be similar to setting "refcat = NA" in multinomial or ordinal models. I know that such models require strong priors to be identified. This is something that we would take care of.

The main reason why we would need this functionality is that for some of the mixture models, the mixture proportions are further decomposed into other model parameters. Something like (not complete):
```
mixFormula <- bf(DV ~ 1,
nlf(theta1 ~ a + b  + c),
nlf(theta2 ~ a + b),
nlf(theta3 ~ a),
a ~ 1 + (1 | ID),
b ~ 1 + (1 | ID),
c ~ 1
)
```
in the brmsformula. Then some of the model parameters are predicted by independent variables. We could just fix _a_ to zero, but previous implementations have fixed _b_ or _c_ for scaling (e.g., fixing _c_ using a constant prior in this example). So to verify that our implementations recover previous implementations it would be great to be able to flexible set which parameter should be fixed for scaling. I see that this should not be a default setting, but it would still be great to call that option should you need it for specific models.

If you need more specific information, just say so, then I am happy to share more detailed code.
Thanks for all your great work.
Cheers, Gidon

GidonFrischkorn:
Closes #1450.

`refcat = NA` on a `mixture()` family lets all `K` mixing proportions be predicted (`softmax` over all components, no reference), instead of the current `K - 1`. Direct analogue of `refcat = NA` in `categorical()`. Default (`refcat = NULL`) is
unchanged.

## Why this exists

In cognitive measurement models the mixing proportions decompose into shared sub-parameters (`theta1 ~ a + b + c`, `theta2 ~ a + b`, `theta3 ~ a`). Forcing one `theta` to be the reference makes this inexpressible and blocks cross-validating against implementations that fix a *different* parameter for scaling. Predicting all `K` lets the user choose which parameter carries the constraint via priors rather than having brms impose it structurally.

## Decisions in this PR

- Opt-in is a family argument, not "just supply all K formulas." All-`K` formulas currently error on purpose; silently making them valid would drop a guardrail. `refcat = NA` makes the under-identified mode a deliberate choice.

- `refcat` here is binary, not a category selector. In `categorical()`, `refcat` names *which* level is the reference. Mixtures have no named categories — in the default mode the reference is whichever `theta<ID>` formula the user omits — so the only new choice is "reference or not." Hence `refcat` accepts only `NULL` or `NA`. The name is reused for cross-family consistency; happy to rename (`softmax_all`, `no_refcat`, …) if it reads too much like a selector.

- No changes to priors/standata/`get_theta()`. They already key off `has_joint_theta()` (true whenever any `theta` is predicted), so each predicted `theta` gets ordinary priors, no Dirichlet `con_theta` is emitted, and post-processing softmaxes whatever thetas exist.

- Identifiability is documented, not enforced at runtime. A free `softmax` is shift-invariant; the default `logistic(0, 1)` intercept priors give soft identification and informative priors are recommended. No runtime warning, this matches the silent `categorical(refcat = NA)`; the caveat lives in `?mixture` and `?brmsformula`.

## Targets for review

- The stancode test asserts on the `theta2 += Intercept_theta2 + ...` accumulation line, *not* on the absence of `rep_vector(0.0, N)` — the latter is the generic predictor initializer, present for every predicted parameter, so it doesn't
  distinguish "reference" from "predicted."
- Feedback welcome on the `refcat` name and on whether a runtime identifiability message would be worth its noise.

### DIFF
--- a/R/families.R
+++ b/R/families.R
@@ -982,7 +982,7 @@ mixture <- function(..., flist = NULL, nmix = 1, order = NULL, refcat = NULL) {
           "as the number of mixture components.")
   }
   if (!is.null(refcat) && !isNA(refcat)) {
-    stop2("For mixture models, 'refcat' can only be NA.")
+    stop2("For mixture models, 'refcat' can only be NULL or NA.")
   }
   dots <- dots[rep(seq_along(dots), nmix)]
   family <- list(
--- a/R/families.R
+++ b/R/families.R
@@ -881,6 +881,16 @@ acat <- function(link = "logit", link_disc = "log",
 #'   If \code{NULL} (the default), \code{order} is set to \code{'mu'}
 #'   if all families are the same and \code{'none'} otherwise.
 #'   Other ordering constraints may be implemented in the future.
+#' @param refcat Optional reference category for the mixing proportions.
+#'   By default (\code{NULL}), when the mixing proportions are predicted,
+#'   all but one of them have to be modeled while the remaining one serves
+#'   as the reference category (its linear predictor is fixed to zero) to
+#'   identify the model. If \code{refcat = NA}, no reference category is used
+#'   and all mixing proportions are predicted via a 'softmax' transformation.
+#'   This is analogous to \code{refcat = NA} in \code{\link{categorical}}
+#'   models. As the resulting model is only weakly identified, informative
+#'   priors on the \code{theta} parameters are strongly recommended (see the
+#'   examples in \code{\link{brmsformula}}).
 #'
 #' @return An object of class \code{mixfamily}.
 #'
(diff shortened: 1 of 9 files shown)


R/families.R
```diff
@@ -960,11 +981,15 @@ mixture <- function(..., flist = NULL, nmix = 1, order = NULL) {
     stop2("The length of 'nmix' should be the same ",
           "as the number of mixture components.")
   }
+  if (!is.null(refcat) && !isNA(refcat)) {
+    stop2("For mixture models, 'refcat' can only be NA.")
```
paul-buerkner: can only be NULL or NA?

paul-buerkner (changes requested):
Thank you! Looks good. I only have one minor change request

R/families.R
GidonFrischkorn: Yes, totally. I pushed the correct message.

paul-buerkner (approved):
Looks good now

GidonFrischkorn:
Ah, sorry.

I forgot about one test that needs to pick up on the adapted message.

Will fix this asap.

tests/testthat/tests.stancode.R
```diff
@@ -1858,7 +1858,7 @@ test_that("Stan code of mixture model is correct", {
   )
   expect_error(
     mixture(gaussian, gaussian, refcat = 2),
```
paul-buerkner: this test seems a bit uncessary, right? Given that is it in tests.stancode but only tests the funcationality of mixture, which is already tested almost identically elsewhere. Perhaps remove the test?

paul-buerkner (changes requested):
sorry hat one more comment

tests/testthat/tests.stancode.R
GidonFrischkorn: Agreed. I removed the redundant test.

paul-buerkner (approved):

paul-buerkner:
Thanks! Merging now.

Issue #1897, opened 2026-07-16
Cox regression and knot placement


wds15:
Hi!

I have been struggling a lot with Cox recently and had to learn the hard way that the brms default placement of knots for the m-spline used to fit these models places the knots based on the quantiles of the distribution of times given to model... which includes censoring and event times. This is counter to the convention (see for example Frank Harrells work on this) to use the quantiles of the **event** time distribution which excludes the censoring events. This did bite me hard as I work on data often where I have administrative censoring at, say, 1 year (end of study). This casts many censoring times to cluster at 1y which can make things as bad as having no knot inside the time interval when events actually happen (during the study).

In short: Can the knots be formed based on the event time distribution excluding the censoring times which is apparently quite common to do when looking at Franks works?

Best,
Sebastian

While working on this another urgent issue popped up, which is also addressed in the respective PR:

For `family = cox()` models, post-processing on **new data** (`standata(fit, newdata =)`,
`posterior_epred()`, `log_lik()`, `loo(..., newdata =)`, `kfold()`, ...) silently
**re-derived the knots of the M-spline baseline hazard from the new data** whenever the
model used a **non-linear formula**, e.g.

```r
bf(y | bhaz(df = 8) + cens(cens) ~ eta, eta ~ x, nl = TRUE, family = cox())
```

The linear case was fine — only the non-linear case was affected. Because the spline
basis then differed from the one used at fitting time, the fitted `sbhaz` coefficients
were combined with a *different* basis, so predictions and `log_lik`/`loo`/`kfold`
values on new data were subtly wrong (and depended on the composition of the new data,
e.g. on the number of rows passed in).

paul-buerkner:
Yes of course! Good catch! Would you have the time to change this and
submit a PR?

wds15 ***@***.***> schrieb am Do. 16. Juli 2026 um 21:38:

wds15:
I can look into that, sure.

paul-buerkner:
Once I am back from vacation I can also take a look

wds15 ***@***.***> schrieb am Do. 16. Juli 2026 um 21:52:

wds15:
I pointed Claude Opus at it... the implementation allows to recover the old behaviour...

wds15:
Closed with #1898

PR #1898, opened 2026-07-16
Base Cox baseline hazard knots on event times only

wds15:
The knots of the baseline hazard M-splines in Cox models were placed at quantiles of both event and censoring times. Place them at quantiles of the event times only, while keeping boundary knots spanning all times so the basis can still be evaluated at censoring times beyond the last event.

The previous behavior can be restored via
options(brms.cox_bhaz_all_times = TRUE).

Resolves issue #1897


R/brmsframe.R
```diff
@@ -70,6 +70,12 @@ brmsframe.btl <- function(x, data, frame = list(), basis = NULL, ...) {
   # frame_ functions (but not vice versa) and are thus evaluated first
   x$frame <- frame
   x$basis <- basis
+  if (is.null(basis)) {
```
paul-buerkner, line 73: can you explain why adding this is necessary?

R/brmsframe.R
```diff
@@ -426,7 +432,25 @@ frame_basis_bhaz <- function(x, data, ...) {
...
+    # behavior of using both event and censoring times
+    y_knots <- y
+    if (!getOption("brms.cox_bhaz_all_times", FALSE)) {
+      cens <- get_cens(x, data)
+      if (!is.null(cens)) {
+        y_events <- y[cens == 0]
+        # fall back to all times if there are no exact events, since the knots
+        # cannot be defined from an empty set of event times (e.g. fully
+        # interval- or left-censored data)
+        if (length(y_events)) {
```
paul-buerkner, line 446: should the alternative `!length(y_events)` result in a warning perhaps? and how long should y_events minimally be to safely compute knots?

paul-buerkner (changes requested):
Thank you! I have small questions before I can approve the PR.

R/brmsframe.R
wds15: I must admit that I am not yet 100% familiar with brms ways to compute these things internally, but I torturted Opus to summarise about the rationale... if you remove the if, then indeed one gets the old wrong knot points which have censoring in the set if points used to construct the basis:

The  if (is.null(basis))  block runs only on the initial model-building path ( standata() / make_standata() / brm()  on the original data), where  brmsframe()  is called without a  basis . On the prediction/newdata path a basis is always passed in (from object$basis ), so the block is skipped and the stored knots are reused. 
                                                                                                                                                                                                                                                                                                                                                                                                                                                                      It's needed because this PR moved the knot-selection policy (event-times-only) entirely into  frame_basis_bhaz().  data_bhaz()  is now a pure evaluator of a precomputed basis, and  bhaz_basis_matrix()  no longer knows about censoring. So the basis has to be built up front via  frame_basis_bhaz()  — that's what this block does. If it's removed,  data_bhaz()  falls back to building the basis itself from all response times, silently reverting to the old all-times-knots behaviour. I verified this: with the block removed,  make_standata()  reproduces the released (all-times)  Zbhaz  exactly, whereas with it the knots are event-only.                                                                                                                                                                                                                                                                                                                                                                                                                        

I chose the eager version deliberately: it builds the basis once via the same  frame_basis_bhaz()  that also produces the stored  object$basis , so fit-time and newdata bases are guaranteed identical and  data_bhaz()  never (re)defines knots. A lazy variant (have data_bhaz()  call  frame_basis_bhaz()  on first use) is equally feasible and I've confirmed it gives the same result — but it reintroduces "this function sometimes defines, sometimes evaluates" and a small risk of the fit-time and stored bases diverging. Happy to switch to the lazy form if you'd prefer.

R/brmsframe.R
wds15: Good idea. You need more events than df, I think. However, running a non-parametric method as Cox against super sparse data should be something users I would expect to be aware of.

wds15:
@paul-buerkner anything you need more input from my end on this PR?

paul-buerkner:
No. Just a long todo list. I will review start of next week. Sorry for
being slow.

wds15 ***@***.***> schrieb am Fr. 28. Aug. 2026 um 17:12:

R/brmsframe.R
paul-buerkner: The idea is that basis comes into brmsframe as fixed and we do so for all other purposes of basis. So I would prefer the same route being taken also here. Otherwise, the logic of brms postprocessing becomes confusing. Is there a problem of going the standard route of contructing and holding basis here? I am okay with making an exception if that is necessary but I didn't quite get it from your response if that is the case.

NEWS.md
```diff
@@ -19,6 +19,9 @@ Thanks to Sermet Pekin. (#1790)
 
 ### Bug Fixes
 
+* Cox models now place the knots of the baseline hazard based on the event
```
paul-buerkner, line 45: please add a thanks to note at the end of the comment.

paul-buerkner (changes requested):
Thank you. I am sorry it took me so long. Can you check and resolve my remaining comments? I am happy to merge afterwards.

R/brmsframe.R
wds15: Thanks for the explanation. Are those sort of things somewhere written down?

So, no special-case exception is required.

The eager

```r
if (is.null(basis))
```

block in `brmsframe.btl()` has been removed. The only change in `R/brmsframe.R` relative to upstream is that `frame_basis_bhaz()` is now a thin storage function.

The knot construction logic has been centralised in a new helper:

```r
bhaz_basis()
```

located in `R/data-response.R`.

This helper is used in both places that need access to the basis:

* `data_bhaz()`
  * Defines the basis at fit time (`basis = NULL`).

* `frame_basis_bhaz()`
  * Stores the basis.
  * Called with `warn = FALSE`, ensuring that `brm()` emits the warning only once.

For `newdata`, `data_bhaz()` never redefines knots. It always evaluates using the supplied basis.

As a result, fit-time and prediction-time behaviour remain identical because both paths execute the same deterministic basis-construction code. This is the same pattern already used for `sm`, `gp`, and `sp`.

The ad hoc compatibility option

```r
options(brms.cox_bhaz_all_times)
```

has been removed and replaced with the established version-gating mechanism

```r
require_old_default("2.23.1")
```

used elsewhere in brms.

wds15:
I noticed another issue with Cox which I addressed in the last two commits (happy to move to a separate PR):

The second issue concerned where the baseline-hazard basis was stored.

Previously it was stored in `frame_basis.btl()`, i.e. alongside the predictor-term representation. However, the baseline hazard depends only on the response and its response additions, not on the predictors.

This happened to work for ordinary linear formulas because:

* `brmsterms()` attaches `$family` and `$respform` only to the `mu` distributional parameter.
* For linear formulas, `mu` is represented by a `btl` object that happens to carry this information.

For `nl = TRUE` models the situation is different:

* Predictor terms are stored in `nlpars`.
* `nlpars` objects do not carry family information.
* Consequently `frame_basis_bhaz()` became a no-op.
* No basis was stored.

**As a result, every `newdata` evaluation silently reconstructed knots from the new data instead of reusing the fit-time basis.**

This led to two problems:

1. Fit-time and prediction-time bases could differ.
2. Small `newdata` sets could fail with

   ```text
   Need two distinct boundary knots
   ```

The bug was reachable through:

* `log_lik()`
* `loo()`
* `kfold()`

but not through:

* `predict()`

because prediction is currently disabled for Cox models.

### Fix

The basis is now stored at the `brmsterms` level, alongside objects such as `resp_levels`.

It is propagated through

```r
x$frame$basis
```

which is already available to both:

* `btl`
* `btnl`

objects.

As a consequence, `frame_basis.btnl()` remains simply

```r
list()
```

and no additional special-case logic is required.

NEWS.md
wds15: I should thank myself?

paul-buerkner: yes, exactly :-D

paul-buerkner (changes requested):
Looks almost good now. Can you bump the minor version number? Because you use require_old_default and that only works reliably if the version number changes along with the new feature.

paul-buerkner:
And thank you for the additional fix!

wds15:
Thanks! Bumped the version to 2.23.2 in `DESCRIPTION` and keyed the backwards-compatibility guard on it (`require_old_default("2.23.2")` in `bhaz_basis`). The guard previously used 2.23.1, which is the current development version and therefore also carries the old knot placement, so it could not have distinguished old from new models.

Also updated accordingly:
- `bhaz` docs (`R/formula-ad.R` + `man/addition-terms.Rd`): "brms versions 2.23.1 and earlier"
- `NEWS.md`: models fitted with brms 2.23.1 or earlier keep their original knots
- `tests/testthat/tests.standata.R`: the old-default test now sets `.brmsfit_version = "2.23.1"` so it pins the new cut-off

`tests.standata.R` and `tests.stancode.R` pass locally.

paul-buerkner:
Thank you! Confirmed that test failures are unrelated to this PR. Merging now :-)

Issue #1903, opened 2026-07-31
Truncated `log_lik` is off by one bin for discrete families


deeb01:
Truncated discrete `log_lik()` normalizes by `P(lb < Y <= ub)`; the fitted model uses `P(lb <= Y <= ub)`. The pmf doesn't sum to 1 and `loo()`/`waic()` disagree with the sampled model, with no `Inf` or `NaN` to signal it.

```r
prep <- structure(list(ndraws = 1L, nobs = 1L), class = "brmsprep")
prep$dpars <- list(mu = matrix(3, ncol = 1))
prep$data <- list(Y = 2, lb = 2, ub = 6)
sum(sapply(2:6, function(y) { prep$data$Y <- y; exp(brms:::log_lik_poisson(1, prep)) }))
#> [1] 1.412371
```

`stancode()` emits `lb - 1` for integer responses (`R/stan-likelihood.R:154`):

```stan
log_diff_exp(poisson_lcdf(ub[n] | mu[n]), poisson_lcdf(lb[n] - 1 | mu[n]))
```

`log_lik_truncate()` passes `lb` unmodified:

```r
ppois(6, 3) - ppois(1, 3)   #> 0.7673432  Stan, == sum(dpois(2:6, 3))
ppois(6, 3) - ppois(2, 3)   #> 0.5433014  log_lik
```

Affects poisson, negbinomial, binomial, geometric, beta_binomial, discrete_weibull, and the zero-inflated/hurdle counts. Continuous families are fine.

Found while looking at #1899, separate cause.

PR #1904, opened 2026-07-31
Improve numerical stability of truncated and censored `log_lik`

### LINKED ISSUE #1899: Improve numerical stability of truncated `log_lik`
## Summary
For truncated continuous families (e.g. `y | trunc(lb, ub) ~ ... with gaussian()`), pointwise `log_lik` can become `+Inf` or `NaN` when the latent mean is far outside `[lb, ub]` and sigma is small. Then `Phi(ub) - Phi(lb)` underflows to 0 and the truncation adjustment in `log_lik_truncate()` fails.

## Current behavior
`log_lik_truncate()` uses:
```
x - log_diff_exp(log_cdf_ub, log_cdf_lb)
```
with `log_diff_exp()` implemented as `log(exp(x) - exp(y))`. When the mass in `[lb, ub]` underflows, this yields `-Inf` or `NaN`, so truncated `log_lik` becomes `+Inf` / `NaN`.

## Proposal
1. Evaluate the truncation normalizer in a more stable way (e.g. complementary CDFs when both CDFs are near 1; a stabler `log_diff_exp`).
2. Optionally warn when truncated `log_lik` is non-finite, so users can check whether the mean/sd is compatible with `[lb, ub]`.

## Reproducible example
```
library(brms)

n_draws <- 20
prep <- structure(list(ndraws = n_draws, nobs = 1), class = "brmsprep")
prep$dpars <- list(
  mu = matrix(c(-0.39, rep(0.2, n_draws - 1)), ncol = 1),
  sigma = c(0.02, rep(0.3, n_draws - 1))
)
prep$data <- list(Y = 0.01, lb = 0, ub = 1)

ll <- brms:::log_lik_gaussian(1, prep)
ll[1]   # Inf
```

## Context
https://discourse.mc-stan.org/t/difficulty-adding-loo-criterion-to-longitudinal-truncated-normal-model/41447/12

deeb01:
Closes #1899.

## Summary

Truncated `log_lik` computes its normalizer as a difference of CDFs, which cancels when both bounds fall far into the same tail and returns `Inf` or `NaN`.

## Problem

With `S = 1 - F` the normalizer has two identical forms, `F(ub) - F(lb)` and `S(lb) - S(ub)`. A probability near 1 carries no relative precision while its complement does, so only one is usable in a given tail: at `F(lb) = 1 - 1e-85` the first cancels to `log(0)`, and deep in the lower tail `F` underflows to `0` so only the first survives.

## Suggested solution

A new internal `log_prob_interval()` selects the form per draw. (A) is well conditioned when `F(lb) <= 1/2` and (B) when `S(lb) < 1/2`, so the cases are exhaustive; the derivation is in the comment above the function. Interval censoring had the same defect at `log(cdf(rcens) - cdf(y))` and shares the helper — separable if you would rather keep this to truncation.

`log1m_exp()` and `log_diff_exp()` are aligned with their Stan definitions using the two-branch algorithm of Maechler (2012). They are brms's R mirrors of Stan built-ins, exposed to user code through `env_stan_functions()`, so matching Stan is the intent; same reasoning as the `step()` alignment in #1734.

`log_lik.brmsfit()` now warns on infinite values, per @florence-bockting's request. A cancelled normalizer surfaces as `+Inf` under truncation and as `-Inf` under censoring, since `log_lik_truncate()` subtracts it and `log_lik_censor()` returns it directly. `anyNA` catches neither, so `loo()` was silently receiving an infinite `elpd`. The check follows `r_eff_helper()` in using `is.infinite()`.

Two caveats on that warning. Like the existing `NA` warning beside it, it is not reached under `pointwise = TRUE`. And it has no test: it lives in a method needing a fitted model, so a case would belong in `tests/local/`, which I cannot run here. Glad to add one if you want it.

Against the exponential's closed-form normalizer, master returns `-Inf` from bounds `[50, 60]` outward and `NaN` at `[700, 800]`. This is accurate to machine epsilon at every scale, and identical to master wherever master is already correct.

## Note

This only reaches families whose CDF is itself accurate on the log scale. `pinv_gaussian`, `pfrechet`, `pexgaussian`, `pasym_laplace`, `pgen_extreme_value`, `pdiscrete_weibull` and the zero-inflated/hurdle CDFs compute `1 - F` before taking the log, so the tail is gone before `log_prob_interval()` sees it. A test pins their current behaviour so that fixing them later is a visible change.

Also unfixed: the same cancellation in the generated Stan code and in `posterior_epred_trunc_*`, and the discrete bound convention in #1903.

## TODO

+ [x] tail switch for truncation and interval censoring
+ [x] warning on infinite `log_lik`
+ [x] tests, including the families this does not reach
+ [ ] `log1m_exp()` inside the naive CDFs (follow-up)
+ [ ] Stan-side tail switch (follow-up)
+ [ ] `tests/local/` case for the new warning, if wanted


deeb01:
`R CMD check` locally, macOS with R 4.3.1:

```
Status: 3 WARNINGs, 2 NOTEs
[ FAIL 0 | WARN 2 | SKIP 15 | PASS 4341 ]
```

No errors. The warnings and notes all come from my setup rather than the branch — unbuilt vignettes and uninstalled Suggests — and none of them names a changed file. The two test warnings are `NaNs produced` in `tests.distributions_relations.R`, which reproduce on master.

R/log_lik.R
```diff
@@ -76,6 +76,14 @@ log_lik.brmsfit <- function(object, newdata = NULL, re_formula = NULL,
         "Alternatively, use 'newdata' to predict only complete cases."
       )
     }
+    if (any(is.infinite(log_lik))) {
+      warning2(
+        "Infinite values were found in the log-likelihood. Most commonly ",
+        "this happens in truncated or censored models when the bounds ",
+        "contain no probability mass under some posterior draws. ",
+        "See ?log_lik for more details."
```
florence-bockting: As there is no @details section in `log_lik` which informs about the issue, we should not point in the warnings to the function documentation. Also adding this information seems to be very technical and probably less of interest for most users. Therefore I would stay with diagnostic information only.

```suggestion
        "contain no probability mass under some posterior draws."
```

R/log_lik.R
```diff
@@ -1016,11 +1024,61 @@ log_lik_censor <- function(dist, args, i, prep) {
     x <- do_call(cdf, c(y, args, log.p = TRUE))
   } else if (cens == 2) {
     rcens <- prep$data$rcens[i]
-    x <- log(do_call(cdf, c(rcens, args)) - do_call(cdf, c(y, args)))
+    # interval censoring needs log[F(rcens) - F(y)], which cancels in the
+    # tails for the same reason truncation does; see log_prob_interval()
```
florence-bockting, line 1021: I would drop this comment as the code is explicit enough and developers would look in the description of `log_prob_interval` for further information anyway.
```suggestion
```

R/log_lik.R
```diff
@@ -1016,11 +1024,61 @@ log_lik_censor <- function(dist, args, i, prep) {
...
+  stopifnot(!is.null(lower) || !is.null(upper))
+  lcdf <- function(q) do_call(cdf, c(q, args, log.p = TRUE))
+  lccdf <- function(q) {
+    do_call(cdf, c(q, args, log.p = TRUE, lower.tail = FALSE))
+  }
+  # F(-Inf) = 0, so an absent lower bound always selects the lower-tail form
+  if (is.null(lower)) {
+    lcdf_upper <- lcdf(upper)
+    return(log_diff_exp(lcdf_upper, rep(-Inf, length(lcdf_upper))))
+  }
```
florence-bockting: This can also be written in the following form, which might be a bit more explicit 
```suggestion
  # one-sided intervals are just a CDF or CCDF evaluation
  if (is.null(lower)) {
    return(lcdf(upper))
  }
  if (is.null(upper)) {
    return(lccdf(lower))
  }
```

R/log_lik.R
```diff
@@ -1016,11 +1024,61 @@ log_lik_censor <- function(dist, args, i, prep) {
...
+  # seeded from a cdf value so that names and dim are carried through
+  out <- lcdf_lower
+  out[] <- NA_real_
+  if (length(upper_tail) < n) {
+    lower_tail <- setdiff(seq_len(n), upper_tail)
+    # F(Inf) = 1 for an absent upper bound
+    lcdf_upper <- if (is.null(upper)) rep(0, n) else lcdf(upper)
+    stopifnot(length(lcdf_upper) == n)
+    out[lower_tail] <-
+      log_diff_exp(lcdf_upper[lower_tail], lcdf_lower[lower_tail])
```
florence-bockting: We could write the following also a bit more explicit and show the different cases:
```suggestion
  out <- lcdf_lower
  out[] <- NA_real_
  upper_tail <- which(lcdf_lower > -log(2))
  lower_tail <- which(lcdf_lower <= -log(2))
  if (length(lower_tail)) {
    out[lower_tail] <- log_diff_exp(
      lcdf(upper)[lower_tail], lcdf_lower[lower_tail]
    )
```

R/log_lik.R
```diff
@@ -1016,11 +1024,61 @@ log_lik_censor <- function(dist, args, i, prep) {
...
+  }
+  if (length(upper_tail)) {
+    # S(Inf) = 0 for an absent upper bound
+    lccdf_upper <- if (is.null(upper)) rep(-Inf, n) else lccdf(upper)
+    out[upper_tail] <-
+      log_diff_exp(lccdf(lower)[upper_tail], lccdf_upper[upper_tail])
```
florence-bockting: And then following the changes from above, we could write:

```suggestion
    out[upper_tail] <- log_diff_exp(
      lccdf(lower)[upper_tail], lccdf(upper)[upper_tail]
    )
```

R/numeric-helpers.R
```diff
@@ -142,13 +142,39 @@ log1p_exp <- function(x) {
...
+# both round to the same double -- exactly what happens for truncation
+# bounds lying far in the same tail (#1899).
 log_diff_exp <- function(x, y) {
   stopifnot(length(x) == length(y))
-  ifelse(x > y, log(exp(x) - exp(y)), NaN)
+  out <- x + log1m_exp(y - x)
+  # as in Stan, equal arguments give log(0) = -Inf rather than NaN,
+  # while x < y remains NaN because the difference is negative
+  eq <- !is.na(x) & !is.na(y) & x == y & x < Inf
+  out[eq] <- -Inf
```
florence-bockting: Could be written as:

```suggestion
  out[which(x == y & x < Inf)] <- -Inf
```

R/numeric-helpers.R
```diff
@@ -142,13 +142,39 @@ log1p_exp <- function(x) {
...
-}
-
+  out <- x
+  out[] <- NaN
+  na <- is.na(x)
+  out[na] <- x[na]
+  # log(2) is the crossover point used by Stan; on either side of it the
+  # branch that is selected is the one that avoids cancellation
+  near_zero <- !na & x < 0 & x > -log(2)
+  far <- !na & x <= -log(2)
```
florence-bockting: Could also be written as:
```suggestion
  near_zero <- which(x < 0 & x > -log(2))
  far <- which(x <= -log(2))
```

R/numeric-helpers.R
```diff
@@ -142,13 +142,39 @@ log1p_exp <- function(x) {
...
+  out[near_zero] <- log(-expm1(x[near_zero]))
+  out[far] <- log1p(-exp(x[far]))
+  out
```
florence-bockting, line 156: And then we could add:
```suggestion
  out[far] <- log1p(-exp(x[far]))
  out[is.na(x)] <- x[is.na(x)]
  out
```

NEWS.md
```diff
@@ -27,6 +27,15 @@ Thanks to Gidon Frischkorn. (#1450)
 
 ### Bug Fixes
 
+* Improve the numerical stability of `log_lik` for truncated and 
+interval-censored models, which previously returned `Inf` or `NaN` whenever 
+both bounds fell far into the same tail. Families whose CDF is not itself 
+accurate on the log scale are not yet covered. 
+Thanks to Florence Bockting. (#1899)
```
florence-bockting: Please write here your name as this PR is your contribution =)
```suggestion
Thanks to [YOUR NAME]. (#1899)
```

florence-bockting (commented):
Hi @deeb01, thank you very much for your PR. It is in really good shape and a very great contribution to the package. 
I have made only some minor comments. Nothing which is really a blocker, mainly some design ideas, but nothing you "have" to accept. 
I like very much that you put a lot of comments in the code, as it helps to review your PR, but I would remove some of the comments which are not highly important in order to understand the code (just for reason of maintainability and keeping the code base clean).
Please indicate the `News.md` your name as it is your contribution.
And probably you want to add a test for the new warning as well.

Once you went over the comments, I think the PR is ready for being approved. I will then ping Paul for the final "Go".

florence-bockting:
## Overview numerical stability of truncated `log_lik`

### Problem description

For a continuous response truncated to $[L, U]$,

```math
\ell_{\mathrm{trunc}}(y) = \log p(y \mid \theta) - \log\bigl(F(U) - F(L)\bigr).
```

Often coded as 
```math 
x - \text{log\_diff\_exp}(\log F(U), \log F(L)) \quad \text{with} \quad \text{log\_diff\_exp}(a, b) = \log(e^{a} - e^{b}).
```

When $F(U) - F(L)$ is tiny (or both CDFs sit at the same extreme of $[0,1]$), double precision loses the difference, so the normalizer becomes $-\infty$ or `NaN` and $\ell_{\mathrm{trunc}}$ becomes $+\infty$ or `NaN`.

- **Upper tail ($F \approx 1$).** Both $F(L)$ and $F(U)$ round to $1$, so $\log F = 0$ and $\log(e^0 - e^0) = -\infty$.
- **Lower tail ($F \approx 0$).** Both CDFs underflow to $0$, so the naive difference underflows to $-\infty$.

### Implemented solution

Compute the interval probability in log space with a stable difference, evaluated in the tail that is not saturated:

```math
\text{log\_diff\_exp}(a, b) = a + \text{log1m\_exp}(b - a), \qquad a \ge b,
```

and use $F(U) - F(L) = S(L) - S(U)$ with $S = 1 - F $:

```math
\log\bigl(F(U) - F(L)\bigr) =
\begin{cases}
\text{log\_diff\_exp}(\log F(U),\, \log F(L)) & \text{if } \log F(L) \le -\log 2,\\
\text{log\_diff\_exp}(\log S(L),\, \log S(U)) & \text{if } \log F(L) > -\log 2.
\end{cases}
```

Stable `log_diff_exp` handles lower-tail underflow (the relative gap $e^{b-a}$ remains representable). Switching to the survival side handles upper-tail $F \approx 1$ saturation. Edge cases (Stan): $a = b < \infty \Rightarrow -\infty$; $a < b \Rightarrow$ `NaN`.

One-sided truncation is the special case: lower-only $\Rightarrow \mathrm{lccdf}(L)$; upper-only $\Rightarrow \mathrm{lcdf}(U)$. The CDF must expose a genuine log-survival (`log.p = TRUE`, `lower.tail = FALSE`); computing $\log(1 - F(\cdot))$ after the fact reintroduces the same cancellation.

@deeb01 would you agree with this short overview? I want just a short "formal" overview in this thread which might help our memory whenever we want to come back to this point in future.

deeb01:
Thanks — all applied, name added to NEWS.

One deviation: I pulled the check into `warn_infinite_log_lik()` so it's testable without a fitted model. `brmsfit_example1` can't be given `trunc()` without losing its group-level coefficients. The test covers both signs — truncation subtracts the normalizer and gives `+Inf`, censoring returns it and gives `-Inf`.

The Windows failure isn't this branch: `unable to load shared object 'rstan.dll'` on the runner, also failing on #1898, #1905 and `add-citation-cff`.

florence-bockting:
> Thanks — all applied, name added to NEWS.
> 
> One deviation: I pulled the check into `warn_infinite_log_lik()` so it's testable without a fitted model. `brmsfit_example1` can't be given `trunc()` without losing its group-level coefficients. The test covers both signs — truncation subtracts the normalizer and gives `+Inf`, censoring returns it and gives `-Inf`.
> 
> The Windows failure isn't this branch: `unable to load shared object 'rstan.dll'` on the runner, also failing on #1898, #1905 and `add-citation-cff`.

Thanks for the changes.
There are two minor things then I think its good to go (@paul-buerkner)
- Please remove the unrelated but committed file: `brms_2.23.1.tar.gz`
- For the warning-wrapper `warn_infinite_log_lik`: Can you please ensure that the warning output does not exceed 80 characters to avoid flooding the terminal when the warning is provided. You can do that by using `\n` (for next line) in the text.

deeb01:
Both done — tarball removed, warning wrapped to 63/60/55 characters.

The overview matches the implementation. One sign slip: lower-only reads $-\mathrm{lccdf}(L)$ against $\mathrm{lcdf}(U)$ for upper-only. Both are the log normalizer, so neither takes the minus there; it belongs where the normalizer is subtracted from $\log p(y \mid \theta)$. The generated Stan carries the minus on both one-sided branches, which is probably where it crept in.

`R CMD check` locally: 0 errors, 4346 passing.

florence-bockting:
> Both done — tarball removed, warning wrapped to 63/60/55 characters.
> 
> The overview matches the implementation. One sign slip: lower-only reads − lccdf ( L ) against lcdf ( U ) for upper-only. Both are the log normalizer, so neither takes the minus there; it belongs where the normalizer is subtracted from log ⁡ p ( y ∣ θ ) . The generated Stan carries the minus on both one-sided branches, which is probably where it crept in.
> 
> `R CMD check` locally: 0 errors, 4346 passing.

Perfect, thank you very much and also thanks for pointing out the typo in the description. I changed it.
I think, we can merge now @paul-buerkner. Thank you @deeb01 for the contribution!

paul-buerkner:
Thank you @deeb01 for the PR and @florence-bockting for reviewing!

PR #1905, opened 2026-08-03
Implement group level mixtures

### LINKED ISSUE #1659: feature request: mixtures over groups / participants
Hi Paul!

We discussed about the possibility to specify mixture models in brms over grouping variables. Right now, the default is that mixtures are calculated on the level of observations, but sometimes we encounter use cases in which we want to specify different possible models for an entire sequence of data points (like the data of one participant), and fit a mixture over these models (basically what is described in the Stan-Manual as the "erroneous way" of vectorzing mixtures: https://mc-stan.org/docs/2_19/stan-users-guide/vectorizing-mixtures.html; also see the math in the attached picture why mixtures over single observations and over sequences of observations lead to different likelihoods). 

For implementing this in `brms`, we discussed that one could add a new "mix" argument in the `aterms` of a `brmsformula`, which specifies over which groups a mixture should be calculated. For example, when attempting to specify a mixture over participants ("ID"), one could write: `y | mix(gr = "ID") ~ ...` 

For translating this into stan-code, one could generate temporary variables for first accumulating the group-level likelihood components, which are then passed to the log_mix function. For a two-component mixture over a grouping-variable this could look something like the example code below, where L_1 and L_2 are the two likelihood components, N_1 is the number of grouping-levels, J_mix is the grouping-indicator per observation (similar to how it is done for random effects), N is the number of total observations, and lambda is the mixture proportion: 
 
```
model {
   vector L_1[N_1] = rep(0, N_1);  // Likelihood component for model 1
   vector L_2[N_1] = rep(0, N_1);  // Likelihood component for model 2

   for (i in 1:N) {
       L_1[J_mix[i]] += log-likelihood-model1
       L_2[J_mix[i]] += log-likelihood-model2
   }

   for (j in 1:N_1) {
      target += log_mix(lambda, L_1[j], L_2[j])
   }
}
```

In case the mixture is specified over single observations instead of groups of observations, this would be ne simplified to:

```
model {
   vector L_1[N] = rep(0, N);  // Likelihood component for model 1
   vector L_2[N] = rep(0, N);  // Likelihood component for model 2

   for (i in 1:N) {
       L_1[i] += log-likelihood-model1
       L_2[i] += log-likelihood-model2
   }

   for (j in 1:N) {
      target += log_mix(lambda, L_1[j], L_2[j])
   }
}
```

Thanks for considering this feature, and I hope the summary of what we discussed helps for implementation!

Best,
Philipp

[Note 21. May 2024.pdf](https://github.com/paul-buerkner/brms/files/15405209/Note.21.May.2024.pdf)

GidonFrischkorn:
Closes #1659.

The changes in this PR implement the feature requested in #1659: `mixture()` can compute the mixture over the levels of a grouping variable instead of over individual observations, so a whole group (e.g. a participant) is assigned to one component. The motivation, the likelihood distinction, and the Stan sketch are available in the issue and its linked note.

In two places I departed from the proposals from the issue:

- The grouping is specified in the call to `mixture()`, not in a `mix()` aterm in the `brmsformula`. The issue proposed `y | mix(gr = "ID") ~ ...`. That works, but a `mix` addition term collides with the existing `mi` term through R's `$` partial matching (`adforms$mi` resolves to a `mix` element when no `mi` is present), which would force `$mi` → `[["mi"]]` edits across ~8 unrelated files. Putting `gr` on the family avoids that and keeps the change scoped. The only thing the family route doesn't get for free is the grouping column reaching the model frame (built from formula variables), so it is added to `allvars` explicitly in `R/brmsterms.R`. Feel free to say if you would prefer the grouping specification as an `aterm` in the `brmsformula` instead.
- I used `log_sum_exp` over a `ps[]` array, instead of the `log_mix` proposed in the issue. This follows other brms code that already builds every mixture as `log_sum_exp` over per-component terms rather than `log_mix`, so the group-level path reuses that and generalizes to N components for free.

### Background on some coding choices

During implementation I noted some things that I want to shortly elaborate on. If you notice errors in my thinking, please say so, and I am happy to adapt the PR accordingly.

- The normalized `_lpdf` in the accumulator is required. Normalization constants cancel out of an observation-level `log_sum_exp` but are summed over the group before mixing here, so dropping them (`_lupdf`) would change the relative component weights. The grouped path therefore never uses the unnormalized form even when `normalize = FALSE`. 
- Predicting `theta` is restricted to group-constant predictors. A per-observation-varying mixing proportion is not meaningful when the component is shared across a group. This is validated in `standata`; when constant, the proportion is read from a representative observation per group (`Jmixrep`).
- `log_lik`/`loo` need to change unit for grouped mixtures. The likelihood for grouped mixtures doesn't factorize over observations, so `log_lik` returns one column per group and `loo`/`waic` become leave-one-group-out — the correct CV unit here. Base `loo`/`waic`/`loo_compare`/`r_eff` are column-count agnostic and handle it. The changes in the PR follow the same unit: `kfold` forms folds over whole groups, `reloo` and `loo_moment_match` refit/match per group, and `loo_subsample` works because the pointwise `log_lik` is already per group. `kfold_predict` works on saved grouped kfolds; predictions stay per observation while the elpd unit is the group.
- Bundled upstream **bugfix** (if you prefer I can add this in a separate PR). The Pointwise `log_lik` (`pointwise = TRUE`) was broken for *all* multivariate models without `rescor`, independent of this feature: `log_lik_pointwise` passed each response's prepared draws as `data_i` instead of `draws` (`lapply(draws$resps, log_lik_pointwise, i = i)`). Only `loo_subsample` exercises this path, which is presumably why it went unnoticed until now. 
- With its multivariate machinery, `brms` did already support independent per-response grouped mixtures. So, in the supported multivariate case, each response carries its *own* mixing proportion and its own per-group component assignment; responses can be coupled only through shared/correlated group-level effects (e.g. `(1 | p | ID)` across formulas). Thus, a group can be in component 1 for one response and component 2 for another: the joint per-group likelihood is the *product of per-response mixtures*, not a *mixture of products*. A single latent class driving all responses (one shared assignment) would require cross-response accumulation into one `Lmix` before a single `log_sum_exp`, cutting against the one-likelihood-block-per-response architecture; I felt that this is out of scope here. I added information to the `mixture()` docs, so they state the independent-assignment semantics explicitly.
- To achieve well-behaved posterior predictives, multivariate grouped mixtures are allowed if *all* responses are
grouped mixtures over the *same* grouping variable with `rescor = FALSE` (`rescor` is independently excluded for mixtures by `allow_rescor`). Then the group remains the pointwise unit jointly: its log-likelihood is the sum of the per-response per-group terms, and every CV method above carries over unchanged. Asymmetric models (grouped mixture + ordinary response) and mismatched `gr` variables are refused with errors that say why: they leave `log_lik`/`loo` without a common pointwise unit.

### Evaluation of the implementation

To test the behavior of the implementation, I reproduced Musfeld, Souza & Oberauer (2023, *PNAS* 120:e2218042120), a per-participant Hebb-repetition mixture (learner vs. non-learner):
- *Univariate accuracy model*: the generated Stan matches their per-subject `log_mix` accumulation, including a shared baseline across components (shared non-linear parameters) and a non-linear hinged-onset learner curve. Fitting all three visual conditions on the authors' OSF data recovers the learner proportion at **0.58–0.59** vs. the published **~0.60**, near-identical between groups (their central between-group null), with learning curves matching their Fig 3D.
- *Multivariate joint model*: their combined learning + awareness Stan model is two independent per-subject mixtures coupled by five 2×2 participant-effect correlations. A prototype fit on their data reproduces the key published result, the learning↔awareness onset correlation: **0.83 [0.63, 0.95]** vs. published **0.82 [0.63, 0.94]**, with awareness onset preceding learning onset as in the paper. (caveats: their ordered-beta awareness family was approximated with `Beta` plus endpoint compression; `zero_one_inflated_beta` is currently not allowed inside `mixture()`.)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,10 @@
 
 ### New Features
 
+* Compute `mixture` models over the levels of a grouping variable (e.g.
+participants) rather than over individual observations via the new `gr`
+argument of `mixture`, so that whole groups are assigned to the same mixture
+component. (#1659)
 * Specify a prior `tag` for use in prior sensitivity analysis
 via `priorsense` thanks to Kallioinen. (#1585)
 * Specify group-level prior weights via argument `pw` in multilevel
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,14 +8,11 @@ correlations, which failed with an error, for example, when using
 
 ### New Features
 
-* Compute `mixture` models over the levels of a grouping variable (e.g.
-participants) rather than over individual observations via the new `gr`
-argument of `mixture`, so that whole groups are assigned to the same mixture
-component. Cross-validation for these models is performed with the group as the
-unit (leave-one-group-out), including `kfold`, `reloo`, `loo_moment_match`, and
-`loo_subsample`. Multivariate models are supported if all responses are grouped
-mixtures over the same grouping variable with `rescor = FALSE`; the group then
-remains the pointwise unit jointly across responses. (#1659)
+* Compute `mixture` models over the levels of a grouping variable rather than
+over individual observations via the new `gr` argument of `mixture`, so that
+whole groups are assigned to the same mixture component. Cross-validation is
+then performed with the group as the pointwise unit (leave-one-group-out).
+(#1659)
 * Specify a prior `tag` for use in prior sensitivity analysis
 via `priorsense` thanks to Kallioinen. (#1585)
 * Specify group-level prior weights via argument `pw` in multilevel
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,11 @@
 # brms 2.23
 
+### Bug Fixes
+
+* Fix pointwise `log_lik` evaluation in multivariate models without residual
+correlations, which failed with an error, for example, when using
+`loo_subsample`.
+
 ### New Features
 
 * Compute `mixture` models over the levels of a grouping variable (e.g.
@@ -7,7 +13,9 @@ participants) rather than over individual observations via the new `gr`
 argument of `mixture`, so that whole groups are assigned to the same mixture
 component. Cross-validation for these models is performed with the group as the
 unit (leave-one-group-out), including `kfold`, `reloo`, `loo_moment_match`, and
-`loo_subsample`. (#1659)
+`loo_subsample`. Multivariate models are supported if all responses are grouped
+mixtures over the same grouping variable with `rescor = FALSE`; the group then
+remains the pointwise unit jointly across responses. (#1659)
 * Specify a prior `tag` for use in prior sensitivity analysis
 via `priorsense` thanks to Kallioinen. (#1585)
 * Specify group-level prior weights via argument `pw` in multilevel
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,7 +5,9 @@
 * Compute `mixture` models over the levels of a grouping variable (e.g.
 participants) rather than over individual observations via the new `gr`
 argument of `mixture`, so that whole groups are assigned to the same mixture
-component. (#1659)
+component. Cross-validation for these models is performed with the group as the
+unit (leave-one-group-out), including `kfold`, `reloo`, `loo_moment_match`, and
+`loo_subsample`. (#1659)
 * Specify a prior `tag` for use in prior sensitivity analysis
 via `priorsense` thanks to Kallioinen. (#1585)
 * Specify group-level prior weights via argument `pw` in multilevel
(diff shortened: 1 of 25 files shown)


paul-buerkner:
Thank you very much for this PR! I will review it when I find the time, hopefully during August. 

Just a quick note: please disclose the use of AI within contributions to brms. Based on the way comments and details are written, I can see AI at play here and I would like to keep a bit track of it where it enters brms. Nothing bad about it obviously (perhaps apart from the very wordy style of commenting), more like a matter of principle here.

GidonFrischkorn:
Hi Paul,

sure, sorry for not directly disclosing the AI use. I sketched out the math myself based on the issue report and explored some additional considerations for the posterior predictive and then used Claude Code for implementation and then checked the code against the math for verification. I am also happy to go over the comments to focus them on the relevant points.

If you need more details, feel free to say so.

PR #1906, opened 2026-08-04
Add CITATION.cff

florence-bockting:
## Description

This PR adds a `CITATION.cff` that facilitates the citation of software using the `Cite this repository` option in GitHub. For further information on CITATION.cff see here: https://citation-file-format.github.io/#/create-a-citationcff-file-now

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,0 +1,49 @@
+# This CITATION.cff file was generated with cffinit.
+# Visit https://bit.ly/cffinit to generate yours today!
+
+cff-version: 1.2.0
+title: >-
+  brms: An R Package for Bayesian Multilevel Models using
+  Stan
+message: >-
+  If you use this software, please cite it using the
+  metadata from this file.
+type: software
+authors:
+  - given-names: Paul-Christian
+    family-names: Bürkner
+    email: paul.buerkner@gmail.com
+    affiliation: TU Dortmund University
+    orcid: 'https://orcid.org/0000-0001-5765-8995'
+identifiers:
+  - type: doi
+    value: 10.18637/jss.v080.i01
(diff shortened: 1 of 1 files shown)


paul-buerkner:
Does this ensure the JSS paper and potentially the two other main brms papers are cited (as per readme)?

florence-bockting:
The CITATION.cff is compatible with GitHub, Zenodo, and Zotero. In this case, the citation would be imported autmatically. And in GitHub users can click on "Cite this repository" on the right upper side in your repo landing page (https://github.com/paul-buerkner/brms) and simply copy&paste the correct BibTex or corresponding rendered citation (APA). Which would look with the current CITATION.cff as follows:

> Bürkner P. (2024). brms: An R Package for Bayesian Multilevel Models using Stan (version 2.22.0). DOI: 10.18637/jss.v080.i01 URL: https://paulbuerkner.com/brms/

That means the JSS paper is currently considered as the preferred citation, unfortunately there is no way to handle multiple citations. Might this be a problem?

paul-buerkner:
Not necessarily, but I might add that the above citation doesn't even really cite the JSS paper. It just uses the corresponding doi.

florence-bockting:
I realise there is much more discussion involved then I anticipated. Sorry, for not opening an Issue first.

The CITATION.cff is primarily for citing software. But in your case the citation is linked to scientific articles which is a format that is not considered. One option would be to actually create a DOI for brms as software using e.g. Zenodo and linking there the related the articles. Then we could refer to the Zenodo DOI for the software itself, which would be separated from the DOIs related to the single articles.

However, the point needs more discussion. Therefore I would close the PR at this point and move the discussion to Slack.

Issue #1909, opened 2026-08-10
`posterior_linpred()` errors for model containing an interaction between an unordered factor and mo() in distributional formula


ciaranmci:
Firstly, thank you to @paul-buerkner and co-developers for the wonderfully useful `brms` package.

It seems that, in `brms` 2.22.0, a model containing an interaction between an unordered factor and `mo()` can fit successfully but fail during `posterior_linpred()` when that interaction occurs in distributional formulas. Below is a reprex.
I think the failure occurs in `prepare_predictions_sp()` while accessing `Csp` when I call `brms::posterior_linpred()`.
Models fit but predictions don't work. The reprex applies to parameters from both the main formula - e.g. `mu` - and from the distributional formula - e.g. `zoi` ( or e.g. `cutzero` when using `ordbetareg`, which is where I originally came across the problem).

```
packageVersion("brms") # ‘2.22.0’
packageVersion("ordbetareg") # ‘0.8’
R.version.string # "R version 4.4.0 (2024-04-24 ucrt)"

library( brms )
library( ordbetareg )

# Create dataset.
set.seed(123)
n <- 60
dat <- data.frame(
  x = factor( sample(letters[1:3], n, replace = TRUE ) ),
  z = ordered( sample( 1:8, n, replace = TRUE ) ),
  group = factor( rep( 1:12, each = 5 ) )
)
dat$y <- ordbetareg::rordbeta(
  n = n,
  mu = 0.5,
  phi = 2,
  cutpoints = c( -2, 2 )
)
priors_ordB <-
  set_prior( "student_t(3, 0, 2.5)", class = "Intercept" ) +
  set_prior( "normal(0, 5)", class = "b" ) +
  set_prior( "normal(0, 5)", class = "b", dpar = "cutzero" ) +
  set_prior( "normal(0, 5)", class = "b", dpar = "cutone" )
priors_brm <-
  set_prior( "student_t(3, 0, 2.5)", class = "Intercept" ) +
  set_prior( "normal(0, 5)", class = "b" ) +
  set_prior( "normal(0, 5)", class = "b", dpar = "zoi" ) +
  set_prior( "normal(0, 5)", class = "b", dpar = "coi" )

# Fit `brms` model without interaction.
brm_no_interaction <-
  brms::brm(
    brms::bf(
      y ~ x + mo(z) + (1 | group),
      zoi ~ x + mo(z),
      coi ~ x + mo(z)
    ),
    data = dat,
    prior = priors_brm,
[...]

PR #1913, opened 2026-08-16
Change definition of `has_covars`

ciaranmci:
Suggested change to address problem raised in issue #1909 .

### DIFF
--- a/R/formula-sp.R
+++ b/R/formula-sp.R
@@ -453,7 +453,7 @@ frame_sp <- function(x, data) {
 
   # extract information on covariates
   # only non-zero covariates are relevant to consider
-  has_covars <- attr(mm, "covars")
+  has_covars <- colSums(mm != 1) > 0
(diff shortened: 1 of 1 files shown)


paul-buerkner:
Thanks! Do all tests pass locally for you?

paul-buerkner:
Just checked myself. Tests are passing, I think this is safe to merge. Thank you!

ciaranmci:
Sorry, only just seen the notifications.
I had only suggested the patch because I had implemented in on my forked version of `brms` that I was using locally. I'm glad to hear that everything works for other tests, though there [seems to have been one failure in the testing suite](https://github.com/paul-buerkner/brms/actions/runs/31979731262).
Anyway, thanks for your time and take care.

paul-buerkner:
I think the issue in the test suite is due to some failure to install dependencies or something. I don't think it's related to the patch.

PR #1918, opened 2026-08-20
Stabilised and debugged inverse Gaussian CDF and CCDF

jaburgoyne:
I am embarrassed to admit that I am (somehow) not a regular **brms** user, and so I have not tested this pull request. As requested, however, here is a version of the inverse Gaussian CDF that has been more stable for me in other contexts. I hope it is helpful even if it isn't yet thoroughly checked.

My implementation elsewhere uses a different parametrisation in order to align with conventions for the generalised inverse Gaussian, and that might also have an effect on stability. Nonetheless, the small changes I've made to this file certainly call more stable variants of the underlying Stan math functions, use some composed functions to do more work on the log scale, and fix a sign error in the CCDF.

### DIFF
--- a/inst/chunks/fun_inv_gaussian.stan
+++ b/inst/chunks/fun_inv_gaussian.stan
@@ -33,8 +33,8 @@
    *   log(P(Y <= y))
    */
    real inv_gaussian_lcdf(real y, real mu, real shape) {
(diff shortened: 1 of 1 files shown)


paul-buerkner:
That's perfect. Thank you so much!

Issue #1916, opened 2026-08-19
Add fast hierarchical parameterization using sum-to-zero from Stancon 2026 talk


spinkney:
The relevant talk is at https://github.com/spinkney/open-talks/tree/main/sum-to-zero-stancon26. The recording of the talk is at https://youtu.be/lHXav9KIEIU?si=Xm8xMnrGwiCzg-JY.

The derivation shows that under elliptically symmetric prior distributions for the varying parameters (e.g. random effects) the conventional bayesian hierarchical model can be equivalently (in distribution) expressed with a sum-to-zero parameterization. This parameterization will typically be much faster due to easier posterior geometry for the sampler to explore. Most hierarchical model specifications will use normal distributions for the prior which satisfies the elliptically symmetric distribution assumption, as will a large class of [prior distributions such as student_t, laplace, and others](https://en.wikipedia.org/wiki/Elliptical_distribution). The downside for coding this in brms is when a user wants a distribution outside of this class, then we must handle this case by using the conventional parameterization.  

The talk shows the path for the scalar case, including varying standard deviations on the varying parameters, and is clear about the benefits. The argument does extend to the multivariate case however this comes with additional computational costs due to the covariance matrix and will need to be further tested. 

Depending on how flexible (and ambitious) we wish to be with users placing additional priors on the varying parameters, we will have to adjust due to the change of variables. For example, the typical case is a single normal prior on the varying effects which can be recovered after the model fit in generated quantities or outside of Stan. However when a user wishes to place additional priors on the effects (probably rare) then this adjustment will need to happen within the sampling stage and account for the change of variables. In the scalar case this adds little overhead but in the multivariate case it will add considerable overhead due to inversion of a covariance matrix.

I suggest adding the scalar case first an working through the design decisions before the multivariate case.

PR #1919, opened 2026-08-21
Add fast sum-to-zero parameterizations for group effects

spinkney:
## Summary

This PR adds an experimental, opt-in sum-to-zero (S2Z) parameterization for ordinary group-level effects:

```r
bf(y ~ x * z + (1 + x * z | gr(g, s2z = TRUE)))
```

It supports varying intercepts, slopes, interactions, correlated or independent effects, multiple grouping factors, and separate distributional predictors. The implementation samples physical zero-sum group contrasts, integrates the omitted common group-effect mean, and reconstructs conventional population and group coefficients in generated quantities.

The group-varying-scale feature is part of the same S2Z construction:

```r
(1 + x | gr(g, s2z = TRUE, scale = "varying"))
```

For coefficient `k` and `J` observed groups, the conventional iid standard-normal log-scale innovations are decomposed into `J - 1` orthonormal S2Z contrasts and one retained mean coordinate:

```text
z[j,k] = c[j,k] + m[k] / sqrt(J)
sum_j c[j,k] = 0

reference_sd[k] =
  sd[k] * exp(sdlog[k] * m[k] / sqrt(J))

sd_level[j,k] =
  reference_sd[k] * exp(sdlog[k] * c[j,k])
```

Therefore,

```text
sd_level[j,k] = sd[k] * exp(sdlog[k] * z[j,k])
```

exactly as in the conventional varying-scale hierarchy. The relative log scales sum to zero, and `reference_sd[k]` is the observed groups’ geometric-mean scale. The log-scale mean remains sampled, so no scale degree of freedom is constrained away or integrated out.

This transformation is orthogonal and has unit absolute determinant. Consequently, the S2Z implementation is a reparameterization of the corresponding conventional iid-lognormal varying-scale hierarchy—not a different statistical model. It is a different hierarchy only when compared with the existing shared-scale model obtained by setting `sdlog = 0`.

The motivation is computational. Conventional varying-scale hierarchies can produce strong dependence between group effects and their scales, including funnel-like geometry. Expressing both the effects and log-scale innovations in aligned mean-and-contrast coordinates removes the conventional common-effect direction and isolates scale heterogeneity into a retained mean and zero-sum relative contrasts.

For the exchangeable Gaussian information calculation, the finite-population effect and group-contrast sectors are orthogonal in expectation. With `||` terms and an intercept-orthogonal slope design, the expected information in the contrast sector is diagonal. Correlated blocks instead produce repeated coefficient-sized blocks. The orthogonal log-scale rotation itself preserves the Fisher spectrum; the improvement comes from aligning the hierarchy and analytically removing the omitted effect mean. The retained log-scale mean can still couple to the baseline `sd`, and realized unbalanced or heterogeneous designs need not have diagonal information. These are geometry results, not a general performance guarantee.

The varying-scale hierarchy adds:

- `sdlog`, controlling coefficient-specific log-scale heterogeneity;
- `sd_level`, exposing realized scales for observed groups;
- corresponding `VarCorr()` output;
- fresh scale innovations for Gaussian new-group predictions;
- optional proper continuous prior factors for selected realized scales.

Named `sd_level` factors allow domain knowledge or known differences in group-level information to regularize selected scales differently. They augment rather than replace the `sd`/`sdlog` hierarchy. Because they distinguish named observed levels, the resulting full prior is no longer exchangeable across those levels.

For shared-scale models and for the conventional varying-scale hierarchy represented above, the committed deterministic and fitted-model checks found no discrepancies beyond their stated tolerances in the tested configurations. These finite checks support the implemented cases but do not prove equivalence for every possible formula or guarantee improved sampling.

### AI and source-material disclosure

This PR was developed with assistance from OpenAI Codex (GPT-5) under Sean Pinkney’s direction. It draws on issue #1916, Sean’s mathematical derivations, prototype Stan programs, recovery identities, benchmark designs, named-level prior interface, and [[hierarchical-model tutorial](https://github.com/spinkney/hierarchical_model_tutorial)](https://github.com/spinkney/hierarchical_model_tutorial).

The underlying derivations and prototypes are not committed. The implementation and committed tests are therefore the reviewable record for the claims above.

Refs #1916.


spinkney:
@paul-buerkner I'm not super familiar with the brms internals so I used chatgpt. This covers both the independent and the dependent cases.

paul-buerkner:
Thank you very much! I have yet to check it in detail. One quick question: Does your OP render properly? To me it reads just like a single blob of text without linebreaks etc.

avehtari:
I also see badly formatted blob.

If ordinal models are not yet supported, the error message could be more informative
```
fit_s2z <- brm(r3 ~ btype + mode + situ + (btype + mode + situ | gr(id, s2z=TRUE)),
                         family = cumulative(),
                         data = VerbAgg)
Error in `re_s2z_info()` at ]8;line = 340:col = 15;file:///u/77/ave/unix/proj/brms/R/re-s2z.Rbrms/R/re-s2z.R:340:15]8;;:
! Every sum-to-zero group-level coefficient must have a matching population-level design column. Missing: 'Intercept'.
Run `]8;;x-r-run:rlang::last_trace()rlang::last_trace()]8;;` to see where the error occurred.
```

spinkney:
> Thank you very much! I have yet to check it in detail. One quick question: Does your OP render properly? To me it reads just like a single blob of text without linebreaks etc.

I'm going to make this a draft and review. Then I'll write the beginning myself as I hate how the AI wrote it. Give me a few days.

> If ordinal models are not yet supported, the error message could be more informative

Thanks for spotting this. I'll see if the AI can make a table of the models supported vs not and why. Then I'll review and post here. 

@paul-buerkner did you see the tests from @avehtari at https://discourse.mc-stan.org/t/help-testing-brms-pr-for-sum-to-zero-and-partial-centering/41542/3?u=spinkney?

paul-buerkner:
The point was more that the OP was not correctly formatted at all, but of course it doesn't need to be readable right now. 

I saw the tests and they look great indeed! Impressive speedups!

I am confused by the varying scales feature though. What is the purpose of this? It does not seem important for the sum to zero part (or is it?) and I would prefer having a separate discussion for it and only then a PR once I am convinced it is a relevant feature.

spinkney:
> I am confused by the varying scales feature though. What is the purpose of this? It does not seem important for the sum to zero part (or is it?) and I would prefer having a separate discussion for it and only then a PR once I am convinced it is a relevant feature.

In the talk I showed that the varying sd's are expressed in the log-scale with a s2z vector. A similar decomposition to the varying intercepts/slopes exists to ensure that this equate to the conventional Bayesian fit. These varying scales often induce funnel like geometry and frustrate the sampler with the conventional fit. The benefit with the s2z parameterization on the log-scale is that it decouple the random effects from their standard deviations, sort of like a non-centering procedure. It happens to be that it is an orthogonal reparameterization of the corresponding conventional iid-lognormal varying-scale hierarchy. So, it's better geometry, and the expected Fisher information is diagonal in the `||` case, in the dense covariance case it becomes block-diagonal so, in general, better behaved. 

It seems the AI wanted to think of this as a "new" model but it's actually not. It's a reparameterization that is enabled by the s2z construction. The pr adds the ability to place different priors on the standard deviations given domain expertise or possibly because the data are sparse in some groups and it makes sense to make some levels of this prior standard deviation vary more.

paul-buerkner:
Thank you for this explanation. I am still a bit confused. Do we need varying SDs for this PR? It seems like an add-on feature that is not at the core of s2z. And it kind of already is supported via argument `by` of the `gr` function, which allows SDs to vary another factor variable. Or am I misunderstanding what you mean by varying SDs?

Given the complexity of getting the core s2z right in all basic scenarios with single SDs (brms standard; i.e., one SD for the random intercepts, one SD per set of random slopes), I would like to focus on that first and then discuss later if/how we want to support varying SDs beyond what `by` already supports. The latter feature has serious implications on user interface and I would like to think about that more carefully first.

I really appreciate your work on this and for supporting it in brms! I just want to make sure nothing sneaks in that I don't yet fully understood and have thought through myself.

spinkney:
Yes, I agree and I think this pr should be split 3-ways. The s2z varying intercept/slope as one, the varying standard deviations as another, and the centering stuff as a 3rd. This makes it clear that these are not dependent on each other.

paul-buerkner:
Great! I would prefer a PR only for the first part at the moment (s2z for varying intercept/slopes, with and without correlations between random effects). 

For the other two parts I am not yet convinced that they are necessary / sufficiently useful for brms given the maintainance burden they might add.

avehtari:
I did not show results separately, but the `center=TRUE` (which is the default) makes the big part of the speedup!

paul-buerkner:
Makes sense, thank you! To be clear, I am not opposed to centering, I just would prefer implementing one centering option (e.g., the current default) without alternatives and without exposing a `center` argument to users.

avehtari:
- `s2z=TRUE` helps most when the conditional posterior for the location is narrow, this can happen with a few groups with strong likelihood each or with many groups with weak likelihood each. Furthermore it helps more when there are many hierarchical components with each going from sum-to-non-zero to sum-to-zero
- `center=TRUE` helps most when each group has strong likelihood, and `center=FALSE` helps most when each group has weak likelihood.
- With strong group likelihoods, `s2z=TRUE, center=TRUE` makes sense, but I have already two examples were `s2z=TRUE, center=FALSE` has better performance, and at least one example where `s2z=TRUE, center="auto"` has the best performance, so I think it would be nice to have `center` exposed to users.
- The current implementation doesn't allow `s2z=FALSE, center=TRUE`, so I didn't test that.

paul-buerkner:
Okay, perhaps I underestimated the importance of centering. Perhaps then leave centering in this PR. Is this a feature that is in principle well defined also without the sum-to-zero constraint?

spinkney:
Centering is really important and the "auto" centering is super cool and something I've wanted to build out more. I do think that splitting this into different prs is best. Let me do the s2z first and I'll split off the centering one and sd to other prs. For the centering pr, this approach works with or without s2z for any distribution that is closed under linear operations, normal being the most widely used but others are in https://en.wikipedia.org/wiki/Location%E2%80%93scale_family.

avehtari:
> Centering is really important and the "auto" centering is super cool and something I've wanted to build out more. 

You could after first sampling improved auto centering by optimizing the centering parameters to maximize the normality of the posterior (this is what Niko Siccha has experimented with)

avehtari:
As I wanted to test this with an ordinal model (family=cumulative()), I asked Claude to implement that and it works for `cumulative`/`sratio`/`cratio`/`acat`. Claude's explanation of derivation sounds reasonable, but I didn't examine the details, but thought to mention that it seems feasible. In the specific verbagg example no benefit from s2z (and huge slowdown if using center=TRUE)

spinkney:
> As I wanted to test this with an ordinal model (family=cumulative()), I asked Claude to implement that and it works for `cumulative`/`sratio`/`cratio`/`acat`. Claude's explanation of derivation sounds reasonable, but I didn't examine the details, but thought to mention that it seems feasible. In the specific verbagg example no benefit from s2z (and huge slowdown if using center=TRUE)

Can you send me the derivation?

avehtari:
> Can you send me the derivation?

I sent it via Slack

mhollanders:
How does the s2z of the varying scales vary from simplex decomposing a single scale parameters across the groups?

Issue #1912, opened 2026-08-16
Should `update.brmsfit()` inherit seed?


LucianoDeBenedictis:
`update.brmsfit()` inherits most arguments from the previous fit if they are not overwritten, such as formula, data, priors, sampling controls, etc... I was wrong to assume that the seed argument gets inherited, because it's not. This is easy enough to check:
```
library(brms)

dat <- data.frame(y = rnorm(100))

fit <- brm(y ~ 1,
            dat,
            seed = 1)

fit[["fit"]]@stan_args[[1]][["seed"]]

dat2 <- dat + 1

#you can also update the formula instead
fit2 <- update(fit,
               newdata = dat2)

fit2[["fit"]]@stan_args[[1]][["seed"]]
```
I'd like to ask if it is desirable to have seed be inherited as well, if it was specified in the original fit and it's not provided to `update`, or maybe rolling the seed forward by one instead of using a random one. After all, the purpose of setting a seed is to ensure a reproducible fit, and this shouldn't be different when using `update`.

paul-buerkner:
It's a good point. Right now, the seed in brm is just passed to the backend (rstan or cmdstanr) and not stored directly. Only in the backend, yes but for reliability retaining it in update, we would have to start storing it directly in the brmsfit. I will consider this as a feature for the future. If you have the time yourself, feel free to open a PR.

LucianoDeBenedictis:
The seed is in some way stored in the fit object: it's in the `stan_args` slot of the `fit` sub-element, which is by-chain. This is how I was able to inspect it in the example, but I agree this might not be the best place to store and retrieve it. When an agreement is reached on where the seed should live to be gathered by `update` (whether the current place or another), I will see how I can be of help.

paul-buerkner:
I think if we want to safely keep the seed upon updating that it would have to be stored as `brmsfit$seed`. However, after thinking more about it, I am not sure we want that. I would prefer reproducibility to be ensured explicitely every call, which would favor seed to be passed to update too instead of reusing the seed. @florence-bockting do you have thoughts on this question?

lunafazio:
Not Florence, but I do have some thoughts on this 😅

[...]

PR #1920, opened 2026-08-21
update examples to show seed usage

LucianoDeBenedictis:
Since `seed` is not inherited by `update()`, the documentation now shows this in the examples. Following #1912


paul-buerkner:
Thank you!

PR #1921, opened 2026-08-21
Fix @aliases placement in posterior.R roxygen documentation

LucianoDeBenedictis:
While working on #1920 and rebuilding the documentation, I noticed an error in posterior.R. This was caused by the `@aliases` line being before what should probably be the description. The error said `@aliases` was three lines instead of one, and the description "Index variables, iterations..." was not rendered. I just moved `@aliases` after the description line and rebuilt the .Rd


paul-buerkner:
Thank you!

PR #1925, opened 2026-08-25
Add isolated physical sum-to-zero group effects

spinkney:
Adds `gr(..., s2z = TRUE)` for physical sum-to-zero sampling coordinates while preserving the conventional population and group-effect model. The likelihood uses finite-population coefficients and zero-sum deviations; generated quantities reconstruct conventional coefficients for existing extraction and prediction APIs.

S2Z validation and predictor metadata live in `R/re-s2z.R`, and Stan generation lives in `R/stan-re-s2z.R`, with small integration hooks in ordinary generation. Each predictor has one plan for coefficient maps, active priors, and group blocks. The specialized numerical paths share conventional-coefficient reconstruction. Unrelated spatial, GP, covariance, and QR changes have been removed.

Supports intercepts/slopes, shared coefficient-specific scales, correlated or independent Gaussian and Student group effects, multiple blocks within a predictor, supported population priors, and mixed ordinary/S2Z models. Known level covariance (`cov`), `by`, `pw`, cross-predictor IDs, ordinal location predictors, centering controls, and varying group scales remain outside this foundation. See `gr()` help and `tests/local/s2z-design.md` for the support boundaries and implementation contract.

This PR includes two supporting fixes with ordinary-model regression tests: exact prior-list lookup for mixed constant/estimated parameters, and retention of fitted group-distribution metadata under reduced prediction formulas. These fixes are included in this PR's review and do not require separate PRs.

Validation:

- Six focused S2Z suites passed, along with formula, prediction preparation, helper, stancode (including local-only tests), and standata tests. The optional Cox fixture was skipped because `splines2` is unavailable.
- Separate-library comparisons found exact Stan source/data equality for 21 conventional fixtures against the upstream baseline plus the two fixes.
- Twenty-two S2Z fixtures matched the previous foundation after removal of the verified-unused `chol2inv_brms` definition; two unsupported-design diagnostics also matched.
- All 22 S2Z models passed CmdStan 2.39 stanc syntax checks. This update makes no new sampling or performance claim.

Developed with OpenAI Codex assistance under Sean Pinkney's direction, using the existing foundation derivations, implementation, and committed tests.

Refs #1916.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,31 @@
 
 ### New Features
 
+* Add an experimental physical sum-to-zero parameterization via
+`gr(..., s2z = TRUE)` for ordinary matched varying intercepts, numeric and
+factor slopes, and interactions. The likelihood uses group deviations that
+sum exactly to zero over observed levels, while conventional population
+coefficients and group effects are reconstructed for the public output.
+Gaussian and Student-t group effects, correlated and independent blocks,
+coefficient-specific scales shared across levels, and multiple local S2Z
+blocks are supported. Separate S2Z IDs may be used in otherwise eligible
+location, distributional, nonlinear, categorical, multinomial, and
+multivariate predictors.
+
+Population coordinates touched by an S2Z omitted-mean map support exact flat,
+normal, Student-t, Cauchy, and logistic priors with numeric constant arguments.
+Logistic priors use explicitly sampled standardized omitted means with their
+exact target density and Jacobian; this includes the default `logistic(0, 1)`
+intercept prior of auxiliary probability predictors. Fixed-only population
+coordinates retain ordinary brms prior handling. The omitted means of multiple
+local blocks are handled jointly, with specialized scalar, diagonal, and
+Gaussian Matheron paths where applicable.
+
+This foundation deliberately excludes ordinal location predictors, centered,
+partially centered, or automatically centered modes, group scales varying by
+level, and structural extensions such as `by`, `cov`, `pw`, multi-membership,
+special group coefficients, cross-predictor IDs, and sparse or QR designs.
+(#1916)
 * Specify a prior `tag` for use in prior sensitivity analysis
 via `priorsense` thanks to Kallioinen. (#1585)
 * Specify group-level prior weights via argument `pw` in multilevel
@@ -2295,5 +2320,3 @@ have proper priors by default.
 # brms 0.1.0
 
 * Initial release version
-
-
--- a/R/brmsframe.R
+++ b/R/brmsframe.R
@@ -42,6 +42,20 @@ brmsframe.brmsterms <- function(x, data, frame = NULL, basis = NULL, ...) {
   data <- subset_data(data, x)
   x$frame$resp <- frame_resp(x, data = data)
   x$frame$ac <- frame_ac(x, data = data)
+  # Keep top-level response and family labels available to local predictor
+  # validation. Category and mixture predictors otherwise carry only their
+  # dpar prefix, even though the enclosing model context is known here.
+  s2z_response <- x$resp %||% ""
+  if (!length(s2z_response) || !nzchar(s2z_response)) {
+    response_vars <- all.vars(x$respform)
+    if (length(response_vars)) {
+      s2z_response <- response_vars[1L]
+    }
+  }
+  x$frame$s2z_context <- list(
+    response = s2z_response,
+    family = x$family$family %||% ""
+  )
   for (dp in names(x$dpars)) {
     x$dpars[[dp]] <- brmsframe(
       x$dpars[[dp]], data, frame = x$frame,
@@ -85,6 +99,11 @@ brmsframe.btl <- function(x, data, frame = list(), basis = NULL, ...) {
   # only store the ranefs of this specific linear formula
   x$frame$re <- subset2(frame$re, ls = check_prefix(x))
   class(x) <- c("bframel", class(x))
+  validate_re_s2z_structure(x, data = data)
+  validate_re_s2z_design(x, data = data)
+  if (has_re_s2z(x)) {
+    x$frame$re_s2z <- re_s2z_infos(x)
+  }
   # these data_ functions may require the outputs of the corresponding
   # frame_ functions (but not vice versa) and are thus evaluated last
   x$sdata$gp <- data_gp(x, data, internal = TRUE)
--- a/R/exclude_pars.R
+++ b/R/exclude_pars.R
@@ -80,6 +80,24 @@ exclude_pars.bframel <- function(x, save_pars, ...) {
       "scales", "merged_Intercept", "zcar", "nszcar", "zerr"
     )
     c(out) <- paste0(par_classes, p)
+    if (has_re_s2z(x)) {
+      c(out) <- paste0(
+        c(
+          "theta_s2z", "theta_s2z_active", "fixed_s2z",
+          "zfixed_s2z", "sdfixed_s2z"
+        ),
+        p
+      )
+      infos_s2z <- re_s2z_infos(x)
+      if (length(infos_s2z)) {
+        n_inactive_s2z <- length(infos_s2z[[1L]]$inactive_q)
+        c(out) <- paste0(
+          "par_fixed_s2z", p, "_", seq_len(n_inactive_s2z)
+        )
+      }
+      q <- length(x$frame$fe$vars)
+      c(out) <- paste0("udf_b_s2z", p, "_", seq_len(q))
+    }
     smframe <- x$frame$sm
     for (i in seq_rows(smframe)) {
       nb <- seq_len(smframe$nbases[i])
@@ -100,6 +118,26 @@ exclude_pars_re <- function(bframe, save_pars, ...) {
   rm_re_pars <- c(if (!save_pars$all) c("z", "L"), "Cor", "r")
   for (id in unique(reframe$id)) {
     c(out) <- paste0(rm_re_pars, "_", id)
+    r <- subset2(reframe, id = id)
+    if (!save_pars$all && isTRUE(r$s2z[1])) {
+      s2z_classes <- c(
+        "z_s2z", "z_mean_s2z", "r_s2z", "H_s2z",
+        "q_explicit_s2z", "prior_mean_s2z",
+        "prior_prec_s2z", "prior_scale_s2z",
+        "W_matheron_s2z", "sqrt_W_matheron_s2z",
+        "L_W_matheron_s2z", "theta_white_matheron_s2z",
+        "group_scale_s2z", "group_prec_s2z",
+        "L_Sigma_s2z", "Q_Sigma_s2z", "P_group_s2z", "h_group_s2z",
+        "P_s2z", "L_P_s2z", "H_joint_s2z", "h_joint_s2z",
+        "D_s2z", "sqrt_D_s2z", "D_diag_s2z", "intercept_map_s2z",
+        "rank1_info_s2z", "group_quad_s2z", "joint_quad_s2z",
+        "mhat_s2z", "qhat_s2z", "white_s2z", "mean_r_s2z",
+        "q_recovered_s2z"
+      )
+      c(out) <- paste0(s2z_classes, "_", id)
+      rp <- usc(combine_prefix(r))
+      c(out) <- paste0("r_s2z_", r$id, rp, "_", r$cn)
+    }
   }
   if (isFALSE(save_pars$group)) {
     p <- usc(combine_prefix(reframe))
--- a/R/exclude_pars.R
+++ b/R/exclude_pars.R
@@ -126,7 +126,7 @@ exclude_pars_re <- function(bframe, save_pars, ...) {
         "prior_prec_s2z", "prior_scale_s2z",
         "W_matheron_s2z", "sqrt_W_matheron_s2z",
         "L_W_matheron_s2z", "theta_white_matheron_s2z",
-        "group_scale_s2z", "group_prec_s2z",
+        "group_scale_s2z", "group_prec_s2z", "group_info_s2z",
         "L_Sigma_s2z", "Q_Sigma_s2z", "P_group_s2z", "h_group_s2z",
         "P_s2z", "L_P_s2z", "H_joint_s2z", "h_joint_s2z",
         "D_s2z", "sqrt_D_s2z", "D_diag_s2z", "intercept_map_s2z",
--- a/R/formula-re.R
+++ b/R/formula-re.R
@@ -13,24 +13,144 @@
 #' @param by An optional factor variable, specifying sub-populations of the
 #'   groups. For each level of the \code{by} variable, a separate
 #'   variance-covariance matrix will be fitted. Levels of the grouping factor
-#'   must be nested in levels of the \code{by} variable.
+#'   must be nested in levels of the \code{by} variable. This argument is not
+#'   currently supported with \code{s2z = TRUE}.
 #' @param cor Logical. If \code{TRUE} (the default), group-level terms will be
 #'   modelled as correlated.
+#' @param s2z Logical. If \code{TRUE}, use the experimental physical
+#'   sum-to-zero parameterization described in the section
+#'   \dQuote{Physical sum-to-zero effects}. The default is \code{FALSE}.
 #' @param id Optional character string. All group-level terms across the model
 #'   with the same \code{id} will be modeled as correlated (if \code{cor} is
-#'   \code{TRUE}). See \code{\link{brmsformula}} for more details.
+#'   \code{TRUE}). With \code{s2z = TRUE}, this sharing is limited to terms in
+#'   one linear predictor: an ID cannot span responses, categories, or
+#'   distributional or nonlinear predictors. See \code{\link{brmsformula}}
+#'   for more details.
 #' @param pw Optional numeric variable specifying prior weights. They weight the
 #'   contribution of each group to the log-prior of the group-level
 #'   coefficients. Should have one distinct value for each level of the
-#'   grouping variable.
+#'   grouping variable. This argument is not currently supported with
+#'   \code{s2z = TRUE}.
 #' @param cov An optional matrix which is proportional to the within-group
 #'   covariance matrix of the group-level effects. All levels of the grouping
 #'   factor should appear as rownames of the corresponding matrix. This argument
 #'   can be used, among others, to model pedigrees and phylogenetic effects. See
 #'   \code{vignette("brms_phylogenetics")} for more details. By default, levels
 #'   of the same grouping factor are modeled as independent of each other.
-#' @param dist Name of the distribution of the group-level effects.
-#' Currently \code{"gaussian"} is the only option.
+#'   This argument is not currently supported with \code{s2z = TRUE}.
+#' @param dist Name of the distribution of the group-level effects. Supported
+#'   options are \code{"gaussian"} and \code{"student"}.
+#'
+#' @section Physical sum-to-zero effects:
+#' With \code{s2z = TRUE}, \pkg{brms} samples each varying coefficient as
+#' deviations whose sum is exactly zero over the grouping levels observed in
+#' the model data. For
+#' \code{1 + x + (1 + x | gr(g, s2z = TRUE))}, the constraint is applied
+#' separately to the varying intercepts and the varying \code{x} slopes.
+#' \pkg{brms} handles each omitted group-effect mean jointly with its matching
+#' population coefficient, then reconstructs conventional population and
+#' group effects. Thus \code{fixef}, \code{ranef}, \code{coef}, and prediction
+#' retain their usual interpretation.
+#'
+#' Every S2Z varying design column must have a population-level column with the
+#' same coefficient name and numerically identical values. This applies to
+#' varying intercepts, numeric and factor slopes, and interaction columns,
+#' including their contrasts. For example, \code{x * f} must appear on both
+#' sides of the grouping bar when all of its expanded columns vary. At least
+#' two grouping levels must be observed.
+#'
+#' The main support boundaries are:
+#'
+#' \tabular{lll}{
+#' Feature \tab Supported? \tab Details\cr
+#' Ordinary designs \tab Yes \tab Matched intercepts, numeric or factor slopes,
+#' and interactions.\cr
+#' Group distribution \tab Yes \tab Gaussian or Student-t.\cr
+#' Coefficient correlation \tab Yes \tab \code{cor = TRUE} or \code{FALSE}
+#' within one predictor-local ID.\cr
+#' Multiple blocks \tab Yes \tab One or more local S2Z IDs in a linear
+#' predictor.\cr
+#' Multiple predictors \tab Yes \tab Separate S2Z IDs in otherwise eligible
+#' response, distributional, nonlinear, or category predictors.\cr
+#' Shared cross-predictor ID \tab No \tab Includes \code{| q |} shared by
+#' responses in \code{mvbind(...)}.\cr
+#' Known level covariance \tab No \tab Includes \code{cov = A} for
+#' phylogenetic or pedigree models.\cr
+#' Structural extensions \tab No \tab \code{by}, \code{pw},
+#' multi-membership, and special group coefficients.\cr
+#' Group scales \tab Yes \tab Coefficient-specific \code{sd} parameters shared
+#' across observed grouping levels.\cr
+#' Active population priors \tab Limited \tab Flat, \code{normal},
+#' \code{student_t}, \code{cauchy}, or \code{logistic}, with numeric constant
+#' arguments.\cr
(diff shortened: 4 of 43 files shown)


paul-buerkner:
Thank you so much!! Will you let me know once you want me to review this PR?

paul-buerkner:
@spinkney I took a quick look at the PR and there seems to be a lot of changes that, at least superficially, are unrelated to the s2z parameterization. Do they need to be there? I worry that we would accidentally break something entirely unrelated by changing things where we don't need to.

spinkney:
Let me know what you think of the updates. Is this still too much? Also, I had to add the Stan function sum-to-zero instead of using the built-in to be compatible with different versions of cmdstan/rstan.

paul-buerkner:
Thank you! This is now reviewable, I think :-)

R/brmsframe.R
```diff
@@ -37,11 +37,26 @@ brmsframe.brmsterms <- function(x, data, frame = NULL, basis = NULL, ...) {
     x$frame$re <- subset2(x$frame$re, resp = x$resp)
   }
   if (!is.null(basis)) {
-    x$frame$basis <- basis[c("resp_levels", "bhaz")]
+    x$frame$basis <- basis[c("resp_levels")]
   }
   data <- subset_data(data, x)
   x$frame$resp <- frame_resp(x, data = data)
   x$frame$ac <- frame_ac(x, data = data)
+  if (any(x$frame$re$s2z)) {
```
paul-buerkner, line 46: Can you elaborate why this is needed? Perhaps we can handle it in a prettier way that adding an element at the top level of $frame?

R/brmsframe.R
```diff
@@ -85,6 +100,11 @@ brmsframe.btl <- function(x, data, frame = list(), basis = NULL, ...) {
   # only store the ranefs of this specific linear formula
   x$frame$re <- subset2(frame$re, ls = check_prefix(x))
   class(x) <- c("bframel", class(x))
+  if (has_re_s2z(x)) {
```
paul-buerkner, line 107: Is this really the right place to validate and store stuff? Why is this not part of $frame$re? Seems to be relevant only for REs after all

R/exclude_pars.R
```diff
@@ -80,6 +80,9 @@ exclude_pars.bframel <- function(x, save_pars, ...) {
       "scales", "merged_Intercept", "zcar", "nszcar", "zerr"
     )
     c(out) <- paste0(par_classes, p)
+    if (has_re_s2z(x)) {
```
paul-buerkner: why is this needed? We have a dedicate exclude_pars_re function that should handle parameter exclusion. Also please have the related helper functions (if needed) in the respective file. Here in `exclude_pars` and name them accordingly. Based on the functions names here, I cannot see that they are even related to exclude_pars

R/formula-re.R
```diff
@@ -13,24 +13,89 @@
...
 #'   can be used, among others, to model pedigrees and phylogenetic effects. See
 #'   \code{vignette("brms_phylogenetics")} for more details. By default, levels
 #'   of the same grouping factor are modeled as independent of each other.
-#' @param dist Name of the distribution of the group-level effects.
-#' Currently \code{"gaussian"} is the only option.
+#'   This argument is not currently supported with \code{s2z = TRUE}.
+#' @param dist Name of the distribution of the group-level effects. Supported
+#'   options are \code{"gaussian"} and \code{"student"}.
+#'
+#' @section Physical sum-to-zero effects:
```
paul-buerkner, line 38: what of this section is actually relevant for the user to know? I think barely anything. If something is not supported, brms should throw an informative error. This doesn't need to be all written here.

R/formula-re.R
```diff
@@ -53,18 +118,28 @@
 #' fit4 <- brm(count ~ Trt + (1|gr(patient, pw = patient_samp_wgt)),
 #'             data = epilepsy)
 #' summary(fit4)
+#'
+#' # physical sum-to-zero varying intercepts, slopes, and interactions
+#' s2z_prior <- prior(exponential(2), class = "sd", group = "patient",
```
paul-buerkner: why specify a dedicated prior object here? Are default priors not working in this case?

R/formula-re.R
```diff
@@ -605,6 +681,7 @@ frame_re <- function(bterms, data, old_levels = NULL) {
       nlpar = re$nlpar[[i]],
       ggn = NA,
       cor = re$cor[[i]],
+      s2z = re$gcall[[i]]$s2z %||% FALSE,
```
paul-buerkner: is `%||% FALSE` needed here? Seems to not be checked for any of the other arguments

R/prepare_predictions.R
```diff
@@ -594,8 +594,9 @@ prepare_predictions_re_global <- function(bframe, draws, sdata, old_reframe, res
     }
     # only prepare predictions of effects specified in the new formula
     cols_match <- c("coef", "resp", "dpar", "nlpar")
-    used_rpars <- which(find_rows(old_reframe_g, ls = reframe_g[cols_match]))
-    used_rpars <- outer(seq_len(nlevels), (used_rpars - 1) * nlevels, "+")
+    used_ranef <- which(find_rows(old_reframe_g, ls = reframe_g[cols_match]))
```
paul-buerkner: are changes to prepare_predictions really needed? Post-processing of random effects is a very delicate thing and I would prefer not touching it unless have really have to. Since s2z does not affect post-processing at all, I believe, I don't think this change here is vital for this PR. Please correct me if I am wrong

R/priors.R
```diff
@@ -1358,6 +1358,10 @@ validate_prior <- function(prior, formula, data, family = gaussian(),
   prior <- prior + prior_no_checks
   rownames(prior) <- NULL
   attr(prior, "sample_prior") <- sample_prior
+  # S2Z prior and cross-predictor constraints depend on the complete effective
+  # prior table, so validate them only after ordinary and special priors have
+  # been resolved. This also makes public validate_prior() authoritative.
+  validate_re_s2z_prior_global(bframe, prior = prior)
```
paul-buerkner: I see the point in having it at top level here but the relevant validation functions should live in prior.R not in a separate file.

R/re-s2z.R
```diff
@@ -0,0 +1,898 @@
+# Helpers for the physical sum-to-zero parameterization of group effects
+
```
paul-buerkner: these functions should be split and moved to the respective R files (following, implicit unfortunately, brms function placement convention) not adding a new file for this specific feature

R/re-s2z.R
```diff
@@ -0,0 +1,898 @@
+# Helpers for the physical sum-to-zero parameterization of group effects
```
paul-buerkner: this file is extremly hard to review. First, figure out if all of these helpers are really needed or (partly) just reinventing what existing brms functionality already offers. Then, move functions to the R files where they are used (priors to priors, etc.). This way, it should become much easier for me to review and understand which of this new machinery is really required and whether its correct.

R/stan-likelihood.R
```diff
@@ -1367,6 +1367,9 @@ args_glm_primitive <- function(bterms, threads = NULL, ...) {
   }
   x <- glue("X{sfx_X}{resp}{slice}")
   beta <- glue("b{sfx_b}{resp}")
+  if (has_re_s2z(bterms)) {
```
paul-buerkner, line 1370: can you clarify why this is needed and what it does, exactly?

R/stan-predictor.R
```diff
@@ -282,6 +282,11 @@ stan_fe <- function(bframe, prior, stanvars, threads, primitive,
   p <- usc(combine_prefix(px))
   resp <- usc(px$resp)
   lpdf <- stan_lpdf_name(normalize)
+  s2z <- has_re_s2z(bframe)
```
paul-buerkner, line 285: can this check be performed instide stan_fe_s2z perhaps? Also why is this added at the start of the function not in the end? Is this really something we need at the top?

R/stan-predictor.R
```diff
@@ -304,47 +309,49 @@ stan_fe <- function(bframe, prior, stanvars, threads, primitive,
...
-      }
-    } else {
-      stopifnot(decomp == "QR")
-      stopif_prior_bound(prior, class = "b", ls = px)
-      if (has_special_prior) {
-        str_add_list(out) <- stan_prior_non_centered(
-          suffix = p, suffix_class = "Q", suffix_K = ct,
-          normalize = normalize
-        )
+    if (!s2z) {
```
paul-buerkner, line 314: the s2z stuff doesn't happen in this function so the if(s2z) part is kind of empty. Why not move the s2z part inside the if-clause. It can be in its separate function but should be part of this if clauses I believe.

R/stan-predictor.R
```diff
@@ -469,6 +482,9 @@ stan_re <- function(bframe, prior, normalize, ...) {
   stopifnot(is.reframe(reframe))
   IDs <- unique(reframe$id)
   out <- list()
+  # Every S2Z predictor owns one system, including singleton blocks.
```
paul-buerkner: I don't understand this comment, not what these functions are supposed to do.

R/stan-predictor.R
```diff
@@ -498,17 +514,26 @@ stan_re <- function(bframe, prior, normalize, ...) {
...
-    normalize = normalize, ...
-  )
-  out <- collapse_lists(ls = c(list(out), tmp))
+  tmp <- lapply(IDs, function(id) {
+    .stan_re(
+      id, bframe = bframe, prior = prior, normalize = normalize,
+      s2z_system = s2z_by_id[[as.character(id)]], ...
+    )
+  })
+  systems <- lapply(s2z_systems, function(system) {
```
paul-buerkner, line 521: I don't understand this "system" part

R/stan-predictor.R
```diff
@@ -624,6 +649,13 @@ stan_re <- function(bframe, prior, normalize, ...) {
     }
   }
 
+  if (isTRUE(r$s2z[1])) {
```
paul-buerkner: seems like the whole s2z stuff is handled outside of the regular logic of RE definition? Can really none of the code be reused?

R/stan-predictor.R
```diff
@@ -2092,18 +2124,19 @@ stan_eta_re <- function(bframe, threads) {
     rpx <- check_prefix(r)
     idp <- paste0(r$id, usc(combine_prefix(rpx)))
     idresp <- paste0(r$id, usc(rpx$resp))
+    rprefix <- str_if(isTRUE(r$s2z[1]), "r_s2z_", "r_")
```
paul-buerkner: I prefer the prefix be rs2z without _ since _ may have hold meaning in brms' naming logic.

R/stan-prior.R
```diff
@@ -132,30 +132,32 @@ stan_prior <- function(prior, class, coef = NULL, group = NULL,
             coef_prior <- stan_constant_prior(
               coef_prior, par_ij, broadcast = broadcast
             )
-            str_add(out$tpar_prior_const) <- paste0(coef_prior, ";\n")
+            str_add(out[["tpar_prior_const"]]) <- paste0(coef_prior, ";\n")
```
paul-buerkner: unncessary change

R/stan-prior.R
```diff
@@ -132,30 +132,32 @@ stan_prior <- function(prior, class, coef = NULL, group = NULL,
...
+              )
             }
           }
         }
       }
     }
     # the base prior may be improper flat in which no Stan code is added
     # but we still have estimated coefficients if the base prior is used
-    has_estimated_priors <- isTRUE(nzchar(out$tpar_prior)) ||
+    has_estimated_priors <- isTRUE(nzchar(out[["tpar_prior"]])) ||
```
paul-buerkner: all of these changes from $ to [[ syntax are unncessary here and should be reverted.

R/stan-re-s2z.R
```diff
@@ -0,0 +1,2199 @@
+# Stan generation for physical sum-to-zero group effects
+
+# Internal-output metadata is kept beside the generators that own these
+# coordinates; ordinary save_pars handling only asks for the relevant names.
+re_s2z_internal_fe <- function(plan) {
```
paul-buerkner: All of these functions should be integrated into their respective R files not put in a separate file just responsible for s2z.

Also, is there really no existing logic that can be reused? As is, I am currently unable to review this file since so much is happening and I cannot tell which is really needed.

R/stancode.R
```diff
@@ -118,6 +118,7 @@ stancode.default <- function(object, data, family = gaussian(),
   parse <- as_one_logical(parse)
   backend <- match.arg(backend, backend_choices())
   silent <- as_one_logical(silent)
+  validate_re_s2z_prior_global(bterms, prior = prior)
```
paul-buerkner: validating priors should not be here

R/standata.R
```diff
@@ -111,6 +111,7 @@ standata.default <- function(object, data, family = gaussian(), prior = NULL,
                       only_response = FALSE, internal = FALSE, ...) {
 
   stopifnot(is.anybrmsframe(bframe))
+  validate_re_s2z_prior_global(bframe, prior = prior)
```
paul-buerkner: also here. priors should not be validated here.

tests/local/benchmark-s2z.R
paul-buerkner, line 1: local files should of course be removed before merging (but this has time).

tests/testthat/tests.prepare_predictions.R
```diff
@@ -0,0 +1,38 @@
+context("Tests for prediction preparation")
```
paul-buerkner: Tests needed?

tests/testthat/tests.s2z-capabilities.R
```diff
@@ -0,0 +1,592 @@
+context("Capability matrix for physical sum-to-zero group-level effects")
```
paul-buerkner, line 1: I would prefer testing more high level functions not all the low level input. Are these tests needed in that light?

tests/testthat/tests.s2z-code.R
paul-buerkner, line 1: I would prefer testing more high level functions not all the low level input. Are these tests needed in that light?

tests/testthat/tests.s2z-optimizations.R
paul-buerkner, line 1: I would prefer testing more high level functions not all the low level input. Are these tests needed in that light?

tests/testthat/tests.s2z-plan.R
paul-buerkner, line 1: I would prefer testing more high level functions not all the low level input. Are these tests needed in that light?

tests/testthat/tests.s2z-validation.R
paul-buerkner, line 1: I would prefer testing more high level functions not all the low level input. Are these tests needed in that light?

tests/testthat/tests.s2z.R
paul-buerkner, line 1: I would prefer testing more high level functions not all the low level input. Are these tests needed in that light?

NEWS.md
```diff
@@ -2,6 +2,16 @@
 
 ### New Features
 
+* Add an experimental physical sum-to-zero parameterization via
+`gr(..., s2z = TRUE)` for matched varying intercepts, slopes, factors, and
+interactions. Gaussian and Student-t group effects, coefficient correlations,
```
paul-buerkner: can the news point be shortened a bit?

paul-buerkner (changes requested):
Currently, I can only partially review this PR since so much new infrastructure is build only for s2z. In other words, the integration into the existing codebase is too small. Is it really not possible to reuse more of the existing infrastructure?

spinkney:
@paul-buerkner this is excellent and I do want you to know that it was mostly ai generated. Sorry it was being sloppy and I'm also sorry I can't devote my time to really investigate. I will have the agent clean this up but if that is insufficient then I am happy for you or one of your students to finalize the PR and ask any math questions about the implementation to me.

paul-buerkner:
Nothing to apologize for! I appreciate the time you invest! Let's see what the agent comes up with and I will do my best to clean up the rest afterwards.

wds15:
Maybe it helps, but this is the prompt I told Opus to work with when fixing the #1935 issue:

Please plan the fix for the bug demonstrated in reprex_brms_mrf_matrix_newdata_reprex.R. Do so with starting with a new unit test, which fails and is your first commit. The following commits should then make the test pass. Please follow closley the conventions of the brms package and do use all utility functions available in the package (or in R / any of the already required R packages) as needed. Try to find a fix which fits the overall architecture of the R package and needs minimal code changes. Do add comments to the code whenever that explains in addition to the code what happens.

So make him follow what is there is key to tell the thing over and over again.

R/brmsframe.R
spinkney: This metadata preserves response and family names for informative errors, and I moved it to an attribute of $frame$re.

R/brmsframe.R
spinkney: The coefficient map now lives on $frame$re, while structural validation runs during frame construction once both population- and group-level designs are available.

R/exclude_pars.R
spinkney: The agent moved S2Z exclusion into exclude_pars_re() and placed its helpers in R/exclude_pars.R with exclude_pars_ names.

R/formula-re.R
spinkney: The agent shortened the help to the feature’s behavior and practical usage requirements, leaving unsupported combinations to informative errors.

R/formula-re.R
spinkney: Default priors work here, so the agent removed the dedicated prior object from the example.

R/prepare_predictions.R
spinkney: You are right that this change is unnecessary for S2Z, so the agent restored prepare_predictions.R to the upstream implementation.

R/priors.R
spinkney: The agent moved the S2Z prior-validation functions into R/priors.R and kept their invocation within .validate_prior().

R/re-s2z.R
spinkney: The agent removed R/re-s2z.R and moved its remaining helpers into the existing formula, prior, and exclusion files according to their responsibilities.

R/re-s2z.R
spinkney: The agent removed redundant wrappers and metadata, reused existing prior-selection utilities, and replaced fixed-only prior argument remapping with ordinary stan_prior() handling.

R/stan-likelihood.R
spinkney: The GLM likelihood must use theta_s2z, which includes the omitted group means, and the agent added a comment explaining why it must match the regular predictor path.

R/stan-predictor.R
spinkney: stan_fe_s2z() now checks whether S2Z is active itself and runs at the end of stan_fe() after the shared setup.

R/stan-predictor.R
spinkney: Ordinary coefficient declarations remain under if (!s2z), while stan_fe_s2z() appends the S2Z declarations at the end after the shared design setup.

R/stan-predictor.R
spinkney: The agent rewrote the comment to explain that grouping factors sharing population coefficients require their omitted means and population priors to be handled jointly once per predictor.

R/stan-predictor.R
spinkney: A system collects the S2Z blocks sharing one predictor so their joint prior contribution and reconstruction are generated once after the individual blocks.

R/stan-predictor.R
spinkney: S2Z now shares the existing group-data declarations, scale and correlation priors, and public-effect declarations, with separate code retained for its constrained effects and reconstruction.

R/stan-predictor.R
spinkney: The agent renamed the group-effect prefix to rs2z throughout code generation, predictor use, reconstruction, exclusions, and tests.

R/stan-prior.R
spinkney: reverted this expression to the original $ syntax

R/stan-re-s2z.R
spinkney: The agent removed R/stan-re-s2z.R, moved its generators into the existing predictor and prior files, and reused ordinary RE generation for shared declarations and priors.

R/stan-prior.R
spinkney: the agent reverted those syntax changes and addressed the underlying partial-matching bug by explicitly initializing both prior-output slots to NULL.

R/stancode.R
spinkney: The agent removed the extra S2Z prior-validation call from stancode() so .validate_prior() handles it centrally.

R/standata.R
spinkney: The agent removed the extra S2Z prior-validation call from standata() for the same reason.

tests/local/benchmark-s2z.R
spinkney: The agent removed all five added local development and benchmark files from the PR.

tests/testthat/tests.prepare_predictions.R
spinkney: the agent removed this test file along with the unrelated prediction change.

tests/testthat/tests.s2z-capabilities.R
spinkney: the agent removed the source-level API audit and retained model-level generation and error tests, including ordinary-versus-S2Z prior comparisons.

tests/testthat/tests.s2z-code.R
spinkney: These tests exercise stancode(), standata(), and empty brm() fits, with two saving-behavior tests still inspecting exclude_pars() on those fits.

tests/testthat/tests.s2z-optimizations.R
spinkney: The agent retained these tests through stancode() because their generated-code checks catch regressions in indexing, matrix dimensions, and placement of fixed calculations.

tests/testthat/tests.s2z-plan.R
spinkney: The agent removed the private-plan test suite and retained coverage through model-level code, data, and validation tests.

tests/testthat/tests.s2z-validation.R
spinkney: The agent removed the private parser tests and retained checks of prior and structural errors through stancode() and standata().

tests/testthat/tests.s2z.R
spinkney: The agent removed this suite because its independent R implementations did not directly test the package’s generated Stan code.

spinkney:
> Nothing to apologize for! I appreciate the time you invest! Let's see what the agent comes up with and I will do my best to clean up the rest afterwards.

Thank you I've had the agent update and another one do a review. I hope this is better.

> Maybe it helps, but this is the prompt I told Opus to work with when fixing the #1935 issue:

Yes it does, I reworded the intro but used the main body to focus the agent on existing brms code. Thank you!

wds15:
Glad it helped. I really think that all this framework wisdom of how brms is build should ideally land in an agents instructions file to ease all of this. With that review should be much easier.

R/brmsframe.R
```diff
@@ -22,6 +22,7 @@ brmsframe.mvbrmsterms <- function(x, data, basis = NULL, ...) {
     )
   }
   class(x) <- c("mvbrmsframe", class(x))
+  validate_re_s2z_ids(x)
```
paul-buerkner, line 25: Can we move this validation to another place? Does it really need to be this "global"?

paul-buerkner: brms has a whole system for validating random effects across the whole model (`frame_re`). This is where such validation should be done.

R/brmsframe.R
paul-buerkner: Can we please move adding this metadata (if really required) inside `frame_re`? Having it at this top level is not brms standard.

R/brmsframe.R
```diff
@@ -55,6 +71,9 @@ brmsframe.brmsterms <- function(x, data, frame = NULL, basis = NULL, ...) {
     )
   }
   class(x) <- c("brmsframe", class(x))
+  if (is.null(frame)) {
```
paul-buerkner, line 74: see above. These validations should all be performed inside `frame_re`. Or what is the reason this has to be done more globally? Likely I don't quite understand all the details required to make r2z work. It seems validating it requires access to more than just the `re` frame (but other frames as well). Is this strictly required?

paul-buerkner (commented):
I started checking the PR again but realized I have too many gaps in my understanding of the method details to do a good review now. Can you point me to your current best write-up of the method so I can check again? Specifically, I need to better understand the interaction between "fixed" and "random" effects to figure how how to best implement a validation framework than combines both information.

Issue #1923, opened 2026-08-21
`posterior_epred()` drops the lower bound for truncated discrete models


deeb01:
For truncated discrete responses `posterior_epred()` averages over `(lb+1):ub` and normalizes by `F(ub) - F(lb)`, so `lb` carries no mass. `posterior_predict()` and the fitted model both include it.

```r
lambda <- 3; lb <- 2; ub <- 6
prep <- structure(list(ndraws = 2L, nobs = 1L), class = "brmsprep")
prep$dpars <- list(mu = matrix(lambda, nrow = 2, ncol = 1))
prep$family <- poisson()
prep$data <- list(Y = 3, lb = lb, ub = ub)

brms:::posterior_epred_trunc_poisson(
  prep, lb = matrix(lb, 2, 1), ub = matrix(ub, 2, 1)
)[1]
#> [1] 3.958763

sum((lb:ub) * dpois(lb:ub, lambda)) /   # E[Y | lb <= Y <= ub]
  (ppois(ub, lambda) - ppois(lb - 1, lambda))
#> [1] 3.386861
```

17% high; `posterior_predict()` averages 3.384 over 40k draws.

In `posterior_epred_trunc_discrete()` the kernel starts at `min_lb + 1` and the denominator uses `cdf(lb)`. Both want `lb - 1`, as `stan_log_lik_trunc()` emits for integer responses. Affects binomial, poisson, negbinomial, negbinomial2 and geometric, so `fitted()` and `conditional_effects()` too.

Same convention as #1903. `pp_cdf()`, `pp_density()` and `pp_quantile()` share it, but they back `posterior_pit()` and probably want their own discussion.

PR #1922, opened 2026-08-21
Use inclusive lower truncation bounds for integer responses

deeb01:
Closes #1903.

## Summary

For truncated integer responses `log_lik()` normalized by $P(lb < Y \le ub)$ while the fitted model uses $P(lb \le Y \le ub)$, so `loo()` and `waic()` disagreed with the model that was sampled.

## Problem

`stan_log_lik_trunc()` emits `lb - 1` for integer responses, so the support is `lb:ub`. `log_lik_truncate()` passed `lb` through unmodified. For `poisson(3)` truncated to `[2, 6]` the pmf summed to 1.41, and disagreed with `posterior_predict()` — which already includes `lb` — by up to 0.12:

| | Y=2 | Y=3 | Y=4 | Y=5 | Y=6 | sum |
|---|---|---|---|---|---|---|
| `posterior_predict()`, 50k draws | 0.2918 | 0.2930 | 0.2183 | 0.1326 | 0.0644 | 1.000 |
| `log_lik()` before | 0.4124 | 0.4124 | 0.3093 | 0.1856 | 0.0928 | **1.412** |
| `log_lik()` after | 0.2920 | 0.2920 | 0.2190 | 0.1314 | 0.0657 | 1.000 |

## Suggested solution

Subtract one from `lb` when `use_int(prep$family)`. Affects poisson, negbinomial, binomial, geometric, beta_binomial, discrete_weibull and the zero-inflated and hurdle counts.

## Note

Continuous responses and upper-only bounds are unaffected, as is interval censoring — the generated Stan emits no `- 1` there. `ceiling()` is applied because `log_lik(fit, newdata = ...)` can introduce a non-integer bound through a data column, which Stan cannot receive.

Before this, `log_lik()`, `posterior_epred()` and `pp_cdf()` all used the exclusive bound and agreed with each other, while `posterior_predict(output = "random")` and the fitted model used the inclusive one. Fixing `log_lik()` leaves two paths behind:

- `posterior_epred_trunc_discrete()` sums over `(lb+1):ub`, running ~14% high on a truncated `poisson(3)` in `[2, 6]`
- `pp_cdf()`, `pp_density()` and `pp_quantile()` normalize by `F(ub) - F(lb)`, disagreeing with `pp_random_discrete()` beside them: `pp_density(2, "pois", lb = 2, ub = 6, lambda = 3)` is 0.4124 against an empirical 0.291

[...]

PR #1928, opened 2026-08-27
Include the lower bound in truncated discrete posterior_predict outputs

deeb01:
Second follow-up from #1922, same convention as #1923. @florence-bockting — this one touches your `posterior_predict` work.

## Summary

`pp_cdf()`, `pp_density()` and `pp_quantile()` normalized by `F(ub) - F(lb)`, so the lower bound carried no mass, while `pp_random_discrete()` beside them includes it. `posterior_predict(output = "probability" | "pit" | "density" | "quantile")` therefore disagreed with `output = "random"`, with `log_lik()` and with the fitted model.

## Problem

`poisson(3)` truncated to `[2, 6]`, before:

- `pp_density()` over `2:6` summed to 1.4124
- `pp_cdf(2, randomized = FALSE)` was 0, so `lb` carried no mass
- `pp_quantile(0.001)` was 3, below the smallest value `posterior_predict()` can draw

Worse, the randomized PIT went negative: over 200k draws from the true truncated pmf, 29% of PIT values were below zero and KS gave `D = 0.292, p < 2e-16`. After, `D = 0.002, p = 0.36`.

## Suggested solution

An `int_response` flag, set by `predict_discrete_helper()`, moving the normalizer to `F(ub) - F(ceiling(lb) - 1)`. `ceiling()` matches `log_lik_truncate()`; plain `lb - 1` would regress non-integer bounds. `pp_quantile()` is floored at `ceiling(lb)`, since `p = 0` otherwise inverts to the value below the support. Continuous paths are untouched.

One thing beyond the title, without which the rest is invisible: `check_discrete_trunc_bounds()` ran for every `output` and ends in `round()`, so the corrected probabilities, PIT values and densities were collapsed to 0 or 1 before reaching the user, with a spurious `"% of all predicted values were invalid"` warning. It now runs only for `output = "random"`, the only one returning predicted responses. Happy to split that out if you would rather.

`expect_pp_truncation()` compared `pp_fun()` against these same functions, so it agreed with either convention. It now passes the flag, and a new test pins the convention against `stats::` references instead.

## TODO

+ [x] `int_response` in `pp_cdf`, `pp_density`, `pp_quantile`
+ [x] tests against `stats::` rather than self-consistency, plus the `"pit"` branch and the rounding
+ [ ] `log_lik_com_poisson()`, which ignores truncation entirely — last of the three

### DIFF
--- a/load.R
+++ b/load.R
@@ -1,2 +1,0 @@
-.libPaths(c("/Users/deeb/src/OSS/brms/rlib", .libPaths()))
-suppressMessages(pkgload::load_all("/Users/deeb/src/OSS/brms/wt-pp", quiet = TRUE, export_all = TRUE))
--- a/load.R
+++ b/load.R
@@ -1,0 +1,2 @@
+.libPaths(c("/Users/deeb/src/OSS/brms/rlib", .libPaths()))
+suppressMessages(pkgload::load_all("/Users/deeb/src/OSS/brms/wt-pp", quiet = TRUE, export_all = TRUE))
--- a/NEWS.md
+++ b/NEWS.md
@@ -27,6 +27,9 @@ Thanks to Gidon Frischkorn. (#1450)
 
 ### Bug Fixes
 
+* Stop rounding `posterior_predict` output to integers for discrete models
+unless `output` is `"random"`, which collapsed probabilities and densities
+to 0 or 1. Thanks to Ahmed Eldeeb. (#1923)
 * Include the lower bound in `posterior_predict` outputs `"probability"`,
 `"pit"`, `"density"` and `"quantile"` for truncated discrete models,
 matching the other post-processing methods. Thanks to Ahmed Eldeeb. (#1923)
--- a/NEWS.md
+++ b/NEWS.md
@@ -27,6 +27,9 @@ Thanks to Gidon Frischkorn. (#1450)
 
 ### Bug Fixes
 
+* Include the lower bound in `posterior_predict` outputs `"probability"`,
+`"pit"`, `"density"` and `"quantile"` for truncated discrete models,
+matching the other post-processing methods. Thanks to Ahmed Eldeeb. (#1923)
 * Normalize truncated `log_lik` by `P(lb <= Y <= ub)` for integer
 responses, matching the generated Stan code, so that `loo` and `waic`
 agree with the model that was fitted. Thanks to Ahmed Eldeeb. (#1903)
--- a/R/posterior_predict.R
+++ b/R/posterior_predict.R
@@ -224,7 +224,10 @@ posterior_predict.brmsprep <- function(object, transform = NULL, sort = FALSE,
     out <- do_call(cbind, out)
   }
   colnames(out) <- rownames(out) <- NULL
-  if (use_int(object$family)) {
+  if (use_int(object$family) && output == "random") {
+    # the other outputs are probabilities, densities or quantiles rather
+    # than predicted responses, so neither the bounds check nor the
+    # rounding applies to them; see #1923
     out <- check_discrete_trunc_bounds(
       out, lb = object$data$lb, ub = object$data$ub
     )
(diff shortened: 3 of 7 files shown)


load.R
florence-bockting, line 1: I think this file slipped through. Please remove it.

florence-bockting (commented):
Thank you for the PR, it looks overall very good. Just some minor comments:

- There is one file `load.R` which needs to be removed before we can merge.
- Then we have one merge conflict (the NEWS.md) due to merging your other PR earlier today.
- And finally, we need to make adjustments to two further test files:
    + In `tests/testthat/helper-distributions.R` your new additions need to be added to `expect_truncated_cdf_density_quantile`:
```
expect_truncated_cdf_density_quantile <- function(entry, lb, ub,
                                                   q = NULL, p = 0.4,
                                                   tol = 1e-7) {
  if (!has_fun(entry, "p") || !isTRUE(entry$flags$truncation)) {
    return(invisible(TRUE))
  }
  if (is.null(q)) q <- entry$q_ref
>  # the lower bound is inclusive for integer responses, so the normalizer
>  # uses F(ceiling(lb) - 1). The rule is written out here rather than taken
>  # from brms, so that these checks pin the convention. See #1923
>  int_response <- isTRUE(entry$flags$discrete_support)
>  lb_norm <- if (int_response) ceiling(lb) - 1 else lb
  F <- function(x) as.numeric(call_dist(entry$p, x, entry$params))
>  denom <- F(ub) - F(lb_norm)
  testthat::expect_true(denom > 0, info = paste(entry$name, "trunc mass"))

  # truncated CDF (cheatsheet)
>  F_tr <- (F(q) - F(lb_norm)) / denom
  F_tr_pp <- do.call(
    brms:::pp_cdf,
    c(list(q = q, distribution = entry$backend, lb = lb, ub = ub,
>           randomized = FALSE, int_response = int_response), entry$params)
  )
  testthat::expect_equal(as.numeric(F_tr_pp), F_tr, tolerance = tol,
                         info = paste(entry$name, "trunc CDF"))

  if (has_fun(entry, "d")) {
    dens <- as.numeric(call_dist(entry$d, q, entry$params))
    dens_tr <- if (q < lb || q > ub) 0 else dens / denom
    dens_pp <- do.call(
      brms:::pp_density,
>      c(list(q = q, distribution = entry$backend, lb = lb, ub = ub,
>             int_response = int_response), entry$params)
    )
    testthat::expect_equal(as.numeric(dens_pp), dens_tr, tolerance = tol,
                           info = paste(entry$name, "trunc density"))
  }

  if (has_fun(entry, "q")) {
>    # p > 0 always inverts to at least ceiling(lb), so the floor that
>    # pp_quantile() applies for integer responses is inactive here
>    p_star <- p * denom + F(lb_norm)
    q_tr <- as.numeric(call_dist(entry$q, p_star, entry$params))
    q_pp <- do.call(
      brms:::pp_quantile,
>      c(list(p = p, distribution = entry$backend, lb = lb, ub = ub,
>             int_response = int_response), entry$params)
    )
    testthat::expect_equal(as.numeric(q_pp), q_tr, tolerance = 1e-5,
                           info = paste(entry$name, "trunc quantile"))
  }
  invisible(TRUE)
}
```
+ and in `tests/testthat/tests.distributions_relations.R` we should add a case in the truncation test where we test fractional lb (line 45):
```
test_that("truncation formulas match compute_* helpers", {
  for (entry in dist_active_entries(truncation = TRUE)) {
    if (isTRUE(entry$flags$discrete_support)) {
      expect_truncated_cdf_density_quantile(entry, lb = 1, ub = 6)
>      # a non-integer bound admits the same support as ceiling(lb), which
>      # separates ceiling(lb) - 1 from a plain lb - 1. See #1923
>      expect_truncated_cdf_density_quantile(entry, lb = 1.5, ub = 6)
    } else {
      qs <- as.numeric(call_dist(entry$q, c(0.2, 0.8), entry$params))
      expect_truncated_cdf_density_quantile(entry, lb = qs[1], ub = qs[2])
    }
  }
})
```

florence-bockting:
@deeb01 Thank you for your time, looks good to me now. @paul-buerkner whenever you are fine as well with it, it can be merged.

paul-buerkner:
Thank you both!

deeb01:
All three done — `load.R` removed (that was a local bootstrap file of mine, sorry), master merged with the NEWS conflict resolved, and both test changes applied as written.

Good catch on `expect_truncated_cdf_density_quantile`. It was asserting the exclusive convention for 13 discrete backends while `helper-posterior-predict.R` asserted the inclusive one, so the suite held both at once. The fractional `lb = 1.5` case is the one that separates `ceiling(lb) - 1` from a plain `lb - 1`.

`R CMD check` clean locally, 4421 passing.

PR #1929, opened 2026-08-28
Align R-side COM-Poisson functions with Stan

### LINKED ISSUE #1927: com_poisson: Stan and R-side parameterisations diverged in #1765
Big fan of brms, thanks for all the work on it.

I think #1765 updated the Stan side of `com_poisson` from the mode to the classical parameterisation, but `R/distributions.R` wasn't part of that PR, so `dcom_poisson()` is still on the old one. They only agree at `shape == 1`.

Result:

```r
set.seed(1)
d <- data.frame(y = pmax(0, round(rnorm(200, 6, 1))))   # mean 6.03, var 0.92
m <- brm(y ~ 1, data = d, family = brmsfamily("com_poisson"), refresh = 0)

c(data_mean = mean(d$y), epred = mean(posterior_epred(m)))
#>  data_mean      epred 
#>      6.025 724010.359
```

In `0d295b0`, `com_poisson_log_lpmf` went from `nu * (y * log_mu - lgamma(y + 1))` to `y * log_mu - nu * lgamma(y + 1)`, and `log_Z_com_poisson_approx()` from `nu * exp(log_mu)` to `nu * exp(log_mu / nu)`. Both changed in the same commit, so the Stan side is internally consistent either way — it's only the R side that's out of step. `dcom_poisson()` still has `shape * (x * log_mu - lgamma(x + 1))` and the `# com_poisson in brms uses the mode parameterization` comment.

Thus `log_lik()` evaluates a different density from the one Stan sampled whenever `shape != 1`, so `loo`, `waic`, `pp_check` and `posterior_epred` are all affected.

Since the truncation bound used in #1765 is formulated for the classical series, that may explain how the switch occurred. I'm not sure whether the intended fix is to restore the historical brms mode parameterisation on the Stan side while evaluating the new normalizer with the appropriate transformation, or to migrate the R-side functions to the classical parameterisation. The latter would change the meaning of coefficients relative to pre-#1765 fits. `com_poisson` isn't currently listed in `?brmsfamily`, so there's no user-facing statement of what `mu` means either way.

Happy to open a PR if that would help.

brms 2.23.0, cmdstanr backend.

LeonidasZhak:
## Summary
Align the R-side `com_poisson` distribution functions and mean with the parameterization used by Stan.

## Thanks to maintainers
Thanks to Paul and the brms maintainers for maintaining the package and for pointing out the preferred direction in the discussion of this issue.

## Issue or motivation
Fixes #1927. After #1765 changed the Stan implementation, the R-side density, CDF, quantile, normalizing constant, and mean calculations still used the previous parameterization.

## Root cause
The R functions multiplied the complete log-Poisson term by `shape`, whereas the updated Stan model applies `shape` only to the factorial term. The approximation formulas were also still expressed in terms of the previous parameterization.

## Change
- Align the exact R-side distribution and mean calculations with the current Stan formulas.
- Port the corresponding normalizing-constant and mean approximations.
- Keep vectorized CDF accumulation aligned for inputs with different cutoffs.
- Add regression coverage for density, CDF, quantile, and mean calculations.

## Tests
- `R CMD build --no-manual --no-build-vignettes .`
- `R CMD check --as-cran --no-manual --ignore-vignettes brms_2.23.1.tar.gz`: passed with the repository's existing two NOTES.
- Full `testthat.R` run: passed.

## Scope
This changes only the R-side calculations and regression coverage for `com_poisson`. It does not change the Stan implementation, the public function interface, or unrelated distributions and vignettes.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -27,6 +27,8 @@ Thanks to Gidon Frischkorn. (#1450)
 
 ### Bug Fixes
 
+* Align the R-side `com_poisson` distribution functions and mean with the
+  parameterization used by Stan. (#1927)
 * Normalize truncated `log_lik` by `P(lb <= Y <= ub)` for integer
 responses, matching the generated Stan code, so that `loo` and `waic`
 agree with the model that was fitted. Thanks to Ahmed Eldeeb. (#1903)
@@ -2295,5 +2297,4 @@ have proper priors by default.
 # brms 0.1.0
 
 * Initial release version
-
 
--- a/R/distributions.R
+++ b/R/distributions.R
@@ -1383,12 +1383,11 @@ mean_discrete_weibull <- function(mu, shape, M = 1000, thres = 0.001) {
 }
 
 # PDF of the COM-Poisson distribution
-# com_poisson in brms uses the mode parameterization
 dcom_poisson <- function(x, mu, shape, log = FALSE) {
   x <- round(x)
   log_mu <- log(mu)
   log_Z <- log_Z_com_poisson(log_mu, shape)
-  out <- shape * (x * log_mu - lgamma(x + 1)) - log_Z
+  out <- x * log_mu - shape * lgamma(x + 1) - log_Z
   if (!log) {
     out <- exp(out)
   }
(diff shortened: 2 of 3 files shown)


paul-buerkner:
@LeonidasZhak was this the same issue as fixed (differently) in #1930? If yes, can we close this PR?

LeonidasZhak:
Yes, #1930 fixed the same underlying #1927 issue along the preferred route: restoring the Stan-side COM-Poisson functions to the explicit mode parameterization while keeping the newer normalizer machinery. That also covered the CDF/normalizer consistency checks this PR was trying to address from the R side.

Thanks for checking. I'll close this PR to avoid keeping the duplicate/conflicting path in the review queue.

PR #1930, opened 2026-09-01
Restore mode parameterization of com_poisson on the Stan side

### LINKED ISSUE #1927: com_poisson: Stan and R-side parameterisations diverged in #1765

jbogomolovas2:
Closes #1927.

This restores the historical brms mode parameterization of `com_poisson` on the Stan side while retaining the numerical improvements introduced in #1765.

In the brms parameterization,

```text
p(y) ∝ (mu^y / y!)^nu
```

so the corresponding classical COM-Poisson rate is

```text
lambda = mu^nu
```

or, on the log scale,

```text
log_lambda = nu * log_mu
```

#1765 changed the Stan likelihood to use the classical parameterization directly, while the R-side distribution functions remained in the mode parameterization. The two therefore disagree whenever `shape != 1`.

## Changes

- Restore the mode-parameterized Stan log-PMF.
- Keep `log_Z_com_poisson()` explicitly parameterized in terms of the brms modal location `log_mu`, matching the R-side function.
- Convert internally with

  ```text
  log_lambda = nu * log_mu
  ```

  before calling the classical-series machinery introduced in #1765.
- Make the CDF use the same transformation consistently.
- Restore the intended rule-of-thumb region for selecting the asymptotic normalizer:

  ```text
  lambda >= 1.5
  lambda^(1 / nu) >= 1.5
  ```

The CDF and approximation-region changes were additional consequences of the same parameterization switch that I found while validating the original fix. They are not independent changes: the CDF contained a mixture of the two coordinate systems, and the approximation predicate had been left textually unchanged when the meaning of `log_mu` changed.

The lower-level helpers `log_k_term()` and `log_Z_com_poisson_approx()` remain expressed in terms of `log_lambda`. These are the classical-series computational helpers introduced in #1765 rather than user-facing distribution functions, and keeping them on that scale allows the computational improvements from #1765 to be retained essentially unchanged.

## R-side computation

I also looked at whether the #1765 stopping criterion should be ported to the R implementation.

The current R exact summation stops once the next log term is below `log(1e-16)`, whereas the Stan implementation uses the #1765 remainder bound with `leps = -23 * log(2)` and checks the stopping criterion in buckets.

On difficult exact-branch cases, the existing R implementation is actually more accurate. For example:

| `mu` | `nu` | R error | Stan error |
|---:|---:|---:|---:|
| 1.0 | 0.05 | 1.1e-14 | 5.2e-09 |
| 0.5 | 0.10 | 4.4e-16 | 5.2e-09 |

So I did not port the Stan stopping criterion to R, since doing so would reduce numerical accuracy rather than improve it.

## Verification

I compiled `fun_com_poisson.stan` with CmdStan and compared the Stan functions directly with the R implementation at `shape != 1`.

The regression grid covers:

- the exact normalizer branch;
- the asymptotic normalizer branch;
- both over- and under-dispersion;
- the CDF, including its first nontrivial terms.

I also checked the approximation boundary directly at `mu = 1.5`; Stan and R agree to numerical precision.

The analytical tests pass:

```text
FAIL 0 | WARN 0 | SKIP 0 | PASS 168
```

Additional tests pin down the meaning of `mu` independently of the normalizer and compare `dcom_poisson()` against a brute-force evaluation of

```text
(mu^y / y!)^nu
```

## Existing fits

Fits produced by affected versions with `shape != 1` were sampled under the classical parameterization. Updating brms cannot reinterpret those draws as mode-parameterized fits, so affected models should be refitted.

A NEWS entry documents this.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -56,6 +56,15 @@ such that `step(0) == 1` thanks to Daniel Sabanes Bov. (#1734)
 * Make `read_csv_as_stanfit()` store `adapt_delta` and `max_treedepth` values in
 `$control` so rstan can find these values. Thanks to Tristan Mahr (#1767).
 * Enable updating argument `data2` for `brmsfit_multiple` objects. (#1776)
+* Restore the mode parameterization of the `com_poisson` family on the Stan
+side, which had switched to the classical parameterization in #1765 while the
+R-side density functions did not, so that R-side post-processing no longer
+disagrees with the sampled density whenever `shape != 1`. Also correct the
+condition selecting the asymptotic normalizing constant, which was left
+unchanged by that switch and so no longer described the intended region.
+Models fitted with `com_poisson` and `shape != 1` using affected versions
+were sampled under the classical parameterization and should be
+refitted. (#1927)
 * Fix several other minor bugs.
 
 ### Other Changes
--- a/inst/chunks/fun_com_poisson.stan
+++ b/inst/chunks/fun_com_poisson.stan
@@ -1,13 +1,15 @@
   // log approximate normalizing constant of the COM poisson distribuion
   // based on equations (4) and (31) of doi:10.1007/s10463-017-0629-6
-  // Args: see log_Z_com_poisson()
-  real log_Z_com_poisson_approx(real log_mu, real nu) {
+  // Args:
+  //   log_lambda: log of the classical rate parameter, that is nu * log(mu)
+  //   nu: positive shape parameter
+  real log_Z_com_poisson_approx(real log_lambda, real nu) {
     real nu2 = nu^2;
-    real log_common = log(nu) + log_mu / nu;
+    real log_common = log(nu) + log_lambda / nu;
     array[4] real resids;
     real ans;
-    real lcte = (nu * exp(log_mu / nu)) -
-      ((nu - 1) / (2 * nu) * log_mu +
+    real lcte = (nu * exp(log_lambda / nu)) -
+      ((nu - 1) / (2 * nu) * log_lambda +
        (nu - 1) / 2 * log(2 * pi()) + 0.5 * log(nu));
     real c_1 = (nu2 - 1) / 24;
     real c_2 = (nu2 - 1) / 1152 * (nu2 + 23);
@@ -22,11 +24,11 @@
 
   // log of kth term of the normalizing series of the COM Poisson distribution
   // Args:
-  //   log_mu: log location parameter
-  //   shape: positive shape parameter
+  //   log_lambda: log of the classical rate parameter, that is nu * log(mu)
+  //   nu: positive shape parameter
   //   k: k-th term
-  real log_k_term(real log_mu, real nu, int k) {
-    return (k - 1) * log_mu - nu * lgamma(k);
+  real log_k_term(real log_lambda, real nu, int k) {
+    return (k - 1) * log_lambda - nu * lgamma(k);
   }
 
   // bound for the remainder of the normalizing series of the COM Poisson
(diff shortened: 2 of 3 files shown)


jbogomolovas2:
I pushed a revised version after finding two additional consequences of the same parameterization switch while validating the fix: the CDF and the asymptotic-branch condition also needed adjustment. I also simplified the implementation so the Stan and R distribution functions now both use `log_mu` at their interface. The PR description is updated accordingly.

paul-buerkner:
Did you forget to push the latest changes? I don't see changes on the R side for example? In terms of parameterization, I would prefer the main distribution functions' interface to still be `mu` rather than `log_mu` but perhaps I misunderstood your comment.

jbogomolovas2:
Sorry about the confusion, I was imprecise with my wording: the main distribution functions still accept `mu`, as you prefer. That hasn’t changed.

I meant the internal `log_Z_com_poisson()` helper. It already accepted `log_mu` in both R and Stan before my patch, but the same name meant different things. The patch fixes its meaning on the Stan side to match R’s mode parameterization.

The latest changes are pushed. I did initially propose bringing the Stan stopping machinery over to R, but after checking it, I left that out: the existing R calculation was more accurate in the difficult cases I tested. I should have said that more clearly.

I propose to divide and conquer: this PR fixes the Stan-side parameterization mismatch, including the CDF and the condition selecting the Gaunt approximation. The R distribution functions already use the historical mode parameterization, so I left their implementation unchanged. 

Now Stan and R are on the same page, which fixes the weird numerical blowouts caused by that mismatch. 

The fuhrer quirks I found, including the approximation’s accuracy in some regimes, can go in a separate issue.

Thanks for checking!

paul-buerkner:
Looks good, thank you! Will merge as soon as checks have been run.

paul-buerkner:
Checks fail for reasons unrelated to this PR. Tests pass locally, so I will merge now. Thank you again for your contribution to brms!

PR #1932, opened 2026-09-15
Evaluate seven cdfs on the log scale in both tails

deeb01:
@florence-bockting — the follow-up promised in the Note on #1904.

## Summary

Seven cdfs formed the survival function as `1 - F`, or the cdf as `1 - exp(log_ccdf)`, and only then took the log, so the tail was gone before `log.p` applied and truncated `log_lik` could not reach these families.

## Problem

```r
brms:::pfrechet(50, 0, 1, 20, lower.tail = FALSE, log.p = TRUE)  #> -Inf, exact -78.24
```

Same for `gen_extreme_value`, `discrete_weibull`, `asym_laplace`, `inverse.gaussian`, `exgaussian`, and the zero-inflated and hurdle wrappers in the opposite tail.

## Suggested solution

Evaluate both tails on the log scale. Four have exact closed forms: `log F = -q^(-shape)` for frechet, `log F = -t` for gen_extreme_value, `log S = (x+1)^shape * log(mu)` for discrete_weibull, and asym_laplace is exact on each side of `mu`. inverse.gaussian takes its cdf from `inv_gaussian_lcdf()` in `inst/chunks/fun_inv_gaussian.stan`, which also removes an overflow of `exp(2 * shape / mu)`.

Two supporting changes: `log_sum_exp(-Inf, -Inf)` returned `NaN` where Stan returns `-Inf`, and a saturated probability reaches `log1m_exp(0)`, which is `NaN` by Stan's convention — right for a generic argument, wrong for a probability — so those call sites go through a new `log1m_prob()`.

## Note

The inverse.gaussian and exgaussian survival functions still saturate far out: their two terms are asymptotically equal, since `2 * shape / mu - b^2 / 2` equals `-a^2 / 2` identically, so the difference cancels however it is arranged. Recovering it needs an asymptotic expansion, left alone and noted in the code. Both are accurate well past the old behaviour — at `q = 20`, -457.93 against a true -457.93.

## TODO

+ [x] log-scale evaluation for the seven cdfs
+ [x] `log_sum_exp` boundary, and `log1m_prob()` for saturated probabilities
+ [x] tests against closed forms and against the boundaries
+ [ ] asymptotic expansion for the inverse.gaussian and exgaussian far tails


paul-buerkner:
Thank you @deeb01! 

@florence-bockting would it be possible for you to review this PR?

R/distributions.R
florence-bockting: There is a typo, which we can probably address directly.

florence-bockting (approved):
Thank you, @deeb01 for the PR. 
Looks overall good to me.

Just a minor style aspect for @paul-buerkner: The code includes longer comments and I don't know whether you like/dislike multi-line comments in your code base.

paul-buerkner:
Thank you! Merging now :-)

Issue #1935, opened 2026-09-16
Prediction broken with mcgv mrf splines


wds15:
`brms` can fit `mrf` type models from `mgcv`, but the posterior predict functions are broken whenever the `newdata`  argument is used. The prediction only works with the data used to fit (and only of the `newdata` argument is not used) - see below:

``` r
suppressPackageStartupMessages(library(brms))

set.seed(1)

G <- 4L
nodes <- paste0("n", seq_len(G))
# Path-graph Laplacian: adjacent node effects are smoothed toward each other.
penalty <- crossprod(diff(diag(G)))
dimnames(penalty) <- list(nodes, nodes)

# Each observation carries a factor matrix of nodes and matching numeric weights.
add_matrices <- function(data, weights) {
  node <- factor(rep(nodes, each = nrow(data)), levels = nodes)
  dim(node) <- dim(weights)
  data$node <- node
  data$weight <- weights
  data
}

weights <- rbind(diag(G), diag(G))
dat <- add_matrices(
  data.frame(y = as.numeric(weights %*% seq(0.1, 0.4, by = 0.1))),
  weights
)

fit <- suppressMessages(brm(
  y ~ 0 + s(
    node, by = weight, bs = "mrf", k = G,
    xt = list(penalty = penalty)
  ),
  data = dat,
  backend = "cmdstanr",
  chains = 1,
  iter = 400,
  refresh = 0,
  silent = 2
))
# Prediction works when brms implicitly reuses the original model data.
dim(posterior_linpred(fit))
#> [1] 200   8

# The factor matrix has one row per original observation.
dim(dat$node)
#> [1] 8 4

# Supplying those identical data explicitly fails during newdata validation.
posterior_linpred(fit, newdata = dat)
#> Error in `[[<-.data.frame`:
#> ! replacement has 32 rows, data has 8
```

<sup>Created on 2026-09-16 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

<details style="margin-bottom:10px;">

<summary>

Session info
</summary>

``` r
sessioninfo::session_info()
#> ─ Session info ───────────────────────────────────────────────────────────────
#>  setting  value
#>  version  R version 4.6.1 (2026-06-24)
#>  os       Ubuntu 26.04.1 LTS
#>  system   aarch64, linux-gnu
#>  ui       X11
#>  language (EN)
#>  collate  en_US.UTF-8
#>  ctype    en_US.UTF-8
#>  tz       UTC
[...]

PR #1936, opened 2026-09-16
Issue mgcv predict, fixes #1935

wds15:
Maintaining the dimension of factors when the newdata code path is used solves the issue for mgcv.


paul-buerkner:
Perfect, thank you!

